PR #11, opened 2020-07-07
tests: use pytest instead of unittest

mxschmitt:
As discussed a small showcase how tests with pytest will look like. They are simpler and its easier to have reusable content with pre and post methods called fixtures. In general its more common to use asserts instead of the expect way like we did in JavaScript. For parallelism there are packages out there like [that](https://github.com/pytest-dev/pytest-xdist).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@
 
 import setuptools
 
-with open('README.md', 'r', encoding="utf-8") as fh:
+with open('README.md', 'r', encoding='utf-8') as fh:
   long_description = fh.read()
 
 setuptools.setup(
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@
 
 import setuptools
 
-with open('README.md', 'r') as fh:
+with open('README.md', 'r', encoding="utf-8") as fh:
   long_description = fh.read()
 
 setuptools.setup(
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
 
 import setuptools
 
-with open('README.md', 'r') as fh:
+with open('README.md', 'r', encoding="utf-8") as fh:
```
pavelfeldman, line 17: I already picked single quotes for the project (not sure it was the right call). But let's stick with it!

test/conftest.py
```diff
@@ -0,0 +1,29 @@
+import playwright_web
```
pavelfeldman, line 1: We need Apache 2 copyright headers

test/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+import pytest
+
+
+@pytest.fixture(scope="session")
```
pavelfeldman, line 5: To make sure I understand, this runs once per entire test session (all modules). What are we doing to the event loop here?

test/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.fixture
+async def context(browser):
+    context = await browser.newContext()
+    yield context
+    await context.close()
+
+@pytest.fixture
+async def page(context):
+    page = await context.newPage()
+    yield page
```
pavelfeldman, line 28: I have to admit I don't understand how this part works. Need to read something about it.

setup.py
mxschmitt: ok for me 👍

PR #12, opened 2020-07-07
cleanup: migrate test suite to pytest

mxschmitt:
Some thoughts:
- Would be cool to have [black](https://github.com/psf/black) as a default formatter enforced, seems [partly](https://github.com/search?q=org%3Amicrosoft+%22psf%2Fblack%22&type=Code) used at MS.
- Not fully sure regarding test structure, probably group them by type in directories (page, elementhandle, context etc) and then by function if there needs to be proper edge case testing be done.
- `is None` instead of `== None` (see [here](https://stackoverflow.com/a/3257957/6512681))
- the "in syntax" like in JavaScript for checking if a string contains a substring

Now we can focus on proper features and tests 😄


README.md
```diff
@@ -47,13 +47,13 @@ python ./build_driver.py
...
-python -m unittest
+pytest
 ```
 
 Run tests with coverage:
 ```
 pip install coverage
-coverage run -m unittest
+coverage run -m pytest
 coverage html
```
suzil, line 57: Nit: Might be nicer in the future to use [pytest-cov](https://pypi.org/project/pytest-cov/).

test/conftest.py
```diff
@@ -14,7 +14,13 @@
 
 import playwright_web
 import pytest
+from .server import start_server
+import _thread
 
+# Will mark all the tests as async
+def pytest_collection_modifyitems(items):
+    for item in items:
+        item.add_marker(pytest.mark.asyncio)
```
suzil, line 23: Neat, didn't know about this.

test/conftest.py
```diff
@@ -41,3 +47,7 @@ async def page(context):
     page = await context.newPage()
     yield page
     await page.close()
+
+@pytest.fixture(autouse=True)
+def start_http_server():
+    _thread.start_new_thread(start_server, ())
```
suzil, line 53: To me, it seems like it would be better to use the higher-level [threading](https://docs.python.org/3/library/threading.html#module-threading) module instead of this one.

> The threading module provides an easier to use and higher-level threading API built on top of this module.

Since it starts with an underscore it seems like it's not meant to be so commonly used.

I think you should be able to write something like:
```py
threading.Thread(target=start_server).start()
```

PR #14, opened 2020-07-07
enh: fixed some Python types and Dicts

### LINKED ISSUE #3: Feature: support Python 3.7+ with typing extensions (interested in TypedDict)
I've seen following here: https://github.com/python/typeshed/issues/3500

import sys
```py
if sys.version_info >= (3, 8):
    from typing import TypedDict, Literal, overload  # pylint: disable=no-name-in-module
else:
    from typing_extensions import TypedDict, Literal, overload
```

mxschmitt:
Fixes #3

Changes:
- added these `if TYPE_CHECKING` conditions to fix circular imports and by that types which were not resolved
- added mypy to the local dependencies and readme to check for typing issues. Currently we have quite a lot of errors (around 40) would be nice to fix them and enforce it in the end on the CI

Tested on Python 3.7.7 and 3.8.2, works. Should I add 3.7 also to our testing matrix in the CI?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ dist/
 env/
 htmlcov/
 .coverage
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -63,6 +63,11 @@ python ./build_package.py
 python ./upload_package.py
 ```
 
+Checking for typing errors
+```sh
+mypy playwright_web
+```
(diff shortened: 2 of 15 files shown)


pavelfeldman:
A bot for 3.7 would be nice, yes!

PR #22, opened 2020-07-08
tests: run on all browser engines

mxschmitt:
Firefox and Chromium should pass. WebKit was not passing locally.


pavelfeldman:
I'll follow up with the WebKit fix

PR #36, opened 2020-07-10
chore: uncomment the Literal type

pavelfeldman:


pavelfeldman:
@mxschmitt Hm, mypy does not like those. I have another patch where I fixed all the mypys, but Literal broke it. Will look into it.

PR #37, opened 2020-07-10
chore: make mypy happy

pavelfeldman:


mxschmitt:
@pavelfeldman shall we enable mypy on the CI? So that they fail if invalid, would be just the mypy command.

PR #38, opened 2020-07-10
test: add click and check tests

pavelfeldman:


tests/test_click.py
```diff
@@ -50,10 +177,849 @@ async def test_not_wait_with_force(page, server):
     assert await page.evaluate("result") == "Was not clicked"
 
 
+async def test_wait_for_display_none_to_be_gone(page, server):
+    done = list()
```
mxschmitt, line 181: nit: done = [] seems more minimal.

tests/test_click.py
```diff
@@ -50,10 +177,849 @@ async def test_not_wait_with_force(page, server):
...
+        await page.click("button", timeout=0)
+        done.append(True)
+
+    clicked = asyncio.ensure_future(click())
+    await give_it_a_chance_to_click(page)
+    assert await page.evaluate("result") == "Was not clicked"
+    assert done == []
+    await page.evalOnSelector("button", "b => b.style.display = 'block'")
+    await clicked
+    assert done == [True]
```
mxschmitt, line 195: thats the cool thing about Python/assert. You can assert lists and dicts which simply works.

PR #43, opened 2020-07-11
tests: added download tests and linting

mxschmitt:
Changes:
- Since no typing errors are persistent anymore, I've enforced it in the CI so only valid code will be shipped
- In VSCode you can use flake8 as a formatter (Seems most common) which I also enforced (but disabled a few rules). But will prevent us from doing bad things. Is integrated nicely in VSCode, GitHub Actions and as a pre-commit hook.
- Added tests from the downloads.spec.js

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,10 @@ jobs:
         pip install -r local-requirements.txt
         pip install .
     - name: Lint
-      run: black --check .
+      run: |
+        black --check .
+        mypy .
+        flake8  playwright tests
     - name: Build driver
       run: python build_driver.py
       env:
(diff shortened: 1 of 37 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +40,10 @@ jobs:
         pip install -r local-requirements.txt
         pip install .
     - name: Lint
-      run: black --check .
+      run: |
+        black --check .
+        mypy .
+        flake8  playwright tests
```
ddelange, line 46: if you have flake8 and black as pre-commit hooks, you can remove them here and simply have `mypy . && pre-commit run --all-files` as lint stage on GH. mypy is an exception, as it should preferably not be used as pre-commit hook as it [won't be able to](https://github.com/pre-commit/mirrors-mypy) perform full analysis

mxschmitt: Good catch! Will do it probably in a refactoring PR for the CI bots.

PR #44, opened 2020-07-11
tests: added network tests and refactored test server

mxschmitt:
This test is missing and failing / hanging (not included in the PR) in the last line with the await. Not sure why it's not passing. From my understanding the RPC server does not return the request body anymore in this case (saw it in the protocol).

```py
async def test_response_text_should_wait_until_responses_completes(page, server):
    await page.goto(server.EMPTY_PAGE)
    ## Setup server to trap request.
    server_response_future = asyncio.Future()

    def handle_request(request):
        # In Firefox, |fetch| will be hanging until it receives |Content-Type| header
        # from server.
        request.setHeader("Content-Type", "text/plain charset=utf-8")
        request.write(b"hello ")
        server_response_future.set_result(request)

    server.set_route("/get", handle_request)
    # Setup page to trap response.
    request_finished = False

    def handle_request_finished(request):
        global request_finished
        request_finished = request_finished or "/get" in request.url

    page.on("requestfinished", handle_request_finished)
    # send request and wait for server response
    page_response = (
        await asyncio.gather(
            page.waitForEvent("response"),
            page.evaluate("() => fetch('./get', { method: 'GET'})"),
            server.wait_for_request("/get"),
        )
    )[0]
    server_response = await server_response_future
    assert server_response
    assert page_response
    assert page_response.status == 200
    assert request_finished is False

    response_text = asyncio.ensure_future(page_response.text())
    # Write part of the response and wait for it to be flushed.
    server_response.write(b"wor")
    # Finish response.
    server_response.write(b"ld!")
    server_response.finish()
    assert await response_text == "hello world!"
```


pavelfeldman:
If I modify your script to store the `fetch` result (promise to response) in property `fooo` and then append the following to after you do `server_response.finish()`:

```py
    print(await page.evaluate('''async () => {
        const response = await window.fooo;
        return response.text();
    }'''))
```

It never resolves. This suggests that the server does something wrong - it does not work without any Playwright logic...

mxschmitt:
> ```python
>    print(await page.evaluate('''async () => {
>         const response = await window.fooo;
>         return response.text();
>     }'''))
> ```

Yep, I was also trying to check if the server is working as expected. But the Chrome DevTools are reporting it correctly.

Screenshot 2020-07-12 at 10 54 14
Screenshot 2020-07-12 at 10 54 05

pavelfeldman:
devtools show you request and response (headers), i bet that the preview is empty though, because body was not retrieved.

PR #46, opened 2020-07-12
test: page tests, part 2

pavelfeldman:


tests/test_page.py
```diff
@@ -395,3 +396,142 @@ async def test_expose_function_should_work_with_complex_objects(page, server):
...
+        await page.setContent(f'<img src="{server.PREFIX + img_path}"></img>')
+    except Error as e:
+        error = e
+    assert "Timeout 1ms exceeded during" in error.message
+    assert isinstance(error, TimeoutError)
+
+
+async def test_set_content_should_await_resources_to_load(page, server):
+    img_path = "/img.png"
+    img_route = page._scope._loop.create_future()
```
mxschmitt, line 505: Is there a difference to asyncio.Futufe()?

tests/test_page.py
```diff
@@ -395,3 +396,142 @@ async def test_expose_function_should_work_with_complex_objects(page, server):
...
+async def test_set_content_should_respect_default_navigation_timeout(page, server):
+    page.setDefaultNavigationTimeout(1)
+    img_path = "/img.png"
+    # stall for image
+    await page.route(img_path, lambda route, request: None)
+
+    error = None
+    try:
+        await page.setContent(f'<img src="{server.PREFIX + img_path}"></img>')
+    except Error as e:
```
mxschmitt, line 497: Nit, would maybe simply it in the future: 
```py
import pytest

def test_zero_division():
    with pytest.raises(ZeroDivisionError):
        1 / 0
```

tests/test_page.py
pavelfeldman: Doc is saying

```This is the preferred way to create Futures in asyncio. This lets third-party event loops provide alternative implementations of the Future object (with better performance or instrumentation)```

tests/test_page.py
pavelfeldman: Will it le me test exception properties?

mxschmitt: Unfortunately not. But good to just check if an exception was thrown which we do in some tests.

PR #48, opened 2020-07-12
devops: separated linting into own job

mxschmitt:
drive-by: enforced typings on public facing API (playwright dir) and added pragmas to not useful if conditions for coverage reporting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,46 @@ on:
   pull_request:
     branches: [ master ]
 jobs:
+  infra:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: microsoft/playwright-github-action@v1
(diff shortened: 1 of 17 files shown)


.github/workflows/ci.yml
```diff
@@ -5,6 +5,46 @@ on:
...
+        python -m pip install --upgrade pip
+        pip install -r local-requirements.txt
+        pip install .
+    - name: Lint
+      run: |
+        black --check .
+        mypy .
+        flake8 playwright tests
+    - name: Build driver
+      run: python build_driver.py
```
pavelfeldman, line 43: Do we even need the driver here?

mxschmitt: Definitely not 👍 first thought was to add the publishing later to that job. But maybe it's better to do the publishing in a separate job.

mxschmitt:
@pavelfeldman should be ready, feel free to merge.

PR #54, opened 2020-07-13
fix: added querySelectorAll to Frame

mxschmitt:
`Page.querySelectorAll` called `Frame.querySelectorAll` which was not existing. Tests will follow in the future.

### DIFF
--- a/playwright/frame.py
+++ b/playwright/frame.py
@@ -134,7 +134,7 @@ class Frame(ChannelOwner):
     async def querySelectorAll(self, selector: str) -> List[ElementHandle]:
         return list(
             map(
-                cast(ElementHandle, from_nullable_channel),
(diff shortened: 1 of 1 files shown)


playwright/frame.py
```diff
@@ -131,6 +131,14 @@ async def querySelector(self, selector: str) -> Optional[ElementHandle]:
             await self._channel.send("querySelector", dict(selector=selector))
         )
 
+    async def querySelectorAll(self, selector: str) -> List[ElementHandle]:
+        return list(
+            map(
+                cast(ElementHandle, from_nullable_channel),
```
pavelfeldman, line 137: Should be from_channel, there can be no nulls.

mxschmitt: thx for fixing

PR #63, opened 2020-07-14
test: browser context tests, part 2

pavelfeldman:


tests/server.py
```diff
@@ -123,15 +127,18 @@ def stop(self):
     async def wait_for_request(self, path):
         future = asyncio.Future()
         self.request_subscribers[path] = future
-        future.add_done_callback(lambda f: self.request_subscribers.pop(path, None))
```
mxschmitt, line 126: Also wondered why I've added that.

PR #66, opened 2020-07-15
feat: implementation of sync API

mxschmitt:
Missing parts:
- [ ] Generate Sync classes out of annotation script with working typings


tests/test_sync.py
```diff
@@ -0,0 +1,149 @@
...
+    for page in [page1, page2]:
+        with pytest.raises(Error):
+            page.content()
+    browser.close()
+
+
+def test_sync_wait_for_selector(browser_name, launch_arguments):
+    browser = browser_types[browser_name].launch(**launch_arguments)
+    page = browser.newPage()
+    page.evaluate("() => setTimeout(() => document.write('<h1>foo</foo>'), 3 * 1000)")
```
pavelfeldman, line 93: Timeouts in tests will flake on busy boxes

tests/test_sync.py
```diff
@@ -0,0 +1,149 @@
...
+    with pytest.raises(Error) as exc:
+        page.waitForSelector("h1", timeout=2000)
+    assert "Timeout 2000ms exceeded during" in exc.value.message
+    browser.close()
+
+
+def test_sync_wait_for_event(browser_name, launch_arguments):
+    browser = browser_types[browser_name].launch(**launch_arguments)
+    page = browser.newPage()
+    page.evaluate(
```
pavelfeldman, line 111: Let's remove the timeout and see how we can reliably handle this use case.

tests/test_sync.py
```diff
@@ -0,0 +1,149 @@
...
+    page.goto(server.EMPTY_PAGE)
+    messages = []
+    page.on(
+        "request", lambda request: messages.append(f">>{request.method}{request.url}")
+    )
+    page.on(
+        "response",
+        lambda response: messages.append(f"<<{response.status}{response.url}"),
+    )
+    response = page.evaluate("""async ()=> (await fetch("/hello-world")).text()""")
```
pavelfeldman, line 144: Let's also log messages before and after the evaluation!

tests/test_sync.py
```diff
@@ -0,0 +1,221 @@
...
+    page = context.newPage()
+    page.setContent(
+        """
+    <h1 id="foo">Bar</h1>
+    """
+    )
+    assert (
+        page.querySelector("#foo").innerText() == page.querySelector("h1").innerText()
+    )
+    browser.close()
```
suzil, line 41:
```py
@pytest.fixture
def browser(browser_name, launch_arguments):
    browser = browser_types[browser_name].launch(**launch_arguments)
    yield browser
    browser.close()
```
seems nicer than having to create it and close it with each test

pavelfeldman:
@dgozman had an idea to try out `with`:

```py
with page.waitForEvent('popup') as popup:
    page.click()
```

It'll __ enter __ the event waiter, perform the click and you can finish waiting in the __ exit __. If it works, that'd mean that we can do sync!

tests/test_sync.py
mxschmitt: definitely!

mxschmitt:
> @dgozman had an idea to try out `with`:
> 
> ```python
> with page.waitForEvent('popup') as popup:
>     page.click()
> ```
> 
> It'll __ enter __ the event waiter, perform the click and you can finish waiting in the __ exit __. If it works, that'd mean that we can do sync!

The issue is, currently waitForEvent is sync. So it would never resolve and by that the click handler would never execute. We could adjust it, so it will not block and run the future in the background, but the issue with that is that the user would not be able to use the return value of the waitForEvent call because __exit__ can't modify it. That would be my understanding, but not fully sure.

tests/test_sync.py
mxschmitt: how do you mean that? just some initial value of "before" and append "end" to it?

pavelfeldman:
@mxschmitt, I was wondering if we could do something like this to get rid of the function in the `with` return value:

```py
class SyncWaitContextManager:
    def __init__(self, future: Future) -> None:
        self._value = LazyValue(future)

    def __enter__(self) -> Any:
        return self._value

    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
        self._value._materialize()


class LazyValue:
    def __init__(self, future: Future) -> None:
        self._future = future
        self._value = None

    def _materialize(self) -> Any:
        if not self._value:
            self._value = loop.run_until_complete(self._future)
        return self._value

    def __getattribute__(self, name: str) -> Any:
        if name.startswith("_"):
            return super().__getattribute__(name)
        return getattr(self._materialize(), name)
```

mxschmitt:
Closed as reworked in #78

PR #67, opened 2020-07-16
tests: added accessibility tests

mxschmitt:


playwright/accessibility.py
```diff
@@ -22,9 +22,9 @@ def __init__(self, channel: Channel) -> None:
         self._channel = channel
 
     async def snapshot(
-        self, interestingOnly: bool = None, root: ElementHandle = None
+        self, interestingOnly: bool = True, root: ElementHandle = None
```
mxschmitt, line 25: Unfortunately there is no undefined in Python. So JSON won't keep it out and will treat it in this case as null -> Client side default parameters will be ignored. So we either have our own "Undefined" which will  remove the keys in the serialisation or we duplicate the default values.

PR #69, opened 2020-07-19
feat(roll): roll Playwright to 1.2.0-next.1595094993390

pavelfeldman:


playwright/frame.py
```diff
@@ -289,21 +328,17 @@ async def selectOption(
...
-                values=convertSelectOptionValues(values),
-                timeout=timeout,
-                noWaitAfter=noWaitAfter,
-            ),
-        )
+    ) -> List[str]:
+        params = locals_to_params(locals())
+        if "values" in params:
+            values = params.pop("values")
+            params = dict(**params, **convert_select_option_values(values))
```
mxschmitt, line 335: nit: `{**params, **convert_select_option_values(values)}`
But probably just a personal preference.

playwright/network.py
```diff
@@ -173,3 +173,18 @@ def request(self) -> Request:
...
     def frame(self) -> "Frame":
         return self.request.frame
+
+
+def serialize_headers(headers: Dict[str, str]) -> List[Header]:
+    headers_list = list()
+    for name in headers:
+        header: Header = dict(name=name, value=headers[name])
+        headers_list.append(header)
+    return headers_list
```
mxschmitt, line 183: that would replace 5 lines: `[{"name": name, "value": value} for name, value in x.items()]`

But probably also a personal preference.

playwright/network.py
```diff
@@ -173,3 +173,18 @@ def request(self) -> Request:
...
+
+
+def parse_headers(headers: List[Header]) -> Dict[str, str]:
```
mxschmitt, line 186: same for here, but feel free to keep your solution if you think its more readable.

`return {header["name"]: header["value"] for header in headers}`

playwright/frame.py
pavelfeldman: I did not know that {} was more Python-ic! Happy to move across the board!

playwright/network.py
pavelfeldman: Done. That's what happens when one doesn't know Python.

playwright/network.py
pavelfeldman: Done

PR #70, opened 2020-07-19
enh: added context manager support

### LINKED ISSUE #68: [feature] context manager support
scope:
- `BrowserType.launch()` -> `withLaunch()`
- `BrowserType.launchServer()` -> `withLaunchServer()`
- `Browser.newContext()` -> `withContext()`
- `Context.newPage()` -> `withPage()`

Should we omit the "new" in the context methods?

Should be easy to integrate with the decorator in the `contextlib`: https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager

mxschmitt:
Added context manager support since its common in the Python world.

Closes #68

### DIFF
--- a/playwright/browser.py
+++ b/playwright/browser.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 import sys
+from contextlib import asynccontextmanager
 from types import SimpleNamespace
-from typing import Dict, List, Union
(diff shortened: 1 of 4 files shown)


pavelfeldman:
Let's hold this one off to see how users adopt pytest and poms

PR #74, opened 2020-07-19
feat(wait helper): add wait helper for client-side waitFor* methods

pavelfeldman:


playwright/network.py
```diff
@@ -58,7 +58,6 @@ def postData(self) -> Optional[str]:
     def headers(self) -> Dict[str, str]:
         return parse_headers(self._initializer["headers"])
 
-    @property
     async def response(self) -> Optional["Response"]:
```
mxschmitt, line 61: Seemed to be the last async property 👍

PR #78, opened 2020-07-22
feat(sync): generate sync api

pavelfeldman:


playwright/sync_base.py
```diff
@@ -0,0 +1,84 @@
...
+        wait_helper = WaitHelper()
+        wait_helper.reject_on_timeout(
+            timeout or 30000, f'Timeout while waiting for event "${event}"'
+        )
+        self._future = asyncio.ensure_future(
+            wait_helper.wait_for_event(sync_base._async_obj, event, predicate)
+        )
+
+    @property
+    def data(self) -> Any:
```
mxschmitt, line 42: I would probably use `.value` instead since it seems more common in the Python world. Pytest is using `.value` e.g. with `with pytest.raises()`

mxschmitt, line 42: I would probably use `.value` instead since it seems more common in the Python world. Pytest is using `.value` e.g. with `with pytest.raises()`

client.py
```diff
@@ -12,72 +12,82 @@
...
+
+    # print("\nCreating page2...")
+    # page2 = context.newPage()
+    # page2.on("framenavigated", lambda frame: print("Frame navigated to %s" % frame.url))
 
-    print("Navigating page2 to https://webkit.org...")
-    await page2.goto("https://webkit.org")
-    print("Page2 tile: %s" % await page2.title())
-    print("Pages in context: %d" % len(context.pages))
+    # print("Navigating page2 to https://webkit.org...")
```
mxschmitt, line 64: commit commented code is not that nice, lets have maybe both versions in it by a cli argument?

playwright/sync_base.py
pavelfeldman: Done

client.py
pavelfeldman: Oops, done

pavelfeldman:
I'll land this for now and continue off another box.

PR #79, opened 2020-07-22
chore(pre-commit): exclude browsers.json for EOL

mxschmitt:


pavelfeldman:
I'd like to ensure the clean install and that's why I was wiping everything out. Also, there seems to be no good reason for us to lock dependencies in the driver. New line is cool, maybe we can exclude that file from the linter scope to make sure it is as is though?

PR #80, opened 2020-07-22
tests(testserver): added support for SSL

mxschmitt:


tests/server.py
```diff
@@ -59,7 +70,7 @@ def start(self):
         self.csp = csp
         self.routes = routes
         self.gzip_routes = gzip_routes
-        static_path = os.path.join(os.path.dirname(__file__), "assets")
+        static_path = _dirname / "assets"
```
mxschmitt, line 73: The `pathlib` (Python standard) is nice and (I heard best practise nowadays). Since `/` will in the end also just call an underscore method (`__truediv__`) it has a great developer experience and feels natural.

PR #82, opened 2020-07-23
test: add test_navigation.py (part 1)

pavelfeldman:
drive-by replacing `asyncio.ensure_future` since "create_task() function is the preferred way for creating new Tasks."

### DIFF
--- a/tests/test_navigation.py
+++ b/tests/test_navigation.py
@@ -1,0 +1,471 @@
+# Copyright (c) Microsoft Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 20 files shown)


tests/test_navigation.py
```diff
@@ -0,0 +1,471 @@
...
+
+async def test_goto_should_fail_when_main_resources_failed_to_load(
+    page, server, is_chromium, is_webkit, is_win
+):
+    with pytest.raises(Error) as exc_info:
+        await page.goto("http://localhost:44123/non-existing-url")
+    if is_chromium:
+        assert "net::ERR_CONNECTION_REFUSED" in exc_info.value.message
+    elif is_webkit and is_win:
+        assert 'Couldn"t connect to server' in exc_info.value.message
```
mxschmitt, line 248:
```suggestion
        assert 'Couldn"t connect to server' in exc_info.value.message
```
```suggestion
        assert 'Couldn't connect to server' in exc_info.value.message
```

mxschmitt: Probably the reason why the bots are failing.

PR #83, opened 2020-07-23
feat(roll): roll Playwright to 1.2.0-next.1595486491367

mxschmitt:


playwright/network.py
```diff
@@ -53,7 +53,10 @@ def method(self) -> str:
 
     @property
     def postData(self) -> Optional[str]:
-        return self._initializer["postData"]
+        b64_content = self._initializer.get("postData")
```
pavelfeldman, line 56: We should also expose postDataBuffer as in TS, returning bytes.

PR #87, opened 2020-07-23
enh(sync): event class mapping to sync instances

mxschmitt:


playwright/sync_base.py
```diff
@@ -34,14 +50,15 @@ def __init__(
...
-        self._future = asyncio.create_task(
+        self._future = loop.create_task(
             wait_helper.wait_for_event(sync_base._async_obj, event, predicate)
         )
 
     @property
     def value(self) -> Any:
         if not self._value:
-            self._value = loop.run_until_complete(self._future)
+            value = loop.run_until_complete(self._future)
```
pavelfeldman, line 60: We have some events with the primitive type values.

Issue #21, opened 2020-07-08
Install browsers upon pip install, not the first launch


pavelfeldman:
That improves user experience

mxschmitt:
Made some research. There is no good native way of having a postinstall script. It could be implemented like that with egg_info which is super hacky, not sure how it performs: https://stackoverflow.com/a/45021666/6512681

pavelfeldman:
O actually made it work as described here: https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools

But pip is suppressing the I/O, so there are no progress bars visible, which defeats the purpose of this UX improvement.

mxschmitt:
I thought about the following design when we compare it with other tools like .

- They have a command line interface which they use to download files. In their case language models `spacy download en_core_web_sm`
- It can be exposed via `setup.py` like package.json>bin in the Node.js world
- If no browsers are persistent, I would download the used browser on launch() instead of all browsers

So the user can decide, either use the CLI to download the browser or the used browser get's downloaded on launch.

pavelfeldman:
That's probably it, we'll have the `plagwright install` for now

ColdHeat:
Thanks for working playwright, it's much needed for automated tests. 

I personally think that I would rather have browsers installed during the `pip install` step even if it didn't have progress bars. 

I looked around for a way to automatically install browsers on first test run perhaps and it looks like you would have to modify `sys.argv` to use the existing `main()` install.

https://github.com/microsoft/playwright-python/blob/fc5485c6057da52d80e5aa9ddcba1fb71e15b2ab/playwright/main.py#L123-L133

Could the install function perhaps be extracted this into a separate function that can be called without modifying `sys.argv`?

PR #102, opened 2020-07-28
feat(install): introduce explict post-installation step

pavelfeldman:
Fix for https://github.com/microsoft/playwright-python/issues/21

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,8 @@ jobs:
         PKG_CACHE_PATH: ${{ steps.node-pkg-cache.outputs.dir }}
     - name: Build package
       run: python build_package.py
+    - name: Install
+      run: python -m playwright install
     - name: Test
       run: pytest -vv --browser=${{ matrix.browser }} --junitxml=junit/test-results-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.browser }}.xml --cov=playwright --cov-report xml
     - name: Coveralls
--- a/playwright/helper.py
+++ b/playwright/helper.py
@@ -243,3 +243,17 @@ class RouteHandlerEntry:
     def __init__(self, matcher: URLMatcher, handler: RouteHandler):
         self.matcher = matcher
         self.handler = handler
+
+
+def not_installed_error(message: str) -> Exception:
+    return Exception(
+        f"""
+================================================================================
+{message}
+Please complete Playwright installation via running
+
+    "python -m playwright install"
+
+================================================================================
+"""
+    )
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -72,6 +72,8 @@ jobs:
         PKG_CACHE_PATH: ${{ steps.node-pkg-cache.outputs.dir }}
     - name: Build package
       run: python build_package.py
+    - name: Install
+      run: python -m playwright install
```
mxschmitt, line 76:
```suggestion
      run: playwright install
```
Since we do `pip install -e .` it should be available in the PATH.

playwright/helper.py
```diff
@@ -243,3 +243,14 @@ class RouteHandlerEntry:
     def __init__(self, matcher: URLMatcher, handler: RouteHandler):
         self.matcher = matcher
         self.handler = handler
+
+def raise_not_installed_error(message: str) -> None:
+    raise Exception(f"""
```
mxschmitt, line 248: nit: Probably I would prefer to create a new Error class which accepts in the constructor the browser name.

playwright/helper.py
```diff
@@ -243,3 +243,14 @@ class RouteHandlerEntry:
...
+================================================================================
+{message}
+Please complete Playwright installation via running
+
+    "python -m playwright install"
```
mxschmitt, line 253: If we compare Playwright Python with Pytest, they also use just the `pytest` command instead of `python -m pytest`. So for that I would use playwright install too.

playwright/helper.py
mxschmitt: Ah might be irrelevant since it does not only accept browser names, idk.

playwright/main.py
```diff
@@ -116,3 +112,31 @@ async def __aexit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
     # Use ProactorEventLoop in 3.7, which is default in 3.8
     loop = asyncio.ProactorEventLoop()
     asyncio.set_event_loop(loop)
+
+def main():
+    if 'install' not in sys.argv:
+        print ('Run "python -m playwright install" to complete installation')
```
mxschmitt, line 118:
```suggestion
        print ('Run "playwright install" to complete installation')
```

.github/workflows/ci.yml
pavelfeldman: The fact that we pollute env with another executable in users's path makes me sad. I'd like to emphasize that playwright is a library, not an executable and recommend its use via `-m`. Once we gain a handful of `cli` capabilities, I'll be comfortable saying `playwright` in my cli!

playwright/helper.py
pavelfeldman: Same as above, let's celebrate the discoverable execution once we have cli options.

PR #111, opened 2020-07-30
enh: error reporting in callbacks and events

### LINKED ISSUE #105: [BUG]: Handle somehow the case if an on_route handle will throw an exception
Not sure if its only about `on_route` events. Probably the same in other event handlers.

Referring to a Slack chat, solution would be to force exit the program with a message and a 1 exit code.

mxschmitt:
Full implementation is tbd.

Fixes #105

### DIFF
--- a/playwright/connection.py
+++ b/playwright/connection.py
@@ -15,6 +15,7 @@
 import asyncio
 import sys
 import traceback
+from asyncio import Future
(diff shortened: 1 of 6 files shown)


mxschmitt:
Closed for now, discussed via chat. Stdout is for now fine.

PR #115, opened 2020-07-31
chore: minor cleanup for coverage

mxschmitt:
> By default, Python uses utf-8 encoding.

Fixed also in the web server the unclosed file handlers.

### DIFF
--- a/playwright/network.py
+++ b/playwright/network.py
@@ -53,7 +53,7 @@ class Request(ChannelOwner):
         b64_content = self._initializer.get("postData")
         if not b64_content:
             return None
-        return base64.b64decode(bytes(b64_content, "utf-8")).decode()
+        return base64.b64decode(b64_content.encode()).decode()
 
     @property
     def headers(self) -> Dict[str, str]:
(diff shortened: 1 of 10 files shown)


playwright/network.py
```diff
@@ -53,7 +53,7 @@ def postData(self) -> Optional[str]:
         b64_content = self._initializer.get("postData")
         if not b64_content:
             return None
-        return base64.b64decode(bytes(b64_content, "utf-8")).decode()
+        return base64.b64decode(b64_content.encode()).decode()
```
pavelfeldman, line 56: Why do we need to encode base64 string into utf8 if b64decode accepts ascii?

playwright/page.py
```diff
@@ -746,9 +746,21 @@ def expect_request(
         return EventContextManagerImpl(self, "request", predicate, timeout)
 
     def expect_response(
-        self, predicate: Callable[[Response], bool] = None, timeout: int = None,
+        self,
+        url: URLMatch = None,
```
pavelfeldman, line 750: Why no love to expect_request?

playwright/network.py
mxschmitt: Probably `base64.b64decode(b64_content).decode()` is enough?

It was more the thought that we have to pass bytes to b64decode.
- b64_content -> str
- b64_content.encode() -> bytes

But seems like a str is also fine: https://docs.python.org/3/library/base64.html#base64.b64decode

PR #119, opened 2020-08-01
devops: separate wheels for OS/drivers

### LINKED ISSUE #117: [Feature] Provide separate wheels for the different operating systems
Only pack the corresponding driver into the wheel for the OS to reduce the size of the installed package instead of install all the drivers on all the operating systems.

Should be possible.

mxschmitt:
Closes #117

Before:
```
(env) ➜  playwright-python git:(devops/publish-os-specific) du -sh dist/*
 89M    dist/playwright-0.0.4.post9+g875b380-py3-none-any.whl
 81M    dist/playwright-0.0.4.post9+g875b380.tar.gz
```
After:
```
(env) ➜  playwright-python git:(devops/publish-os-specific) ✗ du -sh dist/*          
 24M    dist/playwright-0.0.4.post9+g875b380.d20200801-py3-none-macosx_10_13_x86_64.whl
 26M    dist/playwright-0.0.4.post9+g875b380.d20200801-py3-none-manylinux1_x86_64.whl
 24M    dist/playwright-0.0.4.post9+g875b380.d20200801-py3-none-win_amd64.whl
300K    dist/playwright-0.0.4.post9+g875b380.d20200801.tar.gz
```

PIP will automatically in the end install the correct wheel since all were uploaded to the PyPi registry.

### DIFF
--- a/build_package.py
+++ b/build_package.py
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import glob
+import os
 import shutil
 import subprocess
+import zipfile
 
 from playwright.path_utils import get_file_dirname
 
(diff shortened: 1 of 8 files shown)


build_package.py
```diff
@@ -29,3 +32,20 @@
     shutil.rmtree(_egg_dir)
 
 subprocess.run("python setup.py sdist bdist_wheel", shell=True)
+
+base_wheel_location = glob.glob("dist/*.whl")[0]
+without_platform = base_wheel_location[:-7]
+
+pack_data = [
+    ("driver-linux", "manylinux1_x86_64.whl"),
```
pavelfeldman, line 40: How does this work? What picks and extracts these files?

build_package.py
```diff
@@ -29,3 +32,20 @@
...
+    ("driver-macos", "macosx_10_13_x86_64.whl"),
+    ("driver-win.exe", "win_amd64.whl"),
+]
+
+for driver, wheel in pack_data:
+    wheel_location = without_platform + wheel
+    shutil.copy(base_wheel_location, wheel_location)
+    with zipfile.ZipFile(wheel_location, "a") as zipf:
+        driver_location = f"playwright/drivers/{driver}.gz"
```
pavelfeldman, line 49: We no longer need intermediate gzip then, the only reason behind it was to not extract stray platform files, now we don't have them. So this and main.py can change.

PR #122, opened 2020-08-01
tests: added missing interception tests

mxschmitt:


playwright/network.py
```diff
@@ -97,25 +99,40 @@ async def abort(self, errorCode: str = None) -> None:
...
-                headers = {}
-            headers["Content-Type"] = contentType
-            del params["contentType"]
-        if headers:
-            params["headers"] = serialize_headers(headers)
+        length = 0
         if isinstance(body, str):
             params["body"] = body
             params["isBase64"] = False
+            length = len(body)
```
pavelfeldman, line 113: This does not look right - body will be utf-8 encoded, so the length should be byte-length of the encoded string. I wonder if we want to do that on backend (or even in the browser) instead.

playwright/network.py
```diff
@@ -97,25 +99,40 @@ async def abort(self, errorCode: str = None) -> None:
...
         elif isinstance(body, bytes):
             params["body"] = base64.b64encode(body).decode()
             params["isBase64"] = True
+            length = len(body)
+        elif path:
+            file_content = Path(path).read_bytes()
+            params["body"] = base64.b64encode(file_content).decode()
```
pavelfeldman, line 120: This Python contract is weird. I'd explicitly say decode('ascii') to win the reader's trust.

playwright/network.py
```diff
@@ -97,25 +99,40 @@ async def abort(self, errorCode: str = None) -> None:
...
+        if params.get("contentType"):
+            headers["content-type"] = params["contentType"]
+        elif path:
+            headers["content-type"] = (
+                mimetypes.guess_type(str(Path(path)))[0] or "application/octet-stream"
+            )
+        if length and "content-length" not in headers:
+            headers["content-length"] = str(length)
+        params["headers"] = serialize_headers(headers)
+        if "path" in params:
```
pavelfeldman, line 134: nit: move up to L130-131

PR #126, opened 2020-08-02
tests: add focus tests

rwoll:
This ports the JS tests nearly 1:1 with the exception of
variables used with `exposeFunction`. In the JS test `focused`
and `blurred` are booleans but in the Python tests they are lists
that get appended to in the lambda since you cannot assign in lambda
as of 3.7.

### DIFF
--- a/tests/async/test_focus.py
+++ b/tests/async/test_focus.py
@@ -1,0 +1,112 @@
+# Copyright (c) Microsoft Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


tests/async/test_focus.py
```diff
@@ -0,0 +1,112 @@
...
+    assert await page.evaluate("() => document.activeElement.id") == "input-2"
+    await page.keyboard.press("Alt+Shift+Tab")
+    assert await page.evaluate("() => document.activeElement.id") == "link"
+    await page.keyboard.press("Alt+Shift+Tab")
+    assert await page.evaluate("() => document.activeElement.id") == "button"
+    await page.keyboard.press("Alt+Shift+Tab")
+    assert await page.evaluate("() => document.activeElement.id") == "input-1"
+
+
+#   });
```
pavelfeldman, line 111: nit: stray comment

PR #128, opened 2020-08-02
tests: added geolocation tests

mxschmitt:


playwright/async_api.py
```diff
@@ -4922,10 +4924,11 @@ async def setGeolocation(self, geolocation: typing.Dict) -> NoneType:
 
         Parameters
         ----------
-        geolocation : Optional[typing.Dict]
```
pavelfeldman, line 4925: Let's try restoring parameter docs here?

mxschmitt: good catch. Added a fallback for such cases.

playwright/async_api.py
```diff
@@ -4913,7 +4913,9 @@ async def clearPermissions(self) -> NoneType:
         """
         return mapping.from_maybe_impl(await self._impl_obj.clearPermissions())
 
-    async def setGeolocation(self, geolocation: typing.Dict) -> NoneType:
+    async def setGeolocation(
+        self, latitude: int, longitude: int, accuracy: int = None
+    ) -> NoneType:
         """BrowserContext.setGeolocation
 
         Sets the context's geolocation. Passing `null` or `undefined` emulates position unavailable.
```
pavelfeldman, line 4921: `Passing null or undefined emulates position unavailable` - how is this working now?

mxschmitt: Probably I have to revert the changes so that its possible again to specify None for the geolocation argument.

pavelfeldman: @dgozman was saying it did not necessarily work due to caching in the browsers, but I'd say staying consistent with upstream is a good idea for now.

PR #129, opened 2020-08-02
chore: test cleanups

rwoll:
A small follow up MR for my last contributions.

### DIFF
--- a/tests/async/test_dialog.py
+++ b/tests/async/test_dialog.py
@@ -52,36 +52,30 @@ async def test_should_allow_accepting_prompts(page: Page, server):
 
 async def test_should_dismiss_the_prompt(page: Page, server):
     result = []
-
-    def on_dialog(dialog: Dialog):
-        result.append(True)
-        asyncio.create_task(dialog.dismiss())
-
-    page.on("dialog", on_dialog)
+    page.on(
+        "dialog",
+        lambda dialog: (result.append(True), asyncio.create_task(dialog.dismiss())),
+    )
     assert await page.evaluate("prompt('question?')") is None
     assert result
 
 
 async def test_should_accept_the_confirm_prompt(page: Page, server):
     result = []
-
-    def on_dialog(dialog: Dialog):
-        result.append(True)
-        asyncio.create_task(dialog.accept())
-
-    page.on("dialog", on_dialog)
+    page.on(
+        "dialog",
+        lambda dialog: (result.append(True), asyncio.create_task(dialog.accept())),
+    )
     assert await page.evaluate("confirm('boolean?')") is True
     assert result
 
 
 async def test_should_dismiss_the_confirm_prompt(page: Page, server):
     result = []
-
-    def on_dialog(dialog: Dialog):
-        result.append(True)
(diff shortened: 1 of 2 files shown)


tests/async/test_dialog.py
```diff
@@ -22,64 +21,69 @@
...
-    async def on_dialog(dialog: Dialog):
+    def on_dialog(dialog: Dialog):
+        result.append(True)
         assert dialog.type == "prompt"
         assert dialog.defaultValue == "yes."
         assert dialog.message == "question?"
-        on_dialog_actions_complete.set_result(await dialog.accept("answer!"))
+        asyncio.create_task(dialog.accept("answer!"))
 
     page.on("dialog", lambda dialog: asyncio.create_task(on_dialog(dialog)))
```
mxschmitt, line 48: You are creating two asyncio tasks once the dialog event will be received. Only one would be needed. We try to only wrap as less functions inside a task. So I would do it like shown here.
```suggestion
    page.on("dialog", on_dialog)
```

tests/async/test_dialog.py
```diff
@@ -22,64 +21,69 @@
...
-    on_dialog_actions_complete: Future[None] = asyncio.Future()
+    result = []
 
-    async def on_dialog(dialog: Dialog):
-        on_dialog_actions_complete.set_result(await dialog.dismiss())
+    def on_dialog(dialog: Dialog):
+        result.append(True)
+        asyncio.create_task(dialog.dismiss())
 
     page.on("dialog", lambda dialog: asyncio.create_task(on_dialog(dialog)))
```
mxschmitt, line 60: Would be also two tasks and a single lambda would clean it up.

tests/async/test_dialog.py
```diff
@@ -22,64 +21,69 @@
...
```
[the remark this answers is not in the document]
rwoll: Fixed in latest commit throughout. Thank you!

tests/async/test_dialog.py
rwoll: 🤦 Thank you! That had meant to go with these cleanups! Fixed in latest commit—at least it should be :)

rwoll:
@mxschmitt Thanks for the feedback! Feel free to take another quick pass!

rwoll:
With the latest `lambda` rewrite I'm running into `mypy` linter:

```
tests/async/test_dialog.py:57: error: "append" of "list" does not return a value
tests/async/test_dialog.py:67: error: "append" of "list" does not return a value
tests/async/test_dialog.py:77: error: "append" of "list" does not return a value
```

rwoll:
@mxschmitt Whats your mitigation suggestion for the mypy error? The non-lambda code did not have that issue since it was a full function, could be multiline, didn't have to return everything etc.

mxschmitt:
> @mxschmitt Whats your mitigation suggestion for the mypy error? The non-lambda code did not have that issue since it was a full function, could be multiline, didn't have to return everything etc.

Then go with the multi line approach, should be fine for now. In the meantime I try to understand the mypy error.

PR #130, opened 2020-08-02
test: add missing dispatch event tests 

rwoll:


tests/async/test_dispatch_event.py
```diff
@@ -0,0 +1,152 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+async def test_should_dispatch_click_event(page, server):
+    await page.goto(server.PREFIX + '/input/button.html')
+    await page.dispatchEvent('button', 'click')
```
rwoll, line 18: This currently fails 💥 due to (I think) a serialization error: `eventInit: expected object, got null`. 

Full trace:

```
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_should_dispatch_click_event[chromium] ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

page = <playwright.async_api.Page object at 0x1040f14e0>, server = http://localhost:60996

    async def test_should_dispatch_click_event(page, server):
        await page.goto(server.PREFIX + '/input/button.html')
>       await page.dispatchEvent('button', 'click')

tests/async/test_dispatch_event.py:18: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
playwright/async_api.py:3196: in dispatchEvent
    selector=selector, type=type, eventInit=eventInit, timeout=timeout
playwright/page.py:300: in dispatchEvent
    return await self._main_frame.dispatchEvent(**locals_to_params(locals()))
playwright/frame.py:241: in dispatchEvent
    "dispatchEvent", dict(selector=selector, type=type, eventInit=eventInit)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <playwright.connection.Channel object at 0x1040f1710>, method = 'dispatchEvent', params = {'eventInit': None, 'selector': 'button', 'type': 'click'}

    async def send(self, method: str, params: dict = None) -> Any:
        if params is None:
            params = {}
>       result = await callback.future
E       playwright.helper.Error: ('eventInit: expected object, got null', 'Error: eventInit: expected object, got null\n    at /snapshot/driver/node_modules/playwright/lib/rpc/validatorPrimitives.js:73:19\n    at /snapshot/driver/node_modules/playwright/lib/rpc/validator.js:30:20\n    at Object.FrameDispatchEventParams (/snapshot/driver/node_modules/playwright/lib/rpc/validatorPrimitives.js:78:27)\n    at DispatcherConnection._validateParams (/snapshot/driver/node_modules/playwright/lib/rpc/server/dispatcher.js:114:32)\n    at DispatcherConnection.dispatch (/snapshot/driver/node_modules/playwright/lib/rpc/server/dispatcher.js:136:36)\n    at Transport.onmessage (/snapshot/driver/node_modules/playwright/lib/rpc/server.js:35:55)\n    at Immediate._onImmediate (/snapshot/driver/node_modules/playwright/lib/rpc/transport.js:64:26)\n    at processImmediate (internal/timers.js:456:21)')

playwright/connection.py:39: Error
```

rwoll: here's the definition from protocol:

```
    dispatchEvent:
      parameters:
        type: string
        eventInit: SerializedArgument
```

rwoll: @pavelfeldman Should the type of `eventInit` be changed to optional (`SerializedArgument?`) or should we not be getting null here?

tests/async/test_dispatch_event.py
```diff
@@ -0,0 +1,152 @@
...
+    assert await page.evaluate('() => result)') == 'Clicked'
```
rwoll, line 19:
```suggestion
    assert await page.evaluate('() => result') == 'Clicked'
```

tests/async/test_dispatch_event.py
pavelfeldman: `dispatchEvent` should serialize argument to conform to the protocol:

```py
async def dispatchEvent(self, type: str, eventInit: Dict = None) -> None:
        await self._channel.send("dispatchEvent", dict(type=type, eventInit=serialize_argument(eventInit)))
```

rwoll: Thank you! I completed the impl in frame and element handle.

tests/async/test_dispatch_event.py
```diff
@@ -0,0 +1,174 @@
...
+        }""",
+    )
+    await page.setContent('<div onclick="window._clicked=true">Hello</div>')
+    await page.dispatchEvent("dispatchEvent=div", "click")
+    assert await page.evaluate("() => window._clicked")
+
+
+@pytest.mark.skip("TODO: Determine why there is a failure.")
+@pytest.mark.skip_browser("webkit")
+async def test_should_dispatch_drag_drop_events(page, server):
```
rwoll, line 142: This test and the below test are failing hence marked as skip; I have not yet looked into the reason for failure. The test executes but the final assertion fails. I suspect it probably has to with `dataTransfer`

rwoll: Looking at prints from dispatchEvent:

```
print(eventInit)
print(serialize_argument(eventInit))
```

I'm seeing:

```
{<playwright.async_api.JSHandle object at 0x10ea51c50>}
{'value': {'v': 'undefined'}, 'handles': []}
```

So I'm definitely not serializing the JSHandle properly before sending it over the channel.

rwoll: I fixed the accidental set notation, but still seeing a failure to serialize the handle:

```
{'dataTransfer': <playwright.async_api.JSHandle object at 0x10b108d68>}
{'value': {'o': [{'k': 'dataTransfer', 'v': {'v': 'undefined'}}]}, 'handles': []}
```

Stepping into the serialization code now to see why

```
<class 'playwright.async_api.JSHandle'>
```

is not matching the JSHandle check.

rwoll: @pavelfeldman What do you expect `print(await page.evaluateHandle("() => new DataTransfer()"))` to be? Trying to narrow down whether I have the correct thing that's not getting serialized in the eventInit serialization.

mxschmitt: Great finding!

As a workaround you can use for now (add it before the isinstance):
```py
    if type(value).__name__ == "JSHandle" and not isinstance(value, JSHandle):
        value = value._impl_obj
```
In the meantime we'll see if we can find a better solution for that.

pavelfeldman: There is no DataTransfer constructor in WebKit and that's why these tests are marked as failing upstream with `it.fail(WEBKIT)`.  Does this answer your question of there is something I'm not following?

pavelfeldman: Ah, you are saying it does not work on Chromium also. Looking!

pavelfeldman: @rwoll, so I am late to the party, Max already suggested something. We generate sync and async API layers on top of the core implementation in Python and `isinstance` is checking for the internal type, while user passes the wrapper. The solution is to traverse all the incoming dictionaries recursively and translate wrapper types into regular types. I'll send a PR for that!

pavelfeldman: Also, for debugging, run `DEBUGP=1 pytest --browser=chromium -s` to see the protocol log

pavelfeldman: Fixed in https://github.com/microsoft/playwright-python/pull/136/files

rwoll: > We generate sync and async API layers on top of the core implementation in Python and isinstance is checking for the internal type, while user passes the wrapper.

This is what I was tripping on. Couldn't quite make sense of the fully qualified type `playwright.async_api.JSHandle` when I didn't see it in the source. Thanks both for the suggested workaround and the more thorough fix. 

@pavelfeldman Once you MR lands I'll rebase and remove the more hacky workaround. 🎉

PR #132, opened 2020-08-02
tests: migrated some async tests over to sync

mxschmitt:
Failing bots are unrelated.


pavelfeldman:
This looks great. Any idea how we can track which tests inside async have been converted to sync?

mxschmitt:
> This looks great. Any idea how we can track which tests inside async have been converted to sync?

By filename or better by method name.

PR #146, opened 2020-08-05
tests: added keyboard tests

mxschmitt:


tests/async/test_headful.py
```diff
@@ -34,6 +34,7 @@ async def test_headless_should_be_able_to_read_cookies_written_by_headful(
 ):
     if is_chromium and is_win:
         pytest.skip("see https://github.com/microsoft/playwright/issues/717")
+        return
```
rwoll, line 37: I need to add the return in the two cookie tests from yesterday. Good catch!

tests/async/conftest.py
```diff
@@ -17,6 +17,12 @@
 from playwright import async_playwright
 
 
+# Will mark all the tests as async
+def pytest_collection_modifyitems(items):
+    for item in items:
+        item.add_marker(pytest.mark.asyncio)
+
```
rwoll, line 24: Can you explain to me why this is needed? (I'm new to async in Python.)

playwright/element_handle.py
```diff
@@ -151,7 +151,7 @@ async def type(
         timeout: int = None,
         noWaitAfter: bool = None,
     ) -> None:
-        await self._channel.send("text", locals_to_params(locals()))
+        await self._channel.send("type", locals_to_params(locals()))
```
pavelfeldman, line 154: I'm so happy porting tests reveal this.

tests/async/conftest.py
mxschmitt: Before it was in the global `conftest.py`. A `conftest.py` will be executed through the directory three and by that they can overwrite each other (which is nearest to the actual test).

This specific piece added asyncio support to the tests which was before available in all the tests. But since we split async and sync I limited the scope of it.

tests/async/test_headful.py
mxschmitt: Actually I'm not sure if we need it! It was flaky so I added it. But seems like its not needed: https://www.programcreek.com/python/example/25423/pytest.skip

rwoll: ✨ Magic!

tests/async/conftest.py
rwoll: Thanks! So async tests need to be explicitly (or programmatically) marked as async for pytest to run them?

mxschmitt: exactly with: https://github.com/pytest-dev/pytest-asyncio

You can do it manually with the decorator per test or in our case just for all the tests in that directory.

Issue #144, opened 2020-08-05
Implement missing APIs


mxschmitt:
- [x] Method not implemented: Download.createReadStream <- decided to skip for now
- [x] Method not implemented: Download.saveAs 
- [x] Method not implemented: Request.postDataBuffer
- [x] Method not implemented: Request.postDataJSON

PR #147, opened 2020-08-05
feat: added Download.saveAs(path)

mxschmitt:
Relates #144

### DIFF
--- a/playwright/async_api.py
+++ b/playwright/async_api.py
@@ -2939,6 +2939,18 @@ class Download(AsyncBase):
         """
         return mapping.from_maybe_impl(await self._impl_obj.path())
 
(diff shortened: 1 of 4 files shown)


rwoll:
Based on the failures, it looks like we may need to roll to a later playwright (JS) version here since the saveAs we updated saveAs create missing directory behavior in https://github.com/microsoft/playwright/pull/3249.

mxschmitt:
> Based on the failures, it looks like we may need to roll to a later playwright (JS) version here since we updated saveAs create missing directory behavior in [microsoft/playwright#3249](https://github.com/microsoft/playwright/pull/3249).

thx will do 👍

PR #150, opened 2020-08-05
feat: added Request.postDataBuffer

mxschmitt:
Relates #144

### DIFF
--- a/playwright/network.py
+++ b/playwright/network.py
@@ -59,10 +59,10 @@ class Request(ChannelOwner):
 
     @property
     def postData(self) -> Optional[str]:
-        b64_content = self._initializer.get("postData")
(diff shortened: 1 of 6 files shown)


playwright/network.py
```diff
@@ -76,6 +76,13 @@ def postDataJSON(self) -> Optional[Dict]:
             return dict(parse.parse_qsl(post_data))
         return json.loads(post_data)
 
+    @property
+    def postDataBuffer(self) -> Optional[bytes]:
+        b64_content = self._initializer.get("postData")
```
pavelfeldman, line 81: Use it in postData ?

mxschmitt: makes sense

PR #151, opened 2020-08-05
feat: Download.createReadStream()

mxschmitt:
I think thats the "idiomatic python way" of that. works basically like [`json.dump`](https://docs.python.org/3/library/json.html#json.dump).

The `io` type generation handling was not working as expected, maybe you can take a look? Do I need to replace `_io` by `io` in our generation scripts?

Relates #144

### DIFF
--- a/playwright/async_api.py
+++ b/playwright/async_api.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+import io
 import pathlib
 import sys
 import typing
@@ -2761,6 +2761,7 @@ class Selectors(AsyncBase):
             Name that is used in selectors as a prefix, e.g. `{name: 'foo'}` enables `foo=myselectorbody` selectors. May only contain `[a-zA-Z0-9_]` characters.
         source : Optional[str]
             Script that evaluates to a selector engine instance.
+        path : Optional[str]
         contentScript : Optional[bool]
             Whether to run this selector engine in isolated JavaScript environment. This environment has access to the same DOM, but not any JavaScript objects from the frame's scripts. Defaults to `false`. Note that running as a content script is not guaranteed when this engine is used together with other registered engines.
         """
@@ -2950,6 +2951,26 @@ class Download(AsyncBase):
             Path where the download should be saved.
         """
         return mapping.from_maybe_impl(await self._impl_obj.saveAs(path=path))
+
+    async def createReadStream(
+        self, fp: io.BytesIO = None, size: int = None
+    ) -> typing.Union[io.BytesIO, NoneType]:
+        """Download.createReadStream
+
+        Returns readable stream for current download or `null` if download failed.
+
+        Parameters
+        ----------
+        fp : Optional[io.BytesIO]
+        size : Optional[int]
+
+        Returns
+        -------
+        Optional[io.BytesIO]
+        """
+        return mapping.from_maybe_impl(
+            await self._impl_obj.createReadStream(fp=fp, size=size)
+        )
 
 
 mapping.register(DownloadImpl, Download)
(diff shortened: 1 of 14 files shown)


playwright/async_api.py
```diff
@@ -2928,6 +2928,22 @@ async def path(self) -> typing.Union[str, NoneType]:
...
         return mapping.from_maybe_impl(await self._impl_obj.path())
 
+    async def createReadStream(
+        self, fp: _io.BytesIO = None
+    ) -> typing.Union[_io.BytesIO, NoneType]:
+        """Download.createReadStream
+
+        Returns readable stream for current download or `null` if download failed.
+
+        Parameters
```
pavelfeldman, line 2938: No parameters because we could not find fp in api.json. We should still list those w/o docs...

playwright/stream.py
```diff
@@ -0,0 +1,37 @@
...
+        self, parent: ChannelOwner, type: str, guid: str, initializer: Dict
+    ) -> None:
+        super().__init__(parent, type, guid, initializer)
+
+    async def stream(self, fp: io.BytesIO) -> io.BytesIO:
+        if not fp:
+            fp = io.BytesIO()
+
+        while True:
+            result = await self._channel.send("read", dict(size=1))
```
pavelfeldman, line 33: Are you reading single byte at a time? Probably not too efficient!

mxschmitt: Oh ye good catch you won! Currently its not possible to leave out the size, thats why I set it to Node.js default value.

Probably we have to mark it in the protocol as optional so Node.js alone handles the default value?

playwright/sync_api.py
```diff
@@ -3077,6 +3078,26 @@ def saveAs(self, path: typing.Union[pathlib.Path, str]) -> NoneType:
...
+        Returns readable stream for current download or `null` if download failed.
+
+        Parameters
+        ----------
+        fp : Optional[<class 'io.BytesIO'>]
+        size : Optional[int]
+
+        Returns
+        -------
+        Optional[<class 'io.BytesIO'>]
```
pavelfeldman, line 3095: The name isn't very nice

playwright/stream.py
```diff
@@ -0,0 +1,39 @@
...
+from playwright.connection import ChannelOwner
+
+
+class Stream(ChannelOwner):
+    def __init__(
+        self, parent: ChannelOwner, type: str, guid: str, initializer: Dict
+    ) -> None:
+        super().__init__(parent, type, guid, initializer)
+
+    async def stream(self, fp: io.BytesIO, size: Optional[int]) -> io.BytesIO:
```
pavelfeldman, line 28: I don't understand the version that returns the stream. It looks like it'll create one on line 30 and then read all content into its buffer? So there isn't much streaming going one?

pavelfeldman: Looks like you need to subclass https://docs.python.org/3/library/io.html#io.RawIOBase in your stream instead.

mxschmitt: True, we should probably rename it. But in the Python world there are no streams like in Node.js, thats why I used this approach.
Could be used for writing into files or getting the content als a BytesIO stream.
Their documentation says for [io.BytesIO](https://docs.python.org/3/library/io.html#buffered-streams):
> A stream implementation using an in-memory bytes buffer. It inherits BufferedIOBase. The buffer is discarded when the close() method is called.

If we have another name we don't have documentation on the other hand, I would keep the name probably.

scripts/documentation_provider.py
```diff
@@ -50,6 +50,46 @@ def __init__(self) -> None:
         self.printed_entries: List[str] = []
         with open("api.json") as json_file:
             self.api = json.load(json_file)
+        self.api["StreamIO"] = {
```
pavelfeldman, line 53: I don't think we need these - these classes are not a part of the API, they don't need to be exposed. Just a helper class that converts channel into the stream is seemingly all we need.

Issue #140, opened 2020-08-04
Download driver from CDN


mxschmitt:
- [x] move driver into upstream repo
- [ ] Setup driver CDN https://github.com/microsoft/playwright/issues/3503
- [ ] once `python -m playwright install` will get executed the driver gets downloaded first which installs then the browsers https://github.com/microsoft/playwright-python/pull/172

We internally just have a constant for the version which we are using

mxschmitt:
Relates #169 -> Wheel + source don't contain drivers anymore.

PR #152, opened 2020-08-06
fix: wheel integrity by using generated wheel

mxschmitt:
Reverts the old change which probably messed up the package integrity.

Relates to #140


mxschmitt:
> Does main.py need to extract the driver again?

Ye, have to add it.

mxschmitt:
Closed for now, lets see if users will face into it.

PR #163, opened 2020-08-08
feat: roll driver to 1.3.0-next.1596843106133, handle paths on the client side

dgozman:


tests/async/test_defaultbrowsercontext.py
```diff
@@ -57,7 +57,7 @@ async def test_context_cookies_should_work(server, launch_persistent):
             "expires": -1,
             "httpOnly": False,
             "secure": False,
-            "sameSite": "None",
+            "sameSite": "Lax" if is_firefox else "None",
```
pavelfeldman, line 60: Why did this change?

dgozman: Andrey says that Firefox has switched the defaults.

pavelfeldman:
We might need to update client then to ensure priper 🍪 roundtrip.

tests/async/test_defaultbrowsercontext.py
pavelfeldman: "There is currently no timeline to ship this feature to the release channel of Firefox.". Let's make sure to fix it upstream. Filing a bug

dgozman:
Should we postpone this until upstream is fixed?

mxschmitt:
> Should we postpone this until upstream is fixed?

Pavel was fine with it (via Slack).

PR #180, opened 2020-08-26
feat(roll): roll Playwright 1.3.0-next.1598476714110

mxschmitt:
I'm quite stuck with the dispatcher logic, not sure how to proceed there since it seems like rpc client and server share some logic upstream.


playwright/cdp_session.py
```diff
@@ -29,9 +29,9 @@ def _on_event(self, params: Any) -> None:
         self.emit(params["method"], parse_result(params["params"]))
 
     async def send(self, method: str, params: Dict = None) -> Dict:
-        payload = {"method": method}
+        payload: Dict[str, Any] = {"method": method}
         if params:
-            payload["params"] = serialize_argument(params)["value"]
+            payload["params"] = params
```
dgozman, line 34: nit: I guess locals_to_params() should work.

tests/async/test_browsercontext.py
```diff
@@ -429,7 +429,7 @@ async def test_route_should_intercept(context, server):
     def handle(route, request):
         intercepted.append(True)
         assert "empty.html" in request.url
-        assert request.headers["user-agent"]
+        assert request.headers["User-Agent"]
```
dgozman, line 432: We have to do the lowercasing when converting header to Dict. Something like this: https://github.com/microsoft/playwright/blob/master/src/utils/utils.ts#L121-L126

tests/async/test_launcher.py
```diff
@@ -102,142 +85,6 @@ async def test_browser_type_name_should_work(
...
-):
-    browser_server = await browser_type.launchServer(**launch_arguments)
-    remote = await browser_type.connect(wsEndpoint=browser_server.wsEndpoint)
-    assert remote.isConnected()
-    await remote.close()
-    assert remote.isConnected() is False
-    await browser_server.close()
-
-
-async def test_browser_is_connected_should_throw_when_used_after_isConnected_returns_false(
```
dgozman, line 116: Is it possible to skip these tests for now, so that we can restore coverage later?

PR #182, opened 2020-09-02
feat(roll): roll Playwright 1.3.0-next.1599061793983

mxschmitt:


pavelfeldman (approved):
Why do we need yarn lock?

mxschmitt:
> Why do we need yarn lock?

good catch, removed

PR #184, opened 2020-09-03
chore: add missing methods and use a set for it

mxschmitt:
I went through them, let me know if you are okay with that. In my mind there is no need for now to land more methods since they are almost only CR specific.

```
Method not implemented: Download.createReadStream <- PR open

Method not implemented: BrowserContext.serviceWorkers <- CR specific
Method not implemented: BrowserContext.backgroundPages <- CR specific
Method not implemented: Browser.newBrowserCDPSession <- important?

Method not implemented: BrowserType.launchServer <- decided to not land this in RPC clients. By that also the following methods

Method not implemented: BrowserServer.close
Method not implemented: BrowserServer.process
Method not implemented: BrowserServer.kill
Method not implemented: BrowserServer.wsEndpoint

Method not implemented: BrowserContext.newCDPSession <- its there but on the CR browser context
Method not implemented: Logger.isEnabled
Method not implemented: Logger.log
Method not implemented: Browser.stopTracing
Method not implemented: Browser.startTracing
Method not implemented: Page.coverage <- CR only
Method not implemented: Coverage.stopCSSCoverage
Method not implemented: Coverage.startCSSCoverage
Method not implemented: Coverage.startJSCoverage
Method not implemented: Coverage.stopJSCoverage
Method not implemented: BrowserContext.setHTTPCredentials <- deprecated

```

### DIFF
--- a/playwright/async_api.py
+++ b/playwright/async_api.py
@@ -810,6 +810,15 @@ class ElementHandle(JSHandle):
     def __init__(self, obj: ElementHandleImpl):
         super().__init__(obj)
 
+    def toString(self) -> str:
+        """ElementHandle.toString
+
+        Returns
+        -------
+        str
+        """
+        return mapping.from_maybe_impl(self._impl_obj.toString())
+
     def asElement(self) -> typing.Union["ElementHandle", NoneType]:
         """ElementHandle.asElement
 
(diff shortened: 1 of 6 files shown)


pavelfeldman:
Agreed that the state we are in is acceptable. Let's add these into the "expected mismatches" list and fail when there is something new?

pavelfeldman (approved):

PR #207, opened 2020-09-30
devops: added Docker image

mxschmitt:
Fixes #197

Relates https://github.com/microsoft/playwright/issues/3791#issue-695697682


.github/workflows/test_docker.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
+    - name: Get Node PKG Cache dir
+      id: node-pkg-cache
+      run: |
+        node -e "const path = require('path'); const fs = require('fs'); const cachePath = path.join(os.tmpdir(), 'pkg-cache'); fs.mkdirSync(cachePath); console.log('::set-output name=dir::' + cachePath)"
```
aslushnikov, line 28: why not simply `mkdir -p /tmp/pkg-cache` ?

mxschmitt: We used it once it was cross-os compatible.

aslushnikov: Let's remove it then!

.dockerignore
```diff
@@ -0,0 +1,22 @@
+playwright/__pycache__/
```
aslushnikov, line 1: I'm afraid this `dockerignore` will rot over time. 

Instead of having one, you can put `Dockerfile` in an empty directory, e.g. `docker/Dockerfile.focal`, and use the directory as a docker context when building.

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+# 7. (Optional) Install XVFB if there's a need to run browsers in headful mode
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    xvfb
+
+# 8. Feature-parity with node.js base images.
+RUN apt-get update && apt-get install -y --no-install-recommends git ssh
+
+# === BAKE BROWSERS INTO IMAGE ===
+
+# 1. Add tip-of-tree Playwright Python package to install its browsers.
```
aslushnikov, line 64: do we want `pwuser` here?

.dockerignore
mxschmitt: Then its not possible to include files outside of the build context afaik.

aslushnikov (approved):

PR #209, opened 2020-10-02
fix(x86): bake x86 node for Windows

pavelfeldman:


mxschmitt (approved):

mxschmitt (approved):

playwright/main.py
```diff
@@ -134,9 +139,12 @@ def main() -> None:
     if "install" not in sys.argv:
         print('Run "python -m playwright install" to complete installation')
         return
-    package_path = get_file_dirname()
-    driver_name = compute_driver_name()
-    driver_executable = package_path / "drivers" / driver_name
+    driver_executable = compute_driver_executable()
+    # Fix the executable bit during the installation.
+    if not sys.platform == "win32":
```
mxschmitt, line 144: Really weird that this is needed now. Because it was working before right?

pavelfeldman: why do you think it is weird? we publish from windows where there is no such bit.

mxschmitt: okay you won. Didn't keep that in mind, now it makes sense. :+1:

PR #219, opened 2020-10-04
chore: Add py.typed

cbows:
This will allow type checkers to collect type information from this package.

See [PEP 561](https://www.python.org/dev/peps/pep-0561/).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include playwright/drivers/browsers.json
 include playwright/*.py
+include playwright/py.typed
 include README.md
 include SECURITY.md
 include LICENSE


pavelfeldman:
Does this also need a change to `setup.py` changes to include it in `package_data`?

cbows:
I added an entry to `MANIFEST.in`.

According to the [mypy documentation](https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages) it is also necessary to set `zip_safe=False`, which might already be the default for this package (see [setuptools documentation](https://setuptools.readthedocs.io/en/latest/userguide/miscellaneous.html?highlight=zip_safe#setting-the-zip-safe-flag)), so setting it explicitly is probably reasonable.

pavelfeldman (approved):

PR #222, opened 2020-10-05
fix(PermissionError): Make file executable before running the app. 

### LINKED ISSUE #221: Running playwright raises PermissionError on Linux
### Reproducing the error

- Created a virtual environment 
- Installed playwright with `python -m pip install playwright`

Tried to run some code

```python
from playwright import sync_playwright

with sync_playwright() as p:
    for browser_type in [p.chromium, p.firefox, p.webkit]:
        page = browser.newPage()
        page.goto('http://whatsmyuseragent.org/')
        browser.close()
```

Then it raised this error

```python
PermissionError: [Errno 13] Permission denied: '/home/leno/Desktop/open-source/pwright/env/lib/python3.8/site-packages/playwright/drivers/driver-linux
```

I think this is not a normal behavior since running Python under sudo is a terrible idea.

**OS**: Ubuntu 20.04
**Python Version**: 3.8.2
**Playwright Version**:  0.142.3

ycd:
Closes #221, 

### Reproducing the PR

By default driver comes without S_IEXEC symbol, now it gives file S_IEXEC symbol before running it.


mxschmitt (changes requested):
We already have such a logic here in place: https://github.com/microsoft/playwright-python/blob/82ac046daab8f42b016ae0afd1ffad81301c2538/playwright/main.py#L144-L147

And we should add there a condition that it will only get executed when its != win

ycd:
> We already have such a logic here in place:
> 
> https://github.com/microsoft/playwright-python/blob/82ac046daab8f42b016ae0afd1ffad81301c2538/playwright/main.py#L144-L147

Hmmm, that's interesting, it's raising the error before reaching the main. After doing `chmod +x` i saw this message.

```python
Run "python -m playwright install" to complete installation
```

pavelfeldman:
Your change looks good, but you should remove the `os.chmod(driver_executable, st.st_mode | stat.S_IEXEC)` part since it becomes redundant.

ycd:
Thanks for review @pavelfeldman, I removed that part now.

pavelfeldman (approved):

PR #227, opened 2020-10-10
chore: user the driver from cdn

pavelfeldman:


build_package.py
```diff
@@ -15,12 +15,27 @@
...
+import stat
 import subprocess
 import sys
 import zipfile
 
 from playwright.path_utils import get_file_dirname
 
+driver_version = "0.5.2"
+
+if not os.path.exists("driver"):
```
mxschmitt, line 27: Duplicate logic with the paths. I would use Pathlib for that.

build_package.py
```diff
@@ -15,12 +15,27 @@
...
+    os.makedirs("driver")
+if not os.path.exists("playwright/driver"):
+    os.makedirs("playwright/driver")
+
+for platform in ["mac", "linux", "win32", "win32_x64"]:
+    zip_file = f"playwright-cli-{driver_version}-{platform}.zip"
+    if not os.path.exists("driver/" + zip_file):
+        url = "https://playwright.azureedge.net/builds/cli/" + zip_file
+        print("Fetching ", url)
+        subprocess.check_call(["curl", url, "-o", "driver/" + zip_file])
```
mxschmitt, line 37: Let's do it in memory like here? https://github.com/microsoft/playwright-python/pull/172/files#diff-56e4178dc64fcedaf003454c830bc80cR61

build_package.py
```diff
@@ -37,24 +52,46 @@
...
+    if platform == "win32_x64":
+        wheel = "win_amd64.whl"
     wheel_location = without_platform + wheel
     shutil.copy(base_wheel_location, wheel_location)
-    from_location = f"driver/out/{driver}"
-    to_location = f"playwright/drivers/{driver}"
-    with zipfile.ZipFile(wheel_location, "a") as zipf:
-        zipf.write(from_location, to_location)
+    with zipfile.ZipFile(wheel_location, "a") as zip:
+        for file in os.listdir(f"driver/{platform}"):
```
mxschmitt, line 88: Got it. So driver will get still distributed over wheel. Then curl is fine, maybe we should download it on-demand which seems better (source code install would work).

build_package.py
pavelfeldman: Those are different folders or what do you mean?

build_package.py
pavelfeldman: Yes, we still bundle.

PR #275, opened 2020-11-03
fix: update api generator to support new api.json format

yury-s:


mxschmitt:
Unrelated to this PR, but wouldn't it make sense to use the api.json from the playwright-cli (spawn as subprocess with specific CLI flag, already there) instead of storing it inside the Git repository? I can follow up with this change if you think @pavelfeldman it makes sense.

yury-s:
> Unrelated to this PR, but wouldn't it make sense to use the api.json from the playwright-cli 

I think it would, it's always possible to restore api.json for given a version of cli so there is no much sense in storing another copy of it.

scripts/documentation_provider.py
```diff
@@ -78,13 +78,15 @@ def print_entry(
...
+        method = (
+            clazz["methods"].get(method_name)
+            or clazz["properties"].get(method_name)
+            or super_clazz["methods"].get(method_name)
         )
         fqname = f"{class_name}.{method_name}"
         indent = " " * 8
         print(f'{indent}"""{class_name}.{original_method_name}')
-        if method["comment"]:
+        if "comment" in method and method["comment"]:
```
pavelfeldman, line 89:
```suggestion
        if method.get("comment"):
```

scripts/documentation_provider.py
```diff
@@ -134,7 +136,7 @@ def print_entry(
                     code_type = self.serialize_python_type(value)
 
                     print(f"{indent}{original_name} : {code_type}")
-                    if doc_value["comment"]:
+                    if "comment" in doc_value and doc_value["comment"]:
```
pavelfeldman, line 139:
```suggestion
                    if doc_value.get("comment"):
```

scripts/documentation_provider.py
```diff
@@ -156,7 +158,7 @@ def print_entry(
             print("        Returns")
             print("        -------")
             print(f"        {self.serialize_python_type(value)}")
-            if method["returnComment"]:
+            if "returnComment" in method and method["returnComment"]:
```
pavelfeldman, line 161:
```suggestion
            if method.get("returnComment"):
```

pavelfeldman (approved):

PR #288, opened 2020-11-08
chore: remove api.json and use it from driver

### LINKED ISSUE #278: Remove api.json and use it from the driver as subprocess call

mxschmitt:
Closes #278


yury-s (commented):
Now that `api.json` is not stored in the repo there is no way to tell what version of playwright was used to generated python APIs. Let's include playwright-cli version into `*_apy.py` when generating APIs?

mxschmitt:
> Now that `api.json` is not stored in the repo there is no way to tell what version of playwright was used to generated python APIs. Let's include playwright-cli version into `*_apy.py` when generating APIs?

On the bots (GitHub Actions) we ensure that the generated script output is equal with the HEAD, so you can never make changes and forget to generate. By that we always know that the previous HEAD sync/async generated output was the generated with the version of the HEAD, in my mind its fine, or wdyt?

yury-s (approved):
Should be good enough then.

PR #293, opened 2020-11-11
fix(api): remove browserType.connect()

yury-s:


CONTRIBUTING.md
```diff
@@ -1,3 +1,33 @@
+# Contributing
+
+## How to Contribute
+
+### Configuring python environment
+
+The project required python version 3.8+. To set it as default in the environment run the following commands:
```
dgozman, line 7: typo: required -> requires

CONTRIBUTING.md
```diff
@@ -1,3 +1,33 @@
...
+
+Install required dependencies:
+
+```bash
+python -m pip install --upgrade pip wheel
+pip install -r local-requirements.txt
+pip install -e .
+```
+
+For more details look at the [CI configuration](https://github.com/microsoft/playwright-python/blob/master/.github/workflows/ci.yml).
```
dgozman, line 22: I think we can do a relative link here.

dgozman (approved):

CONTRIBUTING.md
yury-s: Done.

CONTRIBUTING.md
yury-s: Done.

vania-pooh:
@yury-s what is the expected way of directly connecting to web-socket having `connect()` method removed? Without such feature parallel remote execution is impossible.

yury-s:
We are going to implement connect at the Playwright level, i.e. the client would be able to do something like `new Playwright(<websocket address>` and from that point talk to the remote server same way as if it was running locally. Otherwise with the previous connect API we first need to establish connection to local playwright driver and then call connect on it passing url for each browser individually. This seemed sub-optimal.

vania-pooh:
@yury-s: any related issues or time estimates?

yury-s:
Just filed https://github.com/microsoft/playwright/issues/4687 We'll likely have it soon given that we are deprecating `connect()` but probably after 1.7 (which we is gonna be next week or so).

PR #299, opened 2020-11-14
devops: push Docker image to MCR

### LINKED ISSUE #290: Publish docker image to MCR
This would enable users to pull the image for their CI environments.

mxschmitt:
Closes #290

`DOCKER_PASSWORD` environment variable needs to be set.


mxschmitt:
I don't get it why its failing, maybe someone has an idea?

mxschmitt:
@pavelfeldman not sure why its failing, only the video test, hmm.

kumaraditya303:
> I don't get it why its failing, maybe someone has an idea?

@mxschmitt It is failing because of timeout, remove timeout and all the tests would pass.

.github/workflows/publish_canary_docker.yml
```diff
@@ -0,0 +1,43 @@
+name: "devrelease:docker"
```
pavelfeldman, line 1: Curious, why do we need this one?

mxschmitt: We can also remove the canary builds. They were useful for testing and Lushnikiv and me used them a lot so I think it brings value.

aslushnikov (approved):

aslushnikov (changes requested):
Marking as "request changes" to not merge accidentally since this the tests are actually failing in the container.

@mxschmitt is looking into this

aslushnikov (approved):

PR #305, opened 2020-11-17
chore: roll Playwright to 0.170.0-next.1605573954344

pavelfeldman:


yury-s (approved):

mxschmitt (approved):

playwright/helper.py
```diff
@@ -105,9 +115,10 @@ class Header(TypedDict):
 
 
 class ContinueParameters(TypedDict, total=False):
-    method: str
-    headers: List[Header]
-    postData: str
+    url: Optional[str]
```
mxschmitt, line 118: since total=False all keys inside the dict are optional, so  afaik you don't have to specify it.

pavelfeldman: I'd like to treat it as a type, probably total=False should not be there

mxschmitt: I think mypy then complains from the user perspective if you only specify a single property.

PR #314, opened 2020-11-21
chore: only publish wheels package on Pypi

### LINKED ISSUE #312: Playwright on unsupported platforms
On [PyPI](https://pypi.org/project/playwright/#files) the source distribution of the project is uploaded as a result on unsupported platforms, running `pip install playwright` installs the project without playwright-cli and the user does not has access to [build_package.py](https://github.com/microsoft/playwright-python/blob/master/build_package.py). It is better to upload only the wheel of the supported platforms to PyPI so that the unsupported platforms does not download the source code and cause errors.

mxschmitt:
Fixes #312

The failing CI is not related.

### DIFF
--- a/build_package.py
+++ b/build_package.py
@@ -49,7 +49,7 @@ _egg_dir = _dirname / "playwright.egg-info"
 if _egg_dir.exists():
     shutil.rmtree(_egg_dir)
 
-subprocess.check_call("python setup.py sdist bdist_wheel", shell=True)
+subprocess.check_call("python setup.py bdist_wheel", shell=True)
 
 base_wheel_location = glob.glob("dist/*.whl")[0]
 without_platform = base_wheel_location[:-7]


pavelfeldman:
I don't think this fixes anything, there is still no recipe on how to use Playwright on unsupported platforms.

mxschmitt:
> I don't think this fixes anything, there is still no recipe on how to use Playwright on unsupported platforms, or perhaps, I am missing something?

With this change Pip raises a unsupported platform error. Before that it leads to a broken installation.

kumaraditya303:
> > I don't think this fixes anything, there is still no recipe on how to use Playwright on unsupported platforms, or perhaps, I am missing something?
> 
> With this change Pip raises a unsupported platform error. Before that it leads to a broken installation.

Yes, because of this error it is clear that the user's platform isn't supported not that it is a problem of playwright!

pavelfeldman:
Merged as a part of another PR

Issue #310, opened 2020-11-20
Interactive mode (REPL) Error


zhonghai9967:
from playwright import sync_playwright
playwright = sync_playwright().start()
==>
AttributeError: 'NoneType' object has no attribute 'fileno'
why?
Thank you！

mxschmitt:
Hi, do you a full stack trace of it available? In which REPL are you executing it?

zhonghai9967:
python ide
win10

dld8888:
>>> from playwright import sync_playwright
>>> playwright = sync_playwright().start()
Traceback (most recent call last):
  File "C:\Python378\lib\site-packages\playwright\main.py", line 49, in _get_stderr_fileno
    return sys.stderr.fileno()
io.UnsupportedOperation: fileno

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "<pyshell#1>", line 1, in <module>
    playwright = sync_playwright().start()
  File "C:\Python378\lib\site-packages\playwright\__init__.py", line 34, in sync_playwright
    return SyncPlaywrightContextManager()
  File "C:\Python378\lib\site-packages\playwright\main.py", line 81, in __init__
    self._connection = run_driver()
  File "C:\Python378\lib\site-packages\playwright\main.py", line 76, in run_driver
    return loop.run_until_complete(run_driver_async())
  File "C:\Python378\lib\asyncio\base_events.py", line 587, in run_until_complete
    return future.result()
  File "C:\Python378\lib\site-packages\playwright\main.py", line 61, in run_driver_async
    stderr=_get_stderr_fileno(),
  File "C:\Python378\lib\site-packages\playwright\main.py", line 54, in _get_stderr_fileno
    return sys.__stderr__.fileno()
**AttributeError: 'NoneType' object has no attribute 'fileno'**

dld8888:
>>> from playwright import sync_playwright
>>> playwright = sync_playwright().start()
Traceback (most recent call last):
  File "C:\Python378\lib\site-packages\playwright\main.py", line 49, in _get_stderr_fileno
    return sys.stderr.fileno()
io.UnsupportedOperation: fileno

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<pyshell#1>", line 1, in <module>
    playwright = sync_playwright().start()
  File "C:\Python378\lib\site-packages\playwright\__init__.py", line 34, in sync_playwright
    return SyncPlaywrightContextManager()
[...]

Issue #158, opened 2020-08-06
Stdout/stderr pipe throws stacktrace if stdout/stderr is mocked by Pytest


mxschmitt:
I think more an upstream issue (here) instead of for the Pytest plugin.

Stacktrace:

```
pytest project-name/browser_tests --browser chromium --base-url http://localhost:8000/
============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
baseurl: http://localhost:8000/
django: settings: config.settings.test (from option)
Using --randomly-seed=1
rootdir: /home/runner/work/project-name/project-name, configfile: setup.cfg
plugins: freezegun-0.4.2, base-url-1.4.2, playwright-0.0.5, cov-2.10.0, sugar-0.9.4, django-3.9.0, randomly-3.4.1, celery-4.4.7, factoryboy-2.0.3, Faker-4.1.1, env-0.6.2
collected 15 items

project-name/browser_tests/test_internal_manager_dashboard.py E [  6%]


==================================== ERRORS ====================================
_ ERROR at setup of test123[chromium] _

launch_browser = <function launch_browser.<locals>.launch at 0x7f6bc73de430>

    @pytest.fixture(scope="session")
    def browser(launch_browser: Callable[[], Browser]) -> Generator[Browser, None, None]:
>       browser = launch_browser()

/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pytest_playwright/pytest_playwright.py:114: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pytest_playwright/pytest_playwright.py:97: in launch
    pw_context = sync_playwright()
/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/playwright/__init__.py:27: in sync_playwright
    return SyncPlaywrightContextManager()
/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/playwright/main.py:72: in __init__
    self._connection = run_driver()
/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/playwright/main.py:67: in run_driver
    return loop.run_until_complete(run_driver_async())
[...]

PR #318, opened 2020-11-22
fix(stderr): remove the stderr patching hacks

pavelfeldman:
A fix for https://github.com/microsoft/playwright-python/issues/310

Will this make any bots fail? I am running it locally with `-n 4` and things work, is that not xdist?

### DIFF
--- a/playwright/main.py
+++ b/playwright/main.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import asyncio
-import io
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


mxschmitt:
That #158 was the original bug report, did you use the `-s` option?

pavelfeldman:
> did you use the -s option?

Yes, that's the only way to see the system err.

mxschmitt:
> > did you use the -s option?
> 
> Yes, that's the only way to see the system err.

Yeah thats the thing. If you are not using it, then pytest-xdist is monkey patching it (patching it during the runtime) and then the error occurs.

pavelfeldman:
@mxschmitt still not getting it...

I can run this:
```
DEBUG=pw:api pytest tests/sync -n 4
```
and it'll work, with no stderr in terminal

I can run this:
```
DEBUG=pw:api pytest tests/sync -n 4 -s
```
and it'll work, with stderr in terminal

Are there any steps to repro the actual issue?

mxschmitt (approved):
I tried to reproduce it: create a test project, playwright + pytest-playwright, using the exact same pytest version and parameters as I did in the original bug-report.

Since I'm not able to reproduce, lets merge I would say and wait for real users who report it.

(This bug report was from an application from my previous company)

PR #320, opened 2020-11-22
fix: Support MacOS BigSur

### LINKED ISSUE #304: playwright-cli not found on MacOS Big Sur
Step to reproduce:
1. Create virtualenv with python 3.8.6: python -m venv .env
2. Activate virtualenv and install playwright: source .env/bin/activate && pip install -r requirements.txt
3. Run playwright install: python -m playwright install

Actual Result:
image

Playwright data inside virtualenv:
image

OS: MacOS Big Sur
Python version: 3.8.6 (install via pyenv)

kumaraditya303:
Fixes #304 and and create wheel compatible with MacOS Big Sur

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,16 +81,16 @@ jobs:
       run: python -m playwright install
     - name: Test Sync API
       if: matrix.os != 'ubuntu-latest'
-      run: pytest -vv tests/sync --browser=${{ matrix.browser }} --timeout 90
+      run: pytest -vv tests/sync --browser=${{ matrix.browser }}
     - name: Test Sync API
       if: matrix.os == 'ubuntu-latest'
-      run: xvfb-run pytest -vv tests/sync --browser=${{ matrix.browser }} --timeout 90
+      run: xvfb-run pytest -vv tests/sync --browser=${{ matrix.browser }}
     - name: Test Async API
       if: matrix.os != 'ubuntu-latest'
-      run: pytest -vv tests/async --browser=${{ matrix.browser }} --timeout 90
+      run: pytest -vv tests/async --browser=${{ matrix.browser }}
     - name: Test Async API
       if: matrix.os == 'ubuntu-latest'
-      run: xvfb-run pytest -vv tests/async --browser=${{ matrix.browser }} --timeout 90
+      run: xvfb-run pytest -vv tests/async --browser=${{ matrix.browser }}
 
   test-package-installations:
     name: Test package installations
--- a/build_package.py
+++ b/build_package.py
@@ -76,7 +76,7 @@ for platform in ["mac", "linux", "win32", "win32_x64"]:
 
     wheel = ""
     if platform == "mac":
-        wheel = "macosx_10_13_x86_64.whl"
+        wheel = "macosx_10_15_x86_64.whl"
     if platform == "linux":
         wheel = "manylinux1_x86_64.whl"
     if platform == "win32":


kumaraditya303:
@mxschmitt It seems that the tests were failing because of timeout and after timeout is removed all the tests passed.

pavelfeldman:
Isn't this change backwards? You roll to 10.15 (Catalina) and remove support for High Sierra and Mojave (10.13 and 10.14)?

kumaraditya303:
@pavelfeldman by this change we only drop support for High Sierra but it is compatible with Catalina, Mojave and BigSur. 
For compatibility with High Sierra we can build separate wheel for High Sierra if we want it.

pavelfeldman:
Nice! I was myself looking for the docs explaining how to make it compatible with BigSur, but did not find any. Could you point me to something?

kumaraditya303:
> Nice! I was myself looking for the docs explaining how to make it compatible with BigSur, but did not find any. Could you point me to something?

We do not have real compatibility issues but because of the version MacOS BigSur we were having issue, to fix compatibility on Big Sur as well as older, I have used 10_15 which is compatible with Mojave, Catalina and BigSur.
You can cross check the version using `pip debug` on BigSur device and match with older versions.

pavelfeldman:
BigSur gives me py3-none-macosx_11_0x86_64. I tried it before, but wheel rejected this name. That's why I am curious about how you make it work.

pavelfeldman:
This is where I tried uploading it: https://github.com/microsoft/playwright-python/runs/1437572510?check_suite_focus=true.
For this proposed change, I don't understand why https://www.python.org/dev/peps/pep-0427/#file-name-convention would allow picking 10_15 when only 11_0 is supported...

kumaraditya303:
@pavelfeldman the macosx Big Sur is version 10_16 in python versioning tag not 11_0 . You may try the version in this PR which is compatible with Big Sur, Catalina, and Majave.

pavelfeldman:
PXL_20201124_032159404

pavelfeldman:
>  the macosx Big Sur is version 10_16 in python versioning tag not 11_0

I was looking for any evidence of that in the docs, but could not find. Do you have any links?

pavelfeldman:
Are you trying it on actual BigSur? I have one in hand...

kumaraditya303:
I tried it on CI with macosx 11.0.1 and I didn't get the tags as in your screenshot

kumaraditya303:
> This is where I tried uploading it: https://github.com/microsoft/playwright-python/runs/1437572510?check_suite_focus=true.
> For this proposed change, I don't understand why https://www.python.org/dev/peps/pep-0427/#file-name-convention would allow picking 10_15 when only 11_0 is supported...
> 
> @kumaraditya303 ^^

As per the logs of GitHub Action it is clear that the tags 11_0 don't exists .

pavelfeldman:
Yes, I've seen 10_16 on the GHA bots, I guess the run some early BigSur beta. The picture that I attached is made from the released version of BigSur.

kumaraditya303:
Did you try installing 10_15 on Big Sur? It would be helpful if you can attach the logs when you installed 10_15 on BigSur ... ^^
@pavelfeldman

pavelfeldman:
> Did you try installing 10_15 on Big Sur

There is no 10_15 wheel, but existing installation is saying `none of the wheel's tags match`. You could probably get a test package, deploy it, check it and that would justify a solution? I don't think 10_15 works because it is no different from 10_13. Overall, we should probably wait for someone who has a clue on how platform wheels are handled in BigSur. There is a bunch of projects in the Python ecosystem that are currently facing the same request, so it should not take long. I would also expect pip to have a bug filed about the `has an unsupported platform tag 'macosx_11_0_x86_64'`.

pavelfeldman:
https://github.com/pypa/packaging/pull/319 seems to answer all our questions, including 10_16 presence.

PR #322, opened 2020-11-24
ci: Removed pytest timeout

kumaraditya303:
Removed pytest timeout as it causes random test fails especially of video tests as the total test time is already limited by github `    timeout-minutes: 30`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,16 +81,16 @@ jobs:
       run: python -m playwright install
     - name: Test Sync API
       if: matrix.os != 'ubuntu-latest'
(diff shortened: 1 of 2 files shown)


pavelfeldman:
Doesn't this compromise the entire suite if one of the tests hang?

mxschmitt:
> Removed pytest timeout as it causes random test fails especially of video tests as the total test time is already limited by github ` timeout-minutes: 30`

I would prefer to increase it to e.g. 4 minutes instead of using the GitHub Actions timeout which will kill the process in the end and not generate a good output which test was failing.

pavelfeldman:
Closing as per above. Overall, we should start with the issue description, so that it was clear what is being addressed!

PR #345, opened 2020-12-04
fix get_file_dirname for pyinstaller (fix #342)

### LINKED ISSUE #342: Inspect fails to retrieve module inside frozen app
# Description of the issue

when freezing playwright with [pyinstaller](https://github.com/pyinstaller/pyinstaller) the assertion in [`get_file_dirname`](https://github.com/microsoft/playwright-python/blob/master/playwright/path_utils.py#L19-L24) fails

```
  File "main.py", line 6, in <module>
    with sync_playwright() as p:
  File "playwright/__init__.py", line 34, in sync_playwright
  File "playwright/main.py", line 81, in __init__
  File "playwright/main.py", line 76, in run_driver
  File "asyncio/base_events.py", line 587, in run_until_complete
  File "playwright/main.py", line 44, in run_driver_async
  File "playwright/main.py", line 36, in compute_driver_executable
  File "playwright/path_utils.py", line 23, in get_file_dirname
AssertionError
```

The reason is that the inspect module fails to find the module on the following lines inside [`get_file_dirname`](https://github.com/microsoft/playwright-python/blob/master/playwright/path_utils.py#L19-L24)

```python
frame = inspect.stack()[1]
module = inspect.getmodule(frame[0])
```

# Context information (for bug reports)

* version of python: 3.7.9
* version of pyinstaller: 4.1
* version of playwright-python: Version 0.162.1
* platform: GNU/Linux (Ubuntu 18.04 LTS)

# Reproducing the bug

install packages

    $ pip install playwright pyinstaller

install the browsers inside playwright

    $ PLAYWRIGHT_BROWSERS_PATH=0 python -m playwright install

create `main.py`

```python
# main.py
import sys
from pathlib import Path
from playwright import sync_playwright

with sync_playwright() as p:
    for browser_type in [p.chromium]:
        browser = browser_type.launch(
            headless=False,
            executablePath=Path(sys.modules['playwright'].__file__).parent / 'driver' / '.local-browsers' / 'chromium-827102' / 'chrome-linux' / 'chrome')
        page = browser.newPage()
        page.goto('http://whatsmyuseragent.org/')
        browser.close()
```

freeze into single binary file with pyinstaller

    $ pyinstaller -F main.py

execute the binary file

```
$ ./dist/main 
  File "main.py", line 6, in <module>
    with sync_playwright() as p:
  File "playwright/__init__.py", line 34, in sync_playwright
  File "playwright/main.py", line 81, in __init__
  File "playwright/main.py", line 76, in run_driver
  File "asyncio/base_events.py", line 587, in run_until_complete
  File "playwright/main.py", line 44, in run_driver_async
  File "playwright/main.py", line 36, in compute_driver_executable
  File "playwright/path_utils.py", line 23, in get_file_dirname
AssertionError
```

tumregels:
fix [get_file_dirname](https://github.com/microsoft/playwright-python/blob/master/playwright/path_utils.py#L19) so that playwright can function when freezed with pyinstaller

### DIFF
--- a/playwright/path_utils.py
+++ b/playwright/path_utils.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 import inspect
+import sys
 from pathlib import Path
 
 
 def get_file_dirname() -> Path:
     """Returns the callee (`__file__`) directory name"""
-    frame = inspect.stack()[1]
-    module = inspect.getmodule(frame[0])
+    frame = inspect.currentframe()
+    module_name = frame.f_back.f_globals["__name__"]
+    module = sys.modules[module_name]
     assert module
     return Path(module.__file__).parent.absolute()


mxschmitt:
Do you know the reason why it is not working in this specific case? The workaround looks quite weird, but I'm also not very experienced in Python.

tumregels:
Not sure, but [this issue](https://github.com/indygreg/PyOxidizer/issues/69) somewhat explains the problem

pavelfeldman:
@tumregels this looks good to me, but do you think we could replace the usage of this method with something more straightforward in `main.py` instead? That's the only call site that works at runtime, the rest is for testing...

tumregels:
@pavelfeldman there seems to be no solid solution to this problem, at least can't suggest any. [This answer](https://stackoverflow.com/a/58941536/2248627) tries to cover the available options. And the recommended [pkgutil](https://docs.python.org/3/library/pkgutil.html) is not an option for this case.

mxschmitt:
> @pavelfeldman there seems to be no solid solution to this problem, at least can't suggest any. [This answer](https://stackoverflow.com/a/58941536/2248627) tries to cover the available options. And the recommended [pkgutil](https://docs.python.org/3/library/pkgutil.html) is not an option for this case.

What about removing the single occurrence in the source and moving it inline into the callee and moving the get_file_dirname function into the tests directory?

tumregels:
@mxschmitt - based on the table [here](https://github.com/wimglenn/resources-example) can suggest to change [`compute_driver_executable`](https://github.com/microsoft/playwright-python/blob/master/playwright/main.py#L34-L39) with this code 

```python
import importlib.resources as pkg_resources

def compute_driver_executable() -> Path:
    _pkg = '.'.join(__name__.split(".")[:-1])
    with pkg_resources.path(_pkg, "") as package_path:
        assert package_path
        platform = sys.platform
        if platform == "win32":
            return package_path / "driver" / "playwright-cli.exe"
        return package_path / "driver" / "playwright-cli"
```

where `_pkg` is a workaround for [this issue](https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/remote/webelement.py#L46-L48)

pavelfeldman:
@tumregels 

Something as simple as below works for me:

```python
from importlib import resources

with resources.path("playwright", "driver") as driver_folder:
```

Should we just use it in `main.py`?

tumregels:
@pavelfeldman 

just checked this one with pyinstaller - it works

```python
import playwright
from importlib import resources

def compute_driver_executable() -> Path:
    with resources.path(playwright, "driver") as driver_folder:
        assert driver_folder
        platform = sys.platform
        if platform == "win32":
            return driver_folder / "playwright-cli.exe"
        return driver_folder / "playwright-cli"
```

we can also change `playwright` with `playwright.__name__` because pycharm wrongly complains with `Expected type 'Union[str, ModuleType]', got '__init__.py' instead `

pavelfeldman:
@pavelfeldman: nice! do you want to update your PR with it?

tumregels:
@pavelfeldman will update the PR today.

mxschmitt:
Can we also move the get_file_dirname function to the tests utils.py file?

tumregels:
@mxschmitt @pavelfeldman the proposed solution is wrong. 
`"driver"` is not a [resource](https://docs.python.org/3/library/importlib.html#importlib.resources.path) - `cannot be a directory`. 
probably the package was fixed in python3.9, therefore it fails.

pavelfeldman:
So we should either add __init__.py into the folder or move the driver executable into the root folder? Either way should do...

tumregels:
@pavelfeldman current implementation should work. 

```
import inspect
import playwright

def compute_driver_executable() -> Path:
    package_path = Path(inspect.getfile(playwright)).parent
    platform = sys.platform
    if platform == "win32":
        return package_path / "driver" / "playwright-cli.exe"
    return package_path / "driver" / "playwright-cli"

```
it's failing because `setup.py` is using `curl` - which is a fragile solution.

regarding previous solution with `importlib.resources` - even adding `__init__.py` will not help - I already tried.
moving to the root is not recommended - as far as know.

Issue #350, opened 2020-12-05
Make sync playwright work on fresh threads


pavelfeldman:
We should create an async loop for threads that don't have any.

Mattwmaster58:
I actually just ran today into this problem when trying to use `sync_playwright` with Flask.

Mattwmaster58:
I actually just ran into this today with Flask. It seems simple enough to implement, I'd like to have a go at writing up this PR. Anything I should need to keep in mind?

Mattwmaster58:
For reference, here's a patch anybody should be able to use (Unix) or (Python 3.8+ and Windows) based on that PR:
```py
import asyncio
import playwright
import threading

# if you're on Windows and Python 3.7
# asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())

def threading_capable_run_driver():
	if not isinstance(threading.current_thread(), threading._MainThread):
		loop = asyncio.new_event_loop()
	else:
		loop = asyncio.get_event_loop()
		if loop.is_running():
			raise playwright.main.Error("Can only run one Playwright at a time.")
	return loop.run_until_complete(playwright.main.run_driver_async())

playwright.main.run_driver = threading_capable_run_driver
```

pavelfeldman:
Should be fixed now.

PR #352, opened 2020-12-07
fix: create event loop manually in case of running within non-main thread

Mattwmaster58:
I believe this fixes what #350 is referring to.

### DIFF
--- a/playwright/main.py
+++ b/playwright/main.py
@@ -16,6 +16,7 @@ import asyncio
 import os
 import subprocess
 import sys
+import threading
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


Mattwmaster58:
This has issues when it comes to Linux + Python 3.7 that will have to be addressed.

playwright/main.py
```diff
@@ -59,9 +60,13 @@ async def run_driver_async() -> Connection:
 
 
 def run_driver() -> Connection:
-    loop = asyncio.get_event_loop()
-    if loop.is_running():
-        raise Error("Can only run one Playwright at a time.")
+    # if we're running in a thread, get_event_loop won't create an event loop for us, we'll need to make it ourselves
+    if not isinstance(threading.current_thread(), threading._MainThread):  # type: ignore
```
pavelfeldman, line 64: We should probably reuse existing loop, regardless of the thread. I have a code health patch that resolves it as a driver-by: https://github.com/microsoft/playwright-python/pull/354/files

Mattwmaster58: I wasn't aware. Feel free to open if this is still relevant.

Mattwmaster58: This is basically checking for one of the conditions which will cause `RuntimeError` to be raised an avoiding it prematurely.

playwright/main.py
```diff
@@ -112,9 +117,7 @@ async def __aexit__(self, *args: Any) -> None:
 
 
 if sys.platform == "win32":
-    # Use ProactorEventLoop in 3.7, which is default in 3.8
-    loop = asyncio.ProactorEventLoop()
-    asyncio.set_event_loop(loop)
+    asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
```
pavelfeldman, line 120: Now I know why you did this change @Mattwmaster58 !

PR #353, opened 2020-12-07
ft: allow async callbacks with Page.on calls

Mattwmaster58:
This allows for async callback functions as an argument to `Page.on`. Relevant [pyee documentation](https://pyee.readthedocs.io/en/latest/#pyee.AsyncIOEventEmitter):

>An event emitter class which can run asyncio coroutines in addition to synchronous blocking functions. For example:
> 
>     @ee.on('event')
>     async def async_handler(*args, **kwargs):
>         await returns_a_future()
> 
>On emit, the event emitter will automatically schedule the coroutine using asyncio.ensure_future and the configured event loop (defaults to asyncio.get_event_loop()).
> 
>Unlike the case with the BaseEventEmitter, all exceptions raised by event handlers are automatically emitted on the error event. This is important for asyncio coroutines specifically but is also handled for synchronous functions for consistency.
> 
>When loop is specified, the supplied event loop will be used when scheduling work with ensure_future. Otherwise, the default asyncio event loop is used.
> 
>For asyncio coroutine event handlers, calling emit is non-blocking. In other words, you do not have to await any results from emit, and the coroutine is scheduled in a fire-and-forget fashion.

### DIFF
--- a/playwright/connection.py
+++ b/playwright/connection.py
@@ -18,14 +18,14 @@ import traceback
 from typing import Any, Callable, Dict, Optional, Union
 
 from greenlet import greenlet
-from pyee import EventEmitter
+from pyee import AsyncIOEventEmitter
 
 from playwright.helper import ParsedMessagePayload, parse_error
 from playwright.sync_base import dispatcher_fiber
 from playwright.transport import Transport
 
 
-class Channel(EventEmitter):
+class Channel(AsyncIOEventEmitter):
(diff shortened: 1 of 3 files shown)


pavelfeldman:
Looks great! Let's update the tests to reflect these changes. For example, pagemost of the `.on("dialog", on_dialog)` occurrences in the tests would benefit from it.

PR #354, opened 2020-12-07
fix(loop): greenlet + async code health code

pavelfeldman:
- bubbles the run_until_complete to main.py
- creates loop if missing
- removes dispatcher greenlet singleton

### DIFF
--- a/playwright/transport.py
+++ b/playwright/transport.py
@@ -15,42 +15,47 @@
 import asyncio
 import json
 import os
+import sys
+from pathlib import Path
 from typing import Dict
 
 
 class Transport:
-    def __init__(
-        self,
-        input: asyncio.StreamReader,
-        output: asyncio.StreamWriter,
-        loop: asyncio.AbstractEventLoop,
-    ) -> None:
+    def __init__(self, driver_executable: Path) -> None:
         super().__init__()
-        self._input: asyncio.StreamReader = input
-        self._output: asyncio.StreamWriter = output
-        self.loop: asyncio.AbstractEventLoop = loop
         self.on_message = lambda _: None
         self._stopped = False
-
-    def run_sync(self) -> None:
-        self.loop.run_until_complete(self._run())
-
-    def run_async(self) -> None:
-        self.loop.create_task(self._run())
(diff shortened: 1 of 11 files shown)


playwright/transport.py
```diff
@@ -15,42 +15,47 @@
...
         self.on_message = lambda _: None
         self._stopped = False
-
-    def run_sync(self) -> None:
-        self.loop.run_until_complete(self._run())
-
-    def run_async(self) -> None:
-        self.loop.create_task(self._run())
+        self._driver_executable = driver_executable
+        self.loop: asyncio.AbstractEventLoop
```
dgozman, line 29: Probably `self._loop` with an underscore?

dgozman (approved):

Mattwmaster58:
Something I just encountered is that with Python 3.7 + threads + unix, will fail with this error:
```
RuntimeError: Cannot add child handler, the child watcher does not have a loop attached
```
The solution is to make sure there's a asyncio event loop in the main thread and that the child watcher is activated:
```py
# (at file level)
asyncio.get_current_loop()
asyncio.get_child_watcher
```
(not necessary in 3.8+ as the default child watcher implementation is better)
Out of the scope this PR, but I feel there should at least be some sort of mention of this in the docs.

Mattwmaster58:
Something I just encountered is that with Python 3.7 + threads + unix, will fail with this error:
```
RuntimeError: Cannot add child handler, the child watcher does not have a loop attached
```
The solution is to make sure there's a asyncio event loop in the main thread and that the child watcher is activated:
```py
# (at file level)
asyncio.get_current_loop()
asyncio.get_child_watcher
```
(not necessary in 3.8+ as the default child watcher implementation doesn't have this limitation)
Out of the scope this PR, but it seems like there should be maybe a mention of this somewhere? Or is ti too niche a use-case?

pavelfeldman:
@Mattwmaster58 I'll add some tests for the Thread. Is that just running sync python on a thread use case?

Mattwmaster58:
Yeah, an example for this setup would gunicorn + flask with endpoints that use the sync interface.

PR #360, opened 2020-12-10
chore: Update dependencies and configure Dependabot for monthly updates

kumaraditya303:
Update dependencies and configure dependabot for monthly updates.


kumaraditya303:
It would be useful to configure dependabot for this project.

mxschmitt:
I wouldn't optimise the images, since the exact images get used in all the different language providers. For dependencies I think its preferred to update them manually every few months, since its only dev dependencies and does not harm anyone. Would also result in less noise in GitHub notifications for the maintainers.

kumaraditya303:
> I wouldn't optimise the images, since the exact images get used in all the different language providers. For dependencies I think its preferred to update them manually every few months, since its only dev dependencies and does not harm anyone. Would also result in less noise in GitHub notifications for the maintainers.

@mxschmitt optimize images means that the images are compressed their dimensions are untouched

mxschmitt:
> > I wouldn't optimise the images, since the exact images get used in all the different language providers. For dependencies I think its preferred to update them manually every few months, since its only dev dependencies and does not harm anyone. Would also result in less noise in GitHub notifications for the maintainers.
> 
> @mxschmitt optimize images means that the images are compressed and their dimensions are untouched

It wouldn't bring any benefit in terms of performance or speed for new developers since we are using Git and the images are already in the tree like that. So the git clone would also take longer since new unknown binary files got added.

kumaraditya303:
@mxschmitt Dependabot is now configured for monthly

pavelfeldman:
Does this fix any bugs?

PR #361, opened 2020-12-10
fix(threads): fix the separate thread execution in 3.7, add tests

pavelfeldman:


tests/sync/test_sync.py
```diff
@@ -205,3 +206,16 @@ def test_expect_response_should_work(page: Page, server):
     assert resp.value.status == 200
     assert resp.value.ok
     assert resp.value.request
+
+
+def test_running_in_thread(browser_name):
+    class TestThread(threading.Thread):
+        def run(self):
+            with sync_playwright() as playwright:
+                browser = playwright.__dict__[browser_name].launch()
```
mxschmitt, line 215: nit: getattr would be better since "dunder methods" should not be used. But since its only a test I'm fine with it.

mxschmitt (approved):

tests/sync/test_sync.py
pavelfeldman: This is dead code, it should have been deleted.

tests/sync/test_sync.py
```diff
@@ -15,6 +15,7 @@
 import os
 
 import pytest
+import threading
```
mxschmitt, line 18:
```suggestion
```

mxschmitt (approved):

PR #375, opened 2020-12-17
api(snake): go snake

pavelfeldman:
Now that 0.170.0 is live, we are going to work on the Playwright v1.8.0. That's a first semver stable with the fixed API that is going to be maintained for years.

With this transition we are going to break our existing customers that use camelCase notation. As promised, we are doing this only once, most of the changes are going to be around the camelCase to snake_case transition in the scripts.


kumaraditya303:
@pavelfeldman How about releasing a snake case pre release candidate ?

PR #378, opened 2020-12-17
api(snake): Make Readme 🐍 friendly

kumaraditya303:


pavelfeldman:
Hi! Thanks for this change, but we should only do this when we actually release the snake_case API.

kumaraditya303:
closing as merged as part of another PR

PR #413, opened 2021-01-05
build: Build Wheels compatible for Apple M1

kumaraditya303:


pavelfeldman:
Did you test it? My Apple M1 does not have `macosx_11_0_universal2`, it has `py3-none-macosx_11_0_universal`

setup.py
```diff
@@ -90,6 +90,10 @@ def run(self) -> None:
                             from_location, os.stat(from_location).st_mode | stat.S_IEXEC
                         )
                     zip.write(from_location, to_location)
+            if platform == "mac":
+                shutil.copyfile(
+                    wheel_location, without_platform + "macosx_11_0_universal2.whl"
```
pavelfeldman, line 95: My recollection is that Python won't accept identical wheel files for different platforms. Did you test this?

pavelfeldman:
I verified that it does not work:
```
HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
File already exists. See https://pypi.org/help/#file-name-reuse for more information
```

PR #426, opened 2021-01-13
docstrings: only render async/sync dialect snippets

pavelfeldman:


dgozman (approved):

scripts/documentation_provider.py
```diff
@@ -190,21 +191,29 @@ def indent_paragraph(self, p: str, indent: str) -> str:
...
             if not line.strip():
                 last_was_blank = True
                 continue
-            if line.strip() == "```js":
-                in_example = True
-            if not in_example:
+            match = re.match(r"\s*```(.+)", line)
+            if match:
+                lang = match[1]
+                if lang in ["html", "yml", "sh", "py", "python"]:
```
mxschmitt, line 203: For the long-term it would be better to move that upstream I think, so each method has language specific examples attached. We loose then the location and have to put them at the end of a method description but should be fine.

PR #473, opened 2021-02-01
test: fix flaky tests by rerunning

kumaraditya303:
Reruns flaky tests five times in total at maximum, speed up tests using `pytest-xdist` , it was already in requirements but was not used.


kumaraditya303:
Some tests have failed because of some network issue during apt update https://github.com/microsoft/playwright-python/pull/473/checks?check_run_id=1805878399

pavelfeldman:
We should not use xdist - it tries to use 2 processes, but each spawns browser and renderer processes, so we overbook 3 times. Also explicit flaky markers are good to know what we need to fix.

PR #495, opened 2021-02-11
fix: do not leak api wrappers

pavelfeldman:


mxschmitt (approved):

playwright/_impl/_impl_to_api_mapping.py
```diff
@@ -77,14 +79,14 @@ def to_impl(self, obj: Any) -> Any:
...
-
-        def wrapper(*args: Any) -> Any:
+        def wrapper_func(*args: Any) -> Any:
             arg_count = len(inspect.signature(handler).parameters)
             return handler(
                 *list(map(lambda a: self.from_maybe_impl(a), args))[:arg_count]
             )
 
-        self._instances[handler] = wrapper
+        wrapper = getattr(handler, INSTANCE_ATTR, None)
```
dgozman, line 88: It seems like this should `__pw_impl_instance`, instead of `__pw_api_instance`?

dgozman (approved):

playwright/_impl/_impl_to_api_mapping.py
pavelfeldman: will follow up

PR #502, opened 2021-02-13
chore: Install typing-extensions only on python_version<=3.8

kumaraditya303:


setup.py
```diff
@@ -29,13 +28,11 @@
...
 
-NoneType = type(None)
 
-
-def extractall(zip: typing.Any, path: str) -> NoneType:
+def extractall(zip: zipfile.ZipFile, path: str) -> None:
     for name in zip.namelist():
         member = zip.getinfo(name)
-        extracted_path = zip._extract_member(member, path, None)
+        extracted_path = zip.extract(member, path, None)
```
mxschmitt, line 35:
```suggestion
        extracted_path = zip.extract(member, path)
```
nit: None is not needed there.

mxschmitt (approved):

pavelfeldman (approved):

PR #516, opened 2021-02-19
chore: roll Playwright to ToT

pavelfeldman:


dgozman (approved):

playwright/_impl/_wait_helper.py
```diff
@@ -13,22 +13,27 @@
 # limitations under the License.
 
 import asyncio
+import uuid
```
mxschmitt, line 16: do we already have it as a dependency? I think it needs to be added to the setup.py.

mxschmitt (approved):

PR #548, opened 2021-03-05
docs(readme): use aka.ms Playwright Slack invite link

mxschmitt:
These links expire after a certain time and use-count. Would be cool if we can use this everywhere to manage it easily.


pavelfeldman:
Which one expired? The one we have upstream as well?

mxschmitt:
> Which one expired? The one we have upstream as well?

Not yet expired but expires probably in the future and then we only have to adjust the aka.ms redirect.

pavelfeldman (approved):
Could you make me Lusha or me the owner of the link?

mxschmitt:
> Could you make me Lusha or me the owner of the link?

current owner is @arjunattam I think. (It's internal only the shortener.)

PR #559, opened 2021-03-09
add a 1 line API for creating a page

verhovsky:


pavelfeldman:
Thanks for the patch! As you are saying, consistency is more important to us, so we'd rather not!

PR #561, opened 2021-03-10
devops: Use Pull Request Labeler

kumaraditya303:


kumaraditya303:
The labeller action is failing because I don't have permission to apply labels @pavelfeldman

pavelfeldman:
Let's wait with this one for now - we anyways need to go over all the files during the review...

Issue #477, opened 2021-02-02
[bug] "firefox" browser was not found.


mrthevinh:
playwright playwright==1.8.0a1
playwright install
Run my script with firefox erro.. webkit and chrome still work
```
"firefox" browser was not found.
Please complete Playwright installation via running

    "python -m playwright install"
```

pavelfeldman:
That's unusual, we have bots testing it. Could you share more on your OS and Python version? I assume first line of your bug should have been

```
pip install playwright==1.8.0a1
```

mrthevinh:
My OS is window 10 python 3.9. I try it on python 3.7 and 3.8 too but firefox still dont work!

```Traceback (most recent call last):  
    browser = browser_type.launch()
  File "C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\sync_api.py", line 6974, in launch
    self._sync(
  File "C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\_sync_base.py", line 97, in _sync
    return future.result()
  File "C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\_browser_type.py", line 77, in launch
    raise not_installed_error(f'"{self.name}" browser was not found.')
Exception: 
```
 Could you tell me how to debug it to get more report

mxschmitt:
Could you try to add a `print(e)` of the original exception in `C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\_browser_type.py` and provide us the output of that?

mrthevinh:
@mxschmitt thank you reply.. My error log show when I add print(e):
```
ERROR: The process "39608" not found.
Traceback (most recent call last):
  File "C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\_impl\_browser_type.py", line 70, in launch
    return from_channel(await self._channel.send("launch", params))
  File "C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\_impl\_connection.py", line 36, in send
    return await self.inner_send(method, params, False)
  File "C:\Users\The Vinh\AppData\Local\Programs\Python\Python39\lib\site-packages\playwright\_impl\_connection.py", line 47, in inner_send
    result = await callback.future
playwright._impl._api_types.Error: Protocol error (Browser.enable): Browser closed.
[...]

PR #572, opened 2021-03-17
fix: enhance executable not found check

### LINKED ISSUE #577: [BUG] misleading error message "browser was not found" if browser fails to start for other reasons
# Environment
playwright==1.9.2
ubuntu 20.04 in docker
chromium v854489 installed via `python3 -m playwright install`

# Problem description:
Unintentionally, the test attempted to launch headful chromium, launch failed with misleading error message:
```
"chromium" browser was not found.
Please complete Playwright installation via running
    "python -m playwright install"
```
Launching browser as `chromium.launch(headless=True)` was succesful, so the problem clearly was not with browser installation, but browser launch inself (most probably, unavailable X11)

# Expected behavior:
browser stderr captured and displayed

mxschmitt:
Addresses #477

Fixes #577

and on [Slack](https://playwright.slack.com/archives/C0182HLNSJV/p1615886948000700?thread_ts=1615818192.075500&cid=C0182HLNSJV)

### DIFF
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -69,7 +69,7 @@ class BrowserType(ChannelOwner):
         try:
             return from_channel(await self._channel.send("launch", params))
         except Exception as e:
-            if f"{self.name}-" in str(e):
+            if "because executable doesn't exist" in str(e):
                 raise not_installed_error(f'"{self.name}" browser was not found.')
             raise e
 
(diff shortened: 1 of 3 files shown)


pavelfeldman (approved):
Please check if this works cross-platform!

mxschmitt:
> Please check if this works cross-platform!

It does, because its a hard-coded message from upstream Playwright: https://github.com/microsoft/playwright/blob/0c7da44465c3802fec6211bafd9609f6b773f5de/src/server/browserType.ts#L172

Shall I create a test for it? (launch browser with non-existing executable path should result in the expected message)

pavelfeldman (approved):

PR #579, opened 2021-03-18
chore: roll to ToT

pavelfeldman:


tests/async/test_page.py
```diff
@@ -30,7 +30,7 @@ async def test_close_should_reject_all_promises(context):
     assert "Protocol error" in exc_info.value.message
 
 
-async def test_closed_should_not_visible_in_context_pages(context):
+async def test_closed_should_not_visible_in_context_pages(context, browser):
```
mxschmitt, line 33: whats the reason for that?

tests/conftest.py
```diff
@@ -48,8 +48,10 @@ def assetdir():
 
 @pytest.fixture(scope="session")
 def launch_arguments(pytestconfig):
+    print(pytestconfig.getoption("--browser-channel"))
```
mxschmitt, line 51: do you think this has potential to get adopted in pytest-playwright? I think so.

mxschmitt (approved):

tests/async/test_page.py
pavelfeldman: i was testing the version to see if channel works... reverting

tests/conftest.py
pavelfeldman: Yes, I think so.

PR #613, opened 2021-04-04
docs: add the -U flag for pip install

mxschmitt:
PIP by default does not download the latest version when there is already a version present on the system. If you want to have "a correct" behaviour, then you need to add the `-U` flag (U means update). Other big Python projects like Celery do that too.

ref: https://playwright.slack.com/archives/C0182HLNSJV/p1617547493062200?thread_ts=1617545917.060000&cid=C0182HLNSJV

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Headless execution is supported for all browsers on all platforms.
 ## Usage
 
 ```sh
-pip install playwright
+pip install -U playwright
 playwright install
 ```
 


kumaraditya303:
@mxschmitt there isn't any need to explicitly want the end user to upgrade to the latest version, it can create confusion in case of an backward incompatible change. If user want to upgrade to new version they can do it but there isn't any need for upgrade flag to be in the Readme. Also the celery project is highly stable so there won't such concerns for that project. Also "correct behaviour" should not mean to always upgrade existing

PR #614, opened 2021-04-06
chore:better type for permissions

kumaraditya303:


kumaraditya303:
@mxschmitt I have given permission to edit PR by maintainers, can you help to fix the linting issue ?

mxschmitt (approved):

kumaraditya303:
@mxschmitt I have generated the api but still it is showing linting errors not sure about it

kumaraditya303:
Even after rebasing to master it is failing

pavelfeldman (commented):
Since those can be browser-specific, I don't think we should suggest those.

kumaraditya303:
https://playwright.dev/docs/api/class-browsercontext#browsercontextgrantpermissionspermissions-options as per the docs there isn't any browser specific permission

PR #621, opened 2021-04-15
fix: fix greenlet version for compatibility with pypy

kumaraditya303:


kumaraditya303:
@mxschmitt Request to review

mxschmitt:
Any resource available for that? Or did a user face into that problem?

kumaraditya303:
I myself was trying to use it with pypy, the problem is that pypy comes pre-installed with a greenlet version which was earlier incompatible , with this PR it is fixed.
You can try it in docker container
```console
$ docker run --rm -it pypy:3 bash
```
@mxschmitt

jamadden:
Since greenlet is built-in to PyPy, the better way to fix this is to specify that the greenlet dependency is CPython specific using [PEP 496 environment markers](https://www.python.org/dev/peps/pep-0496/), e.g., gevent writes its dependency as:
```python
'greenlet >= 0.4.17, < 2.0; platform_python_implementation=="CPython"'
```

PR #622, opened 2021-04-15
feat(roll): roll to 1618513089000

dgozman:


mxschmitt (approved):
Thought to move the assert into a more generic place like `documentation_provider.py`, but that should already be ok where it is now.

PR #536, opened 2021-03-01
fix: handle cancelled tasks

cruzdanilo:
prevents `InvalidStateError` when the task has been cancelled, here:

```python
callback.future.set_exception(parsed_error)
```

PR #624, opened 2021-04-16
fix(connection): handle canceled tasks correctly

mxschmitt:
Was hanging before

Closes #536

### DIFF
--- a/tests/async/test_issues.py
+++ b/tests/async/test_issues.py
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from asyncio import FIRST_COMPLETED, CancelledError, create_task, wait
+
 import pytest
+
+from playwright.async_api import Page
 
 
 @pytest.mark.only_browser("chromium")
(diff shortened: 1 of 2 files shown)


tests/async/test_issues.py
```diff
@@ -30,3 +34,18 @@ async def test_issue_195(playwright, browser):
...
+        create_task(page.wait_for_selector("input")),
+        create_task(page.wait_for_selector("#will-never-resolve")),
+    }, return_when=FIRST_COMPLETED)
+    assert len(done) == 1
+    assert len(pending) == 1
+    for task in pending:
+        task.cancel()
+        with suppress(CancelledError):
+            await task
+    assert list(pending)[0].cancelled() is True
```
kumaraditya303, line 51:
```suggestion
    assert list(pending)[0].cancelled()
```

kumaraditya303 (changes requested):

tests/async/test_issues.py
```diff
@@ -30,3 +34,18 @@ async def test_issue_195(playwright, browser):
...
+    await page.set_content("<input />")
+    done, pending = await wait({
+        create_task(page.wait_for_selector("input")),
+        create_task(page.wait_for_selector("#will-never-resolve")),
+    }, return_when=FIRST_COMPLETED)
+    assert len(done) == 1
+    assert len(pending) == 1
+    for task in pending:
+        task.cancel()
+        with suppress(CancelledError):
```
kumaraditya303, line 49:
```suggestion
        with pytest.raises(CancelledError):
```

kumaraditya303 (changes requested):

tests/async/test_issues.py
kumaraditya303: This ensures that task was cancelled 🚀

mxschmitt: good catch, simplified it too.

kumaraditya303 (approved):

pavelfeldman (approved):

PR #625, opened 2021-04-16
test: re-renable CORS tests

### LINKED ISSUE #608: Release-blocker: uncomment cors tests
Search for this issue URL in the code.

mxschmitt:
After enabling the video test again WK+FF failed. The reason for that was a race between `context.on("page")` which comes after `page.on("video")` in WK and FF. This resulted in that the options were not yet there since the `page`->`browserContext` association was not yet there.

Seems like the condition which was removed was a leftover from the old video channels. In upstream PW we don't have such a logic, so lets keep it in sync.

Fixes #608

### DIFF
--- a/playwright/_impl/_page.py
+++ b/playwright/_impl/_page.py
@@ -779,8 +779,6 @@ class Page(ChannelOwner):
     def video(
         self,
     ) -> Optional[Video]:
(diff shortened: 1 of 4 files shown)


kumaraditya303:
@mxschmitt It seems like an upstream issue

pavelfeldman (approved):

PR #630, opened 2021-04-19
feat: added BrowserType.connect

### LINKED ISSUE #529: [BUG] Bring browser_type.connect in.

mxschmitt:
Closes #529

Quite an early implementation, need to think about the last skipped test.

### DIFF
--- a/local-requirements.txt
+++ b/local-requirements.txt
@@ -11,8 +11,9 @@ pixelmatch==0.2.3
 pre-commit==2.10.1
 pyOpenSSL==20.0.1
 pytest==6.2.2
-pytest-asyncio==0.14.0
+pytest-asyncio==0.15.0
 pytest-cov==2.11.1
+pytest-repeat==0.9.1
 pytest-sugar==0.9.4
 pytest-timeout==1.4.2
 pytest-xdist==2.2.1
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -149,23 +149,13 @@ class Browser(ChannelOwner):
         if self._is_closed_or_closing:
             return
         self._is_closed_or_closing = True
-        if self._is_connected_over_websocket:
-            await self._connection.stop_async()
-            self._notify_remote_closed()
-            return
         try:
             await self._channel.send("close")
         except Exception as e:
             if not is_safe_close_error(e):
                 raise e
-
-    def _notify_remote_closed(self) -> None:
-        # Emulate all pages, contexts and the browser closing upon disconnect.
-        for context in self.contexts:
-            for page in context.pages:
-                page._on_close()
-            context._on_close()
-        self._on_close()
+        if self._is_connected_over_websocket:
+            await self._connection.stop_async()
 
     @property
     def version(self) -> str:
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -146,13 +146,13 @@ class Browser(ChannelOwner):
         return page
 
     async def close(self) -> None:
+        if self._is_closed_or_closing:
+            return
+        self._is_closed_or_closing = True
         if self._is_connected_over_websocket:
             await self._connection.stop_async()
             self._notify_remote_closed()
             return
-        if self._is_closed_or_closing:
-            return
-        self._is_closed_or_closing = True
         try:
             await self._channel.send("close")
         except Exception as e:
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -50,6 +50,7 @@ class Browser(ChannelOwner):
         self._is_connected = True
         self._is_closed_or_closing = False
         self._is_remote = False
+        self._is_connected_over_websocket = False
 
         self._contexts: List[BrowserContext] = []
         self._channel.on("close", lambda _: self._on_close())
@@ -59,7 +60,7 @@ class Browser(ChannelOwner):
 
     def _on_close(self) -> None:
         self._is_connected = False
-        self.emit(Browser.Events.Disconnected)
+        self.emit(Browser.Events.Disconnected, self)
         self._is_closed_or_closing = True
 
     @property
@@ -145,6 +146,10 @@ class Browser(ChannelOwner):
         return page
 
     async def close(self) -> None:
+        if self._is_connected_over_websocket:
+            await self._connection.stop_async()
+            self._notify_remote_closed()
+            return
         if self._is_closed_or_closing:
             return
         self._is_closed_or_closing = True
@@ -153,6 +158,14 @@ class Browser(ChannelOwner):
         except Exception as e:
             if not is_safe_close_error(e):
                 raise e
+
+    def _notify_remote_closed(self) -> None:
+        # Emulate all pages, contexts and the browser closing upon disconnect.
+        for context in self.contexts:
+            for page in context.pages:
+                page._on_close()
+            context._on_close()
+        self._on_close()
 
     @property
     def version(self) -> str:
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -23,11 +23,7 @@ from playwright._impl._api_structures import (
 )
 from playwright._impl._browser import Browser, normalize_context_params
 from playwright._impl._browser_context import BrowserContext
-from playwright._impl._connection import (
-    ChannelOwner,
-    from_channel,
-    from_nullable_channel,
-)
+from playwright._impl._connection import ChannelOwner, Connection, from_channel
 from playwright._impl._helper import (
     BrowserChannel,
     ColorScheme,
@@ -35,6 +31,7 @@ from playwright._impl._helper import (
     locals_to_params,
     not_installed_error,
 )
+from playwright._impl._transport import WebSocketTransport
 
 
 class BrowserType(ChannelOwner):
@@ -157,6 +154,24 @@ class BrowserType(ChannelOwner):
         if default_context:
             browser._contexts.append(default_context)
             default_context._browser = browser
+        return browser
+
+    async def connect(
+        self, ws_endpoint: str, timeout: float = None, slow_mo: float = None
+    ) -> Browser:
+        transport = WebSocketTransport(ws_endpoint, timeout)
+
+        connection = Connection(None, self._connection._object_factory, transport)
+        connection._loop = self._connection._loop
+        connection._loop.create_task(connection.run())
+        playwright = await connection.wait_for_object_with_known_name("Playwright")
+        pre_launched_browser = playwright._initializer.get("preLaunchedBrowser")
+        assert pre_launched_browser
+        browser = cast(Browser, from_channel(pre_launched_browser))
+        browser._is_remote = True
+        browser._is_connected_over_websocket = True
+
+        transport.on("close", browser._on_close)
 
         return browser
 
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -23,7 +23,12 @@ from playwright._impl._api_structures import (
 )
 from playwright._impl._browser import Browser, normalize_context_params
 from playwright._impl._browser_context import BrowserContext
-from playwright._impl._connection import ChannelOwner, Connection, from_channel
+from playwright._impl._connection import (
+    ChannelOwner,
+    Connection,
+    from_channel,
+    from_nullable_channel,
+)
 from playwright._impl._helper import (
     BrowserChannel,
     ColorScheme,
@@ -177,10 +182,7 @@ class BrowserType(ChannelOwner):
         browser._is_remote = True
         browser._is_connected_over_websocket = True
 
-        def handle_transport_close() -> None:
-            browser._on_close()
-
-        transport.once("close", handle_transport_close)
+        transport.once("close", browser._on_close)
 
         return browser
 
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -161,9 +161,14 @@ class BrowserType(ChannelOwner):
     ) -> Browser:
         transport = WebSocketTransport(ws_endpoint, timeout)
 
-        connection = Connection(None, self._connection._object_factory, transport)
+        connection = Connection(
+            self._connection._dispatcher_fiber,
+            self._connection._object_factory,
+            transport,
+        )
         connection._loop = self._connection._loop
         connection._loop.create_task(connection.run())
+        self._connection._child_ws_connections.append(connection)
         playwright = await connection.wait_for_object_with_known_name("Playwright")
         pre_launched_browser = playwright._initializer.get("preLaunchedBrowser")
         assert pre_launched_browser
@@ -171,7 +176,15 @@ class BrowserType(ChannelOwner):
         browser._is_remote = True
         browser._is_connected_over_websocket = True
 
-        transport.on("close", browser._on_close)
+        def handle_transport_close() -> None:
+            browser._on_close()
+
+        transport.on("close", handle_transport_close)
+
+        browser.once(
+            Browser.Events.Disconnected,
+            lambda b: transport.remove_listener("close", handle_transport_close),
+        )
 
         return browser
 
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -166,6 +166,7 @@ class BrowserType(ChannelOwner):
             self._connection._object_factory,
             transport,
         )
+        connection._is_sync = True
         connection._loop = self._connection._loop
         connection._loop.create_task(connection.run())
         self._connection._child_ws_connections.append(connection)
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -180,12 +180,7 @@ class BrowserType(ChannelOwner):
         def handle_transport_close() -> None:
             browser._on_close()
 
-        transport.on("close", handle_transport_close)
-
-        browser.once(
-            Browser.Events.Disconnected,
-            lambda b: transport.remove_listener("close", handle_transport_close),
-        )
+        transport.once("close", handle_transport_close)
 
         return browser
 
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -166,7 +166,7 @@ class BrowserType(ChannelOwner):
             self._connection._object_factory,
             transport,
         )
-        connection._is_sync = True
+        connection._is_sync = self._connection._is_sync
         connection._loop = self._connection._loop
         connection._loop.create_task(connection.run())
         self._connection._child_ws_connections.append(connection)
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -44,7 +44,12 @@ class Channel(AsyncIOEventEmitter):
         if params is None:
             params = {}
         callback = self._connection._send_message_to_server(self._guid, method, params)
-        result = await callback.future
+
+        done, pending = await asyncio.wait(
+            {callback.future, self._connection._transport.on_error_future},
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+        result = next(iter(done)).result()
         # Protocol now has named return values, assume result is one level deeper unless
         # there is explicit ambiguity.
         if not result:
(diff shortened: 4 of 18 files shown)


playwright/_impl/_browser_type.py
```diff
@@ -135,6 +136,26 @@ async def launch_persistent_context(
                 raise not_installed_error(f'"{self.name}" browser was not found.')
             raise e
 
+    async def connect(
+        self, ws_endpoint: str, timeout: float = 30000, slow_mo: float = None
+    ) -> Browser:
+        transport = WebSocketTransport(ws_endpoint, timeout)
+        from playwright._impl._object_factory import create_remote_object
```
kumaraditya303, line 143: import at top

playwright/_impl/_transport.py
```diff
@@ -35,9 +38,25 @@ def _get_stderr_fileno() -> Optional[int]:
 
 
 class Transport:
+    def __init__(self) -> None:
```
kumaraditya303, line 41: Consider using abstract class using `ABC`

playwright/_impl/_transport.py
```diff
@@ -97,3 +116,43 @@ def send(self, message: Dict) -> None:
...
+        while not self._stopped:
+            try:
+                message = await self._connection.recv()
+                obj = json.loads(message)
+                self.on_message(obj)
+            except websockets.exceptions.ConnectionClosed:
+                self.emit("close")
+                break
+            except Exception as exc:
+                print("got unhandled exception", exc)
```
kumaraditya303, line 154: Raise Playwright `Error`

playwright/_impl/_transport.py
```diff
@@ -97,3 +116,43 @@ def send(self, message: Dict) -> None:
...
+
+    def send(self, message: Dict) -> None:
+        msg = json.dumps(message).encode()
+        asyncio.ensure_future(self._connection.send(msg))
```
kumaraditya303, line 158: Use Asyncio `create_task` for consistency

playwright/async_api/_generated.py
```diff
@@ -9472,6 +9472,36 @@ async def launch_persistent_context(
             )
         )
 
+    async def connect(
+        self, abc: str, *, timeout: float = None, slow_mo: float = None
```
kumaraditya303, line 9476: abc ? should be `ws_endpoint`

kumaraditya303 (changes requested):

playwright/_impl/_browser_type.py
mxschmitt: The reason for that was to quickly workaround the circular dependency. But I ended up now locally to use the object_factory from the parent connection. Seems also better.

playwright/_impl/_transport.py
mxschmitt: Seems what I want, thanks for pointing that out!

playwright/_impl/_transport.py
mxschmitt: Will probably remove that completely, was more for debugging. That's why its a draft PR, it was mostly not yet ready for review.

playwright/async_api/_generated.py
mxschmitt: This was just temporarily, with the next generation (next push9 its ws_endpoint.

tests/async/test_browsertype_connect.py
```diff
@@ -0,0 +1,179 @@
...
+import pytest
+
+from playwright._impl._driver import compute_driver_executable
+from playwright.async_api import BrowserType, Error
+from tests.server import Server
+
+
+class RemoteServer:
+    def __init__(self, browser_name: str) -> None:
+        self.process = subprocess.Popen(
```
kumaraditya303, line 29: Maybe use `asyncio` `create_subrocess_exec` as it is used on async context @mxschmitt

mxschmitt: Please do not mention me all the time with the @, this will result in a notification on my phone every time.
In this case I keep it because I want to use it in sync tests too. But thanks for your suggestions.

kumaraditya303: Ok I won't @mention you onwards, It is better to keep current as you also wanna use it to test in `sync` context

tests/async/test_browsertype_connect.py
```diff
@@ -0,0 +1,142 @@
...
+async def test_browser_type_connect_should_throw_when_used_after_is_connected_returns_false(
+    browser_type: BrowserType, launch_server
+):
+    remote_server = launch_server()
+    browser = await browser_type.connect(remote_server.ws_endpoint)
+    page = await browser.new_page()
+
+    remote_server.kill()
+
+    with pytest.raises(Error) as exc_info:
```
kumaraditya303, line 125: Consider using `match` with `raises` https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions

mxschmitt: Interesting, but going to keep it for now for consistency reasons in the codebase.

tests/sync/test_browsertype_connect.py
```diff
@@ -0,0 +1,141 @@
...
+
+@pytest.mark.skip
+def test_browser_type_connect_should_reject_navigation_when_browser_closes(
+    server: Server, browser_type: BrowserType, launch_server
+):
+    remote_server = launch_server()
+    browser = browser_type.connect(remote_server.ws_endpoint)
+    page = browser.new_page()
+    server.set_route("/one-style.css", lambda r: None)
+    page.on("request", lambda: browser.close())
```
mxschmitt, line 137: @kumaraditya303 maybe you have an idea here. browser.close blocks in our sync layer, that's why I marked this test as skipped. Will also take a look but maybe you are more experienced with greelet.

tests/async/test_browsertype_connect.py
kumaraditya303: :+1: It can be changed later in a atomic change in the codebase

tests/conftest.py
```diff
@@ -201,3 +204,43 @@ def compare(received_raw: bytes, golden_name: str):
         assert diff_pixels == 0
 
     return compare
+
+
+class RemoteServer:
+    def __init__(self, browser_name: str) -> None:
+        options = {}
+        if is_win():
```
kumaraditya303, line 212: Did you mean not `is_win` here as `os.setsid` is not available on windows

mxschmitt: oh yeah. I'm thinking about using node there directly, that would ease-up our life with killing the whole sub-tree of processes. See here how Java is doing it:

https://github.com/microsoft/playwright-java/blob/20322470b5f23698faf7b19704015c822861e360/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java#L57-L58

kumaraditya303: Yeah it would be better

tests/sync/test_browsertype_connect.py
kumaraditya303: Will take a look after the tests are fixed

tests/conftest.py
```diff
@@ -201,3 +205,41 @@ def compare(received_raw: bytes, golden_name: str):
...
+            stdout=subprocess.PIPE,
+            stderr=sys.stderr,
+            cwd=driver_dir,
+        )
+        assert self.process.stdout
+        self.ws_endpoint = self.process.stdout.readline().decode().strip()
+
+    def kill(self):
+        # Send the signal to all the process groups
+        self.process.send_signal(signal.SIGKILL)
```
kumaraditya303, line 229: `SIGKILL` not available on windows

tests/conftest.py
```diff
@@ -201,3 +204,44 @@ def compare(received_raw: bytes, golden_name: str):
...
+            stderr=sys.stderr,
+            cwd=driver_dir,
+        )
+        assert self.process.stdout
+        self.ws_endpoint = self.process.stdout.readline().decode().strip()
+
+    def kill(self):
+        # Send the signal to all the process groups
+        if sys.platform == "win32":
+            subprocess(f"taskkill /F /PID {self.process.pid}")
```
kumaraditya303, line 229: Missing `check_call` 😆

mxschmitt: Please do not comment everything, I will see it when its red on the bots :)

tests/sync/test_browsertype_connect.py
mxschmitt: should be mostly ready, the only issue which remains that test_browser_type_connect_should_reject_navigation_when_browser_closes timeouts sometimes, will try to debug it later. The other failing tests are general flaky tests which we can fix/investigate also after this PR is merged. Would be nice if there would be more green runs.

kumaraditya303:
I was able to fix the last skipped test `test_browser_type_connect_should_reject_navigation_when_browser_closes` with this code, what it does essentially is runs the greenlet nested with other greenlets hence it does not block, you can try the code given below also it is passing the tests locally to me:
```python

def test_browser_type_connect_should_reject_navigation_when_browser_closes(
    server: Server, browser_type: BrowserType, launch_server
):
    import greenlet

    remote_server = launch_server()
    browser = browser_type.connect(remote_server.ws_endpoint)
    page = browser.new_page()
    server.set_route("/one-style.css", lambda r: None)
    g = greenlet.greenlet(browser.close)
    page.on("request", lambda: g.switch())

    with pytest.raises(Error) as exc_info:
        page.goto(server.PREFIX + "/one-style.html")
    assert "Playwright connection closed" in exc_info.value.message

```

playwright/_impl/_transport.py
```diff
@@ -34,10 +40,46 @@ def _get_stderr_fileno() -> Optional[int]:
...
+
+    async def run(self) -> None:
+        self._loop = asyncio.get_running_loop()
+        self.on_error_future: asyncio.Future[bool] = asyncio.Future()
+
+    @abstractmethod
+    def send(self, message: Dict) -> None:
+        pass
+
+    def serialize_message(self, message: Dict) -> Any:
```
kumaraditya303, line 66: The return type should be `bytes`

playwright/_impl/_transport.py
```diff
@@ -79,21 +121,71 @@ async def run(self) -> None:
...
+                print(f"Received unhandled exception: {exc}")
+                self.on_error_future.set_exception(exc)
+
+    def send(self, message: Dict) -> None:
+        self.test_if_closed()
+        data = self.serialize_message(message)
+        self._loop.create_task(self._connection.send(data))
+
+    def test_if_closed(self) -> None:
+        if self._stopped or self._connection.closed:
```
kumaraditya303, line 190: You can skip this one: It would be better to inline this method as it is used only once

tests/conftest.py
```diff
@@ -201,3 +207,60 @@ def compare(received_raw: bytes, golden_name: str):
...
+            stderr=sys.stderr,
+            cwd=driver_dir,
+        )
+        assert self.process.stdout
+        self.ws_endpoint = self.process.stdout.readline().decode().strip()
+
+    def kill(self):
+        # Send the signal to all the process groups
+        if sys.platform == "win32":
+            os.system(f"taskkill /F /PID {self.process.pid}")
```
kumaraditya303, line 241:
```suggestion
            subprocess.check_call(["taskkill", "/F", "PID", f"{self.process.pid}"])
```

kumaraditya303 (approved):
The failing tests seems flaky

playwright/_impl/_browser.py
```diff
@@ -145,6 +146,10 @@ async def new_page(
         return page
 
     async def close(self) -> None:
+        if self._is_connected_over_websocket:
```
pavelfeldman, line 149: Do you want to move it to under _is_closed_or_closing = True? Also, why is logic different from the one in TS? Does TS close the browser on browser.close?

To me, the ideal workflow would be: `close` terminates the underlying connection, connection termination calls _on_close everywhere. That way, terminating connection due to networking issues would result in the same proper cleanup.

playwright/_impl/_browser.py
```diff
@@ -154,6 +159,14 @@ async def close(self) -> None:
             if not is_safe_close_error(e):
                 raise e
 
+    def _notify_remote_closed(self) -> None:
+        # Emulate all pages, contexts and the browser closing upon disconnect.
```
pavelfeldman, line 163: I.e I would expect this to be called on web socket disconnect somehow.

playwright/_impl/_browser_type.py
```diff
@@ -135,6 +136,39 @@ async def launch_persistent_context(
...
+        browser = cast(Browser, from_channel(pre_launched_browser))
+        browser._is_remote = True
+        browser._is_connected_over_websocket = True
+
+        def handle_transport_close() -> None:
+            browser._on_close()
+
+        transport.on("close", handle_transport_close)
+
+        browser.once(
```
pavelfeldman, line 165: Given that _on_close that dispatches this event is only called from _on_close above, why do we need this remote_listener?

playwright/_impl/_connection.py
```diff
@@ -168,10 +179,16 @@ async def run(self) -> None:
...
         self._transport.stop()
         self._dispatcher_fiber.switch()
+        self.cleanup()
 
     async def stop_async(self) -> None:
         self._transport.stop()
         await self._transport.wait_until_stopped()
+        self.cleanup()
+
+    def cleanup(self) -> None:
```
pavelfeldman, line 189: You would typically call this `dispose`.

playwright/_impl/_connection.py
```diff
@@ -168,10 +179,16 @@ async def run(self) -> None:
...
+        for ws_connection in self._child_ws_connections:
+            ws_connection._transport.cleanup()
```
pavelfeldman, line 191: ditto

playwright/_impl/_transport.py
```diff
@@ -34,10 +40,46 @@ def _get_stderr_fileno() -> Optional[int]:
...
+    def cleanup(self) -> None:
+        pass
+
+    @abstractmethod
+    async def wait_until_stopped(self) -> None:
+        pass
+
+    async def run(self) -> None:
+        self._loop = asyncio.get_running_loop()
+        self.on_error_future: asyncio.Future[bool] = asyncio.Future()
```
pavelfeldman, line 60: Is Future typed in 3.7?

playwright/_impl/_transport.py
```diff
@@ -34,10 +40,46 @@ def _get_stderr_fileno() -> Optional[int]:
         return sys.__stderr__.fileno()
 
 
-class Transport:
+class Transport(ABC):
+    def __init__(self) -> None:
+        self.on_message = lambda _: None
+
+    @abstractmethod
+    def stop(self) -> None:
```
pavelfeldman, line 48: let's rename to `request_stop` for clarity.

playwright/_impl/_transport.py
```diff
@@ -79,21 +121,73 @@ async def run(self) -> None:
...
+        Transport.__init__(self)
+
+        self._stopped = False
+        self.ws_endpoint = ws_endpoint
+        self.timeout = timeout
+        self._loop: asyncio.AbstractEventLoop
+
+    def stop(self) -> None:
+        self._stopped = True
+        self._loop.create_task(self._connection.close())
```
pavelfeldman, line 151: Declare self._connection in __init__, did our mypy break?

playwright/_impl/_browser_type.py
```diff
@@ -135,6 +136,39 @@ async def launch_persistent_context(
...
+        )
+        connection._is_sync = self._connection._is_sync
+        connection._loop = self._connection._loop
+        connection._loop.create_task(connection.run())
+        self._connection._child_ws_connections.append(connection)
+        playwright = await connection.wait_for_object_with_known_name("Playwright")
+        pre_launched_browser = playwright._initializer.get("preLaunchedBrowser")
+        assert pre_launched_browser
+        browser = cast(Browser, from_channel(pre_launched_browser))
+        browser._is_remote = True
```
pavelfeldman, line 157: Now that you are adding it, we need to introduce the code that depends on it. That would be our artifacts, namely 'download' and 'video'.

playwright/_impl/_transport.py
mxschmitt: removed, since we have it never like that and in this case I only use it to throw an exception.

playwright/_impl/_transport.py
mxschmitt: Yes its broken currently, will follow up to fix it.

playwright/_impl/_browser.py
mxschmitt: the naming came originally from Java, but open to change.

pavelfeldman: So I looked at the TS version, looks like you don't need any special code here, just close the browser. But our server seems to be missing the connection termination upon close for some reason.

playwright/_impl/_browser.py
mxschmitt: move it below it.

TS does it via an event, see here in another place: https://github.com/microsoft/playwright/blob/fff1f3d45c243bec65c7e6c43e6fe5e36607dc59/src/client/browserType.ts#L163-L179

playwright/_impl/_browser_type.py
mxschmitt: Seems not needed anymore, was from a try to fix the flaky tests.

tests/async/test_browsertype_connect.py
```diff
@@ -0,0 +1,143 @@
...
+    remote_server = launch_server()
+    browser = await browser_type.connect(remote_server.ws_endpoint)
+    page = await browser.new_page()
+    await browser.close()
+
+    with pytest.raises(Error) as exc_info:
+        await page.goto(server.PREFIX + "/one-style.html")
+    assert "Playwright connection closed" in exc_info.value.message
+
+    with pytest.raises(Error) as exc_info:
```
kumaraditya303, line 141: Remove the last assertions as it is ducplicated

local-requirements.txt
```diff
@@ -11,8 +11,9 @@ pixelmatch==0.2.3
 pre-commit==2.10.1
 pyOpenSSL==20.0.1
 pytest==6.2.2
-pytest-asyncio==0.14.0
+pytest-asyncio==0.15.0
 pytest-cov==2.11.1
+pytest-repeat==0.9.1
```
kumaraditya303, line 16: Where is this new plugin used ?

playwright/_impl/_transport.py
mxschmitt: It can automatically infer that I set it in WebSocketTransport.run() 

But yes your assumption was also correct that mypy was broken. Fixed in #643

tests/async/test_browsertype_connect.py
mxschmitt: intended to ensure it works multiple times.

local-requirements.txt
mxschmitt: it's helpful for debugging flaky tests.

playwright/_impl/_browser_type.py
mxschmitt: good catch, added it.

playwright/_impl/_browser.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         self._is_connected = True
         self._is_closed_or_closing = False
         self._is_remote = False
+        self._is_connected_over_websocket = False
```
pavelfeldman, line 53: It seems like you can use is_remote here.

playwright/_impl/_browser_type.py
```diff
@@ -157,6 +159,30 @@ async def connect_over_cdp(
...
+        browser._is_connected_over_websocket = True
```
pavelfeldman, line 183: Yeah, it looks the same as is_remote...

playwright/_impl/_browser_type.py
```diff
@@ -157,6 +159,30 @@ async def connect_over_cdp(
...
+
+        transport.once("close", browser._on_close)
```
pavelfeldman, line 185: nit: this looks like a layering violation, should go transport -> connection -> browser. Not a big deal though.

playwright/_impl/_connection.py
pavelfeldman: ...and you did not :)

pavelfeldman (approved):

playwright/_impl/_browser.py
mxschmitt: remote is also used for connect_over_cdp, we only want to close the underlying connection when connect is used.

x0day:
I test code from master(https://github.com/microsoft/playwright-python/commit/9d796623a104e082f57878d29d9aea20009b88c1), but `BrowserType.connect` not handle the `ConnectionRefusedError` ?

mxschmitt:
Good catch, will take a look.

PR #640, opened 2021-04-22
feat: added BrowserType.connect_over_cdp

mxschmitt:


tests/async/test_browsertype_connect_cdp.py
```diff
@@ -0,0 +1,88 @@
...
+    assert len(cdp_browser1.contexts[0].pages) == 2
+    assert len(cdp_browser2.contexts[0].pages) == 2
+
+    await cdp_browser1.close()
+    await cdp_browser2.close()
+    await browser_server.close()
+
+
+def _ws_endpoint_from_url(url: str) -> str:
+    response = requests.get(url)
```
kumaraditya303, line 66: Can be done using `urllib.request.urlopen` and then json.loads

kumaraditya303 (approved):
:+1:

tests/async/test_browsertype_connect_cdp.py
mxschmitt: using requests since its more convenient

PR #644, opened 2021-04-23
test: unskip tests

mxschmitt:


mxschmitt:
note: ran it 3 times an all of them it looked fine to me

PR #648, opened 2021-04-25
chore: added headers param for BrowserType.connect

kumaraditya303:


mxschmitt (commented):
What is the reason of changing that? Does it fix something?

kumaraditya303:
Calling `str` and `Path` is redundant and in case  could result in not resolve symlink as pathlib requires `Path.resolve()` for symlink resolving. Python has a dedicated function for it `os.fspath`  which is more appropriate for it

tests/async/test_download.py
```diff
@@ -194,7 +194,7 @@ async def test_should_error_when_saving_after_deletion(tmpdir, browser, server):
     await download.delete()
     with pytest.raises(Error) as exc:
         await download.save_as(user_path)
-    assert "File already deleted. Save before deleting." in exc.value.message
```
mxschmitt, line 197: this is weird, that seems not expected, but probably related to the roll and not to your changes.

mxschmitt: Edit: fine was changed upstream too.

mxschmitt (approved):

PR #649, opened 2021-04-25
fix: recursively include packages for conda using setuptools

kumaraditya303:


mxschmitt (commented):
Either you use
```
find_packages(exclude=["tests*", "playwright._*"])
```

which will result in

```
['playwright', 'playwright.async_api', 'playwright.sync_api']
```
or you just hard-code the 3 output values of it. Having it like you suggested would result in 
```
['playwright', 'tests.common', 'tests.async', 'tests.sync', 'playwright.async_api', 'playwright.sync_api', 'playwright._impl']
```

which is not expected from my understanding.

kumaraditya303:
Did you mean `include` instead of `exclude` ?

mxschmitt:
> Did you mean `include` instead of `exclude` ?

Just having `packages=['playwright', 'playwright.async_api', 'playwright.sync_api']` as before and adding sync/async packages to the list which should result in the same and has less magic.

kumaraditya303:
`packages=['playwright', 'playwright.async_api', 'playwright.sync_api']` do you want to hard code each value, in that case we do need `_impl` package

mxschmitt:
> `packages=['playwright', 'playwright.async_api', 'playwright.sync_api']` do you want to hard code each value, in that case we do need `_impl` package

ah I guessed impl was not expected there. :+1: Then feel free to adjust it to `find_packages(exclude=["tests*"])`

kumaraditya303:
Ok

mxschmitt (approved):

Issue #650, opened 2021-04-26
[internal] hang bug in PlaywrightContextManager and Transport


x0day:
with recently [feature(pull/630),](https://github.com/microsoft/playwright-python/pull/630) playwright-python can connect to remote websocket endpoint, but with this feature, a bug become obvious. 

`playwright-python` doesn't handle the transport's connection error during the connecting, if error found during the connecting. the task will be hang, because below code will never return.

https://github.com/microsoft/playwright-python/blob/master/playwright/async_api/_context_manager.py#L37

```
self._connection.wait_for_object_with_known_name("Playwright")
```

eg:

`Connect call failed` in `WebSocketTransport`.
 
maybe we should wait the `obj` and `Transport.on_error_future` same time in `PlaywrightContextManager` like this:

```python
class WebSocketTransport(AsyncIOEventEmitter, Transport):
    ...
    async def run(self) -> None:
        ...

        try:
            self._connection = await websockets.connect(self.ws_endpoint, **options)
        except Exception as exc:
            self.emit("close")
            self.on_error_future.set_exception(exc)
       ...
```

```python
class PlaywrightContextManager:
    ...
    async def __aenter__(self) -> AsyncPlaywright:
        ...
        done, pending = await asyncio.wait(
            {
                self._connection._transport.on_error_future,
                self._connection.wait_for_object_with_known_name("Playwright"),
            },
            return_when=asyncio.FIRST_COMPLETED,
        )
        obj = next(iter(done)).result()
        playwright = AsyncPlaywright(obj)
        ...
```

PR #651, opened 2021-04-26
fix(transport): handle connection error correctly (#650)

x0day:
#650


playwright/_impl/_transport.py
```diff
@@ -96,14 +96,19 @@ async def run(self) -> None:
...
+        try:
+            self._proc = proc = await asyncio.create_subprocess_exec(
+                str(self._driver_executable),
+                "run-driver",
+                stdin=asyncio.subprocess.PIPE,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=_get_stderr_fileno(),
+                limit=32768,
+            )
+        except FileNotFoundError:
```
kumaraditya303, line 108: In what case do you think this exception would be raised because on PyPI only wheels are uploaded and there's no chance of not having playwright driver installed ?

mxschmitt:
A test for ws connection timeout for sync and async would be awesome.

x0day:
> A test for ws connection timeout for sync and async would be awesome.

ok

playwright/_impl/_transport.py
x0day: first time if you clone the code from github, doesn't install the driver, run the pytest command and it hangs.

kumaraditya303: If you want to install from github repo this is what you should be really doing: 
```console
$ pip install git+https://github.com/microsoft/playwright-python
```

x0day: I want to fix the issue, so I need to test the source code and find the bug.

kumaraditya303: Then, You can suggest to run `python setup.py bdist_wheel` which would fix the issue rather than just saying that it is not there

x0day: yes! it's my mistake.

kumaraditya303: You can read the contributing guide for some guidance on how to get everything setup for working on the code  https://github.com/microsoft/playwright-python/blob/master/CONTRIBUTING.md

playwright/async_api/_context_manager.py
```diff
@@ -33,9 +33,15 @@ async def __aenter__(self) -> AsyncPlaywright:
         loop = asyncio.get_running_loop()
         self._connection._loop = loop
         loop.create_task(self._connection.run())
-        playwright = AsyncPlaywright(
-            await self._connection.wait_for_object_with_known_name("Playwright")
+        done, pending = await asyncio.wait(
```
x0day, line 36: whatever you use remote server or local pipe service, the `PipeTransport` will always started and this code will not work, so maybe we should give a option for not start the local process when use remote endpoint?

mxschmitt: yep would be nice but so far not doable, because BrowserType depends currently on the rpc channel logic.

x0day: I want to move this in `launch`, so we can wait the `Playwright` and the transport connection error in `launch` and `connect`

x0day: I can't just move to `launch`, because lots things depends on `Channel`, maybe we should re-design `PlaywrightImpl` and `PlaywrightContextManager`, allow channel can be inited after `connect` or `launch`.

x0day:
add the test test_connect_to_closed_server_without_hangs, please check.

playwright/async_api/_context_manager.py
x0day: fixed by catch the error in `connect`

playwright/_impl/_browser_type.py
```diff
@@ -171,11 +172,24 @@ async def connect(
             self._connection._object_factory,
             transport,
         )
+        await connection._transport.start()
         connection._is_sync = self._connection._is_sync
         connection._loop = self._connection._loop
         connection._loop.create_task(connection.run())
+        obj = asyncio.ensure_future(
```
kumaraditya303, line 179:
```suggestion
        obj = asyncio.create_task(
```
Use `asyncio.create_task` for consistency for rest of the codebase

playwright/_impl/_transport.py
```diff
@@ -55,9 +56,13 @@ def dispose(self) -> None:
...
         pass
 
-    async def run(self) -> None:
+    async def start(self) -> None:
+        if not hasattr(self, "on_error_future"):
+            self.on_error_future = asyncio.Future()
         self._loop = asyncio.get_running_loop()
-        self.on_error_future: asyncio.Future = asyncio.Future()
+
+    async def run(self) -> None:
```
kumaraditya303, line 64: Make it a `abstractmethod`

kumaraditya303 (changes requested):

kumaraditya303 (approved):

mxschmitt (commented):
Looks good but the currently handling of having run and start is confusing I think from my perspective. Will take a more detailed look later how to avoid that.

x0day:
> Looks good but the currently handling of having run and start is confusing I think from my perspective. Will take a more detailed look later how to avoid that.

yep maybe we can fix hang error at this time and think about  allow `channel` be inited after `connect` or launch at next version.

mxschmitt:
Hey @x0day thanks for your Pull Request, I opened another PR #660 for a slightly different implementation. Should also fix the issue.

mxschmitt:
Hey @x0day thanks for your Pull Request, I opened another PR #660 for a slightly different implementation. Should also fix the issue.

mxschmitt:
Hey @x0day if you want you can apply the changes from #660 into your PR, so we can merge yours instead of mine. I did there some enhancements.

PR #657, opened 2021-04-30
feat: conda support

kumaraditya303:


kumaraditya303:
Conda update: We have decided to publish playwright conda recipe in the `microsoft` channel https://github.com/microsoft/playwright-python/pull/657 you can check that PR for progress. This way we get a finer control over the recipe rather than replying on a third party repo as for conda forge also this way it allows more customisation which otherwise were not possible. @fcollonval

mxschmitt:
Status-update: waiting for credentials on our side.

meta.yaml
```diff
@@ -0,0 +1,58 @@
+package:
+  name: playwright
+  version: "{{ environ.get('GIT_DESCRIBE_TAG') }}"
+
+source:
+  path: .
+
+build:
```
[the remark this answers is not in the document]
kumaraditya303: Playwright conda package is not so simple, and to be clear playwright package is not cross platform, the wheels are platform dependent hence noarch cannot be used here.

.github/workflows/ci.yml
```diff
@@ -163,3 +163,23 @@ jobs:
       with:
         name: ${{ matrix.browser-channel }}-${{ matrix.os }}
         path: pw-log.txt
+
+  build-conda:
```
mxschmitt, line 167: maybe we can only run it when the setup.py is changed? Since it contains the playwright driver version and that would avoid us having 3 extra bots for each run.

meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  summary: Python version of the Playwright testing and automation library.
+  description: |
+    Playwright is a Python library to automate Chromium,
+    Firefox and WebKit browsers with a single API. Playwright
+    delivers automation that is ever-green, capable, reliable
+    and fast. See how Playwright is better.
+  doc_url: https://playwright.dev/python/docs/intro/
+  dev_url: https://github.com/microsoft/playwright-python
+
+extra:
```
mxschmitt, line 55: this extra -> recipe-maintainers seems conda-forge specific so we can remove it.

.github/workflows/ci.yml
```diff
@@ -163,3 +163,23 @@ jobs:
...
+    name: Conda Build
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
mxschmitt, line 172: I would prefer there using the exact versions, sometimes they change it and then its's e.g. broken.
```suggestion
        os: [ubuntu-20.04, macos-10.15, windows-2019]
```

Updating and bumping them manually every 2 years is preferred.

mxschmitt (approved):
LGTM

.github/workflows/publish.yml
```diff
@@ -6,20 +6,42 @@ jobs:
...
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Get conda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
```
mxschmitt, line 38: lets specify the Python version there too? 3.8 since 3.9 has some bugs in combination conda build.
e.g. https://github.com/Anaconda-Platform/anaconda-client/issues/555

.github/workflows/ci.yml
```diff
@@ -163,3 +163,23 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - name: Get conda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          channels: conda-forge
+      - name: Prepare
+        run: conda install conda-build conda-verify
+      - name: Build
+        run: conda build .
```
mxschmitt, line 185:
```suggestion
        run: conda build --user microsoft .
```
not sure if this needed but I used it locally. Since a token is for a user we need to _somehow_ explicitly say that it should use the microsoft group.

mxschmitt:
Would also be nice if you could add some docs, I thought about a badge:

Conda

(with a link to anaconda.org)

and maybe under usage a code block for anaconda with `conda install -c microsoft playwright`.

.github/workflows/publish.yml
kumaraditya303: I tried it and it worked without even so I didn't added that, would add in next commit

meta.yaml
kumaraditya303: Will follow up

.github/workflows/publish.yml
mxschmitt: Locally it failed uploading for me yesterday silently without any errors which took me a hour to figure out. So I expect it also failing uploading in the CI. :/ I want to avoid recreating tags etc

Talking about releasing, we can extend the deployment GitHub workflow by a custom event trigger so we can trigger it manually later on the release branch to upload v1.12 from GHA directly: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

.github/workflows/ci.yml
kumaraditya303: Better to keep it for now since conda build is not as same as pip, remember that setuptools find_package one which was different from pip, and hence we would be safe from potential breakage

zooba:
> I thought about a badge

Not sure if you need to be logged in to see it, but there are badges provided by anaconda.org at https://anaconda.org/Microsoft/playwright/badges

PR #658, opened 2021-05-01
feat: logging using logging module

### LINKED ISSUE #440: Using logger with playwright-python
I can currently get logger output on screen using `DEBUG=pw:api` environment variable.

However, it is possible to get the output to a file specified by the caller using the logging module?

For e.g., I am building a custom RobotFramework library and want the Playwright logging to get merged into RobotFramework HTML log output.

kumaraditya303:
Closes #440
With this change users can make use of logging module to customise it as per their requirement:
Test logging with this script:
```python
import logging

logging.getLogger("playwright").setLevel(logging.DEBUG)
logging.getLogger("playwright").addHandler(logging.StreamHandler())
from pathlib import Path
from playwright.sync_api import Playwright, sync_playwright


def main(playwright: Playwright) -> None:
    for browser_type in [playwright.chromium, playwright.firefox, playwright.webkit]:
        page = browser.new_page()
        here = Path(__file__).parent.resolve()
        page.screenshot(path=here / f"{browser_type.name}.png")
        page.close()
        browser.close()


if __name__ == "__main__":
    with sync_playwright() as p:
        main(p)
```


kumaraditya303:
@pavelfeldman @mxschmitt Ready for review!

mxschmitt (approved):
LGTM

pavelfeldman:
This looks good overall! Let's Align it with our Node version a bit:
- Instead of generating code like this, add logging into `_sync` / `_async` in the `SyncBase` / `AsyncBase`.
- Log it in the same form as `_wrapApiCall` does in Node version

kumaraditya303:
@pavelfeldman done!

playwright/_impl/_async_base.py
```diff
@@ -56,10 +57,16 @@ def __str__(self) -> str:
...
-        task = asyncio.current_task()
-        setattr(task, "__pw_api_name__", api_name)
-        setattr(task, "__pw_stack_trace__", traceback.extract_stack())
-        return coro
+        try:
+            logger.debug(f"=> {api_name} started")
+            task = asyncio.current_task()
+            setattr(task, "__pw_api_name__", api_name)
+            setattr(task, "__pw_stack_trace__", traceback.extract_stack())
+            logger.debug(f"<= {api_name} succeeded")
```
pavelfeldman, line 65: Since you don't await coro in the middle, your log will be always saying started/succeeded w/o anything in between. The actual task will only start after you say succeeded.

kumaraditya303: The coroutine is awaited in the generated in the `_generated.py` file hence that's why in the previous commit I had done logging in the docs generator. Will revert back to it since it works without any issues!

mxschmitt: You can also await it inside to resolve the Awaitable of the specific function inside it. Would look like that:

sync:
```py
            ...
            asyncio._set_running_loop(self._loop)
            return_value = task.result()
            logger.debug(f"<= {api_name} succeeded")
            return return_value
        except Exception as e:
            logger.debug(f"<= {api_name} failed")
            raise e
```

async:
```py
    async def _async(self, api_name: str, coro: Awaitable) -> Any:
        try:
            logger.debug(f"=> {api_name} started")
            task = asyncio.current_task()
            setattr(task, "__pw_api_name__", api_name)
            setattr(task, "__pw_stack_trace__", traceback.extract_stack())
            return_value = await coro
            logger.debug(f"<= {api_name} succeeded")
            return return_value
        except Exception as e:
            logger.debug(f"<= {api_name} failed")
            raise e
```

and add please a failed test with something like that:

```py
                   try:
                        page.wait_for_selector("foo", timeout=100)
                    except Error:
                        pass
```

which will result in a failed API call which throws and logs an failed log entry.

mxschmitt (approved):

tests/common/test_logging.py
```diff
@@ -0,0 +1,94 @@
...
+                    page = browser.new_page()
+                    page.goto("https://example.com")
+                    with pytest.raises(Error):
+                        page.wait_for_selector("will never find", timeout=100)
+                    page.close()
+                    browser.close()
+            self.assertEqual(
+                logger.output,
+                [
+                    "DEBUG:playwright:=> browser_type.launch started",
```
pavelfeldman, line 41: This is not how the pw:api looks typically, there are log entries inside it as well. I think Python users will be surprised not to see them:

Screen Shot 2021-05-05 at 12 01 14 PM

kumaraditya303: This is how it is done in `playwright-java` too 
https://github.com/microsoft/playwright-java/blob/0fa416b45931a3db254665f84a5c8375edf67b58/playwright/src/main/java/com/microsoft/playwright/impl/LoggingSupport.java#L41 We can add more logging in another PR but let this one get merged first

PR #664, opened 2021-05-06
feat(roll): roll Playwright to 1.11.0-1620262237000

kumaraditya303:


tests/sync/test_accessibility.py
```diff
@@ -216,57 +215,95 @@ def test_accessibility_filtering_children_of_leaf_nodes_rich_text_editable_field
...
-    page,
+    page, browser_channel
 ):
     page.set_content(
         """
     <div contenteditable="plaintext-only" tabIndex=0>Edit this image:<img src="fakeimage.png" alt="my fake image"></div>"""
     )
     snapshot = page.accessibility.snapshot()
-    assert snapshot["children"][0] == {"role": "generic", "name": ""}
+    if browser_channel:
```
mxschmitt, line 299: this is going to break when the channels will get updated. It would be better to check by the Chromium version, see this patch which needs to get applied: https://github.com/microsoft/playwright/pull/6372

kumaraditya303: Yeah it would break once channel is updated which would notify us that the special case for it should be removed, if we check by version then it would always pass the tests and never break. I think it is better the way since this is not the actual library code but rather tests.

mxschmitt:
Please use 1.11.0-1620331022000 as a version. It's the latest which includes some more fixes.

kumaraditya303:
@mxschmitt done!

PR #665, opened 2021-05-06
fix: restrict installation to Ubuntu 18.04

kumaraditya303:
With [PEP 600](https://github.com/mayeut/pep600_compliance/blob/master/DETAILS.rst) this PR restricts installation to Ubuntu 18.04 or higher.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,7 +92,7 @@ class PlaywrightBDistWheelCommand(BDistWheelCommand):
             if platform == "mac":
                 wheel = "macosx_10_13_x86_64.whl"
             if platform == "linux":
-                wheel = "manylinux1_x86_64.whl"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -92,7 +92,7 @@ def run(self) -> None:
             if platform == "mac":
                 wheel = "macosx_10_13_x86_64.whl"
             if platform == "linux":
-                wheel = "manylinux1_x86_64.whl"
+                wheel = "manylinux_2_27_x86_64.whl"
```
mxschmitt, line 95: Note for other reviewers. This means that at least glibc version 2.27 is required which is used in Ubuntu 18.04 (verified).

mxschmitt (approved):

pavelfeldman (approved):

Issue #673, opened 2021-05-07
[Feature] BrowserType.connect slowMo is missing


mxschmitt:
Upstream: https://github.com/microsoft/playwright/blob/5e4badd654e0064aa5ca1877f46411b1ee0f9ceb/src/client/browserType.ts#L122-L124

PR #674, opened 2021-05-08
fix: added BrowserType.connect slow_mo param

kumaraditya303:
Fixes https://github.com/microsoft/playwright-python/issues/673


tests/sync/test_browsertype_connect.py
```diff
@@ -12,12 +12,28 @@
...
+):
+    remote_server = launch_server()
+    t1 = time.monotonic()
+    browser = browser_type.connect(remote_server.ws_endpoint, slow_mo=100)
+    browser_context = browser.new_context()
+    page = browser_context.new_page()
+    assert page.evaluate("11 * 11") == 121
+    page.goto(server.EMPTY_PAGE)
+    browser.close()
+    assert (time.monotonic() - t1) > 3.8
```
mxschmitt, line 34: 3.8 seems like it would result in flakyness, what about testing it just between the evaluate function instead of the page.goto, close etc.?

kumaraditya303: It would not flake as I have counted the number of sleep call and it sums up to 3.8 also there is greater than so it won't

PR #682, opened 2021-05-12
feat(roll): roll to Playwright 1.12.0-next-1621456974000

kumaraditya303:
Roll Playwright driver to 1.12.0-next-1620880294000

### DIFF
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -148,6 +148,12 @@ class Browser(ChannelOwner):
     async def close(self) -> None:
         if self._is_closed_or_closing:
             return
+        if self._is_remote:
+            for context in self.contexts:
+                for page in context.pages:
+                    page._on_close()
+                context._on_close()
+            self._on_close()
         self._is_closed_or_closing = True
         try:
             await self._channel.send("close")
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -148,12 +148,6 @@ class Browser(ChannelOwner):
     async def close(self) -> None:
         if self._is_closed_or_closing:
             return
-        if self._is_remote:
-            for context in self.contexts:
-                for page in context.pages:
-                    page._on_close()
-                context._on_close()
-            self._on_close()
         self._is_closed_or_closing = True
         try:
             await self._channel.send("close")
--- a/playwright/_impl/_browser_context.py
+++ b/playwright/_impl/_browser_context.py
@@ -35,6 +35,7 @@ from playwright._impl._helper import (
 )
 from playwright._impl._network import Request, Route, serialize_headers
 from playwright._impl._page import BindingCall, Page, Worker
+from playwright._impl._tracing import Tracing
 from playwright._impl._wait_helper import WaitHelper
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -64,7 +65,7 @@ class BrowserContext(ChannelOwner):
         self._options: Dict[str, Any] = {}
         self._background_pages: Set[Page] = set()
         self._service_workers: Set[Worker] = set()
-
+        self._tracing = Tracing(self)
         self._channel.on(
             "bindingCall",
             lambda params: self._on_binding(from_channel(params["binding"])),
@@ -79,7 +80,6 @@ class BrowserContext(ChannelOwner):
                 from_channel(params.get("route")), from_channel(params.get("request"))
             ),
         )
-
         self._channel.on(
             "backgroundPage",
             lambda params: self._on_background_page(from_channel(params["page"])),
@@ -239,7 +239,7 @@ class BrowserContext(ChannelOwner):
 
     def _on_close(self) -> None:
         self._is_closed_or_closing = True
-        if self._browser:
+        if self._browser and self in self._browser.contexts:
             self._browser._contexts.remove(self)
 
         self.emit(BrowserContext.Events.Close)
@@ -299,3 +299,7 @@ class BrowserContext(ChannelOwner):
         return from_channel(
             await self._channel.send("newCDPSession", {"page": page._channel})
         )
+
+    @property
+    def tracing(self) -> Tracing:
+        return self._tracing
--- a/playwright/_impl/_browser_context.py
+++ b/playwright/_impl/_browser_context.py
@@ -22,7 +22,11 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Unio
 from playwright._impl._api_structures import Cookie, Geolocation, StorageState
 from playwright._impl._api_types import Error
 from playwright._impl._cdp_session import CDPSession
-from playwright._impl._connection import ChannelOwner, from_channel
+from playwright._impl._connection import (
+    ChannelOwner,
+    from_channel,
+    from_nullable_channel,
+)
 from playwright._impl._event_context_manager import EventContextManagerImpl
 from playwright._impl._helper import (
     RouteHandler,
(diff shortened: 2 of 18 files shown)


playwright/_impl/_transport.py
```diff
@@ -153,7 +153,6 @@ def __init__(
     ) -> None:
         super().__init__(loop)
         Transport.__init__(self, loop)
-
```
mxschmitt, line 156: please revert this, since there is no actual change in this file.

playwright/_impl/_browser.py
```diff
@@ -148,6 +148,12 @@ async def new_page(
     async def close(self) -> None:
         if self._is_closed_or_closing:
             return
+        if self._is_remote:
```
mxschmitt, line 151: It's possible that the WebSocket connection get's disconnected without the user's control. This will end up that the Browser.close method get's never called. Moving the loop to when the WebSocket connection actually closes would fix it, see here how it's done upstream:

https://github.com/microsoft/playwright/blob/e018c5dd426c7cef9bbf6eb497112083013bb6ba/src/client/browserType.ts#L170-L179

Seems like there is no test yet for it, having one for that would be neat.

mxschmitt:
Please make the file change from CRLF to LF in another PR.

playwright/_impl/_browser.py
mxschmitt: Upstream we have this loop only a single time, would be good if we could have it here also at a single place only. It feels wrong to have it inside: _on_close.

kumaraditya303: When the transport closes it calls `_on_close` when websocket connection is closed

tests/async/test_tracing.py
```diff
@@ -103,3 +103,18 @@ async def test_should_support_a_buffer_without_a_path(
     await page.goto(server.PREFIX + "/grid.html")
     trace = await browser.stop_tracing()
     assert "screenshot" in trace.decode()
+
+
+@pytest.mark.only_browser("chromium")
```
mxschmitt, line 108: this is different from the existing tracing, so it's not Chromium specific. Also it might make sense to move the existing tracing into a `test_chromium_tracing.py` and the new tracing into `test_tracing.py`

playwright/_impl/_browser.py
mxschmitt: Upstream we do it next to the ws close event and not in the browser, I would move it into our browser_type.py when we receive the disconnect event from the WS client.

See here how it's done upstream: https://github.com/microsoft/playwright/blob/4c3bd11820665274071ee24663a58594a0b34b94/src/client/browserType.ts#L168-L177

kumaraditya303: It uses event listener for websocket close but in python we don't have that, for that a custom protocol class would be required which on close would do that

playwright/_impl/_browser.py
```diff
@@ -60,6 +60,11 @@ def __repr__(self) -> str:
 
     def _on_close(self) -> None:
         self._is_connected = False
+        if self._is_remote:
```
kumaraditya303, line 63: This is currently used when the transport closes unexpectedly

playwright/_impl/_browser.py
```diff
@@ -148,6 +153,12 @@ async def new_page(
     async def close(self) -> None:
         if self._is_closed_or_closing:
             return
+        if self._is_remote:
```
kumaraditya303, line 156: This is used when user calls close

playwright/_impl/_browser.py
mxschmitt: We have it, see here :) https://github.com/microsoft/playwright-python/blob/master/playwright/_impl/_browser_type.py#L212

mxschmitt: In my opinion we still should move this code away from the browser and into connect method as it was done upstream since it's specific to remote connections via connect.

mxschmitt (approved):

playwright/_impl/_browser_context.py
```diff
@@ -299,3 +372,7 @@ async def new_cdp_session(self, page: Page) -> CDPSession:
         return from_channel(
             await self._channel.send("newCDPSession", {"page": page._channel})
         )
+
+    @property
+    def tracing(self) -> Tracing:
```
dgozman, line 377: I don't think we are ready for the tracing just yet, we are still working on the API shape. Can we roll without tracing?

playwright/_impl/_browser_type.py
```diff
@@ -209,7 +212,14 @@ async def connect(
         browser._is_remote = True
         browser._is_connected_over_websocket = True
 
-        transport.once("close", browser._on_close)
+        def handle_transport_close() -> None:
```
dgozman, line 215: Why is this a part of the roll? Is it a bug fix?

tests/async/test_tracing.py
dgozman: Please drop the marker, as Max says.

tests/sync/test_browsertype_connect.py
```diff
@@ -26,10 +26,10 @@ def test_browser_type_connect_slow_mo(
     remote_server = launch_server()
     browser = browser_type.connect(remote_server.ws_endpoint, slow_mo=100)
     browser_context = browser.new_context()
-    page = browser_context.new_page()
     t1 = time.monotonic()
+    page = browser_context.new_page()
     assert page.evaluate("11 * 11") == 121
-    assert (time.monotonic() - t1) >= 0.100
+    assert (time.monotonic() - t1) >= 0.2
```
dgozman, line 32: Why this change?

playwright/_impl/_transport.py
```diff
@@ -161,6 +161,7 @@ def __init__(
 
     def request_stop(self) -> None:
         self._stopped = True
+        self.emit("close")
```
dgozman, line 164: I did not find a place that listens to this. Is it a bug fix?

playwright/_impl/_browser_type.py
mxschmitt: The disconnect handling was changed from you in the last roll, I adopted your changes.

tests/sync/test_browsertype_connect.py
mxschmitt: it was flaky before, this should "fix" the flakyness

playwright/_impl/_transport.py
mxschmitt: with the new disconnect handling, we use it to listen on the transport websocket disconnect via this

tests/sync/test_browsertype_connect.py
mxschmitt: extracted in https://github.com/microsoft/playwright-python/pull/709

playwright/_impl/_browser_context.py
mxschmitt: discussed offline, rolling with it and maybe reverting / changing it then.

kumaraditya303: @dgozman @mxschmitt fyi: Tracing was implemented in playwright-java too https://github.com/microsoft/playwright-java/pull/444

playwright/_impl/_transport.py
kumaraditya303: Required for when the websocket connection is closed without client closing

tests/sync/test_browsertype_connect.py
kumaraditya303: Fine to extract it away but should allow these kind of small fixes for flaky tests anyway

PR #685, opened 2021-05-12
fix: exception when page got closed via browser in persistent context

### LINKED ISSUE #681: Error accessing page and error closing browser if extension is loaded: "AttributeError: 'NoneType' object has no attribute '_pages'"
Hi!

Both for sync api and async api I have the same error, happens about 80% of the time. `browser.close()` will produce `AttributeError: 'NoneType' object has no attribute '_pages'` where stack trace comes outside of my script and I cannot catch it with `try: except AttributeError` within my code.

On top of this, `playwright-python` opens a new blank page + metamask page (not-expected); JS Playwright only opens + metamask page.

Extensions I used are 1) Metamask; 

I cannot access the page, PlayWright shows 3 pages visually but lists only 2, only the blank ones; MetaMask page is both non-accessible and also 80% of the time raises outside-script stack trace from browser closed "AttributeError: 'NoneType' object has no attribute '_pages'"

```

# these need to be replaced to your ones; Metamask unpacked can be downloaded here
# https://github.com/MetaMask/metamask-extension/releases/tag/v9.5.0
EXT_1_DIR = "C:/Users/<individual_path>/metamask_unpck"
USER_DIR = "C:/Users/<individual_path>/User Data/Profile 2"

ARGS = [
    "--disable-extensions-except={}".format(EXT_1_DIR),
    "--load-extension={}".format(EXT_1_DIR)
]

with sync_playwright() as p:
    browser = p.chromium.launch_persistent_context(USER_DIR, headless=False, args=ARGS)
    page = browser.new_page()
    _ = input("Wait till Metamask loads? ")
    print("len_pages()", len(browser.pages))
    for pg in browser.pages:
    	pg.close()

    _ = input("Close browser? ")
    browser.close()
```

    
This issue is `playwright-python` specific and there's no issue on `playwright-for-javascript` pls help. The issue is the same for both `sync_api` and `async_api` but I used `sync_api` in this example as it's a tad simpler and a tad less code.

Here's illustrated powershell output screenshot;
https://imgur.com/a/iMDSPcS

OS: Windows 10, playwright-python 1.11.0; Python 3.9

PROBLEM TL;DR: prevents me from using MetaMask extension and raises error on browser.close() on top - an issue I don't have in PlayWright for JS. Prob. same issue on other than MetaMask extensions as well, but I didn't have time to bulk up the sample.

mxschmitt:
Fixes #681

not sure about a test for it.

A Set in JavaScript simply does not throw when there is no element which gets removed.

it throws here: https://github.com/microsoft/playwright-python/blob/9d796623a104e082f57878d29d9aea20009b88c1/playwright/_impl/_page.py#L274

### DIFF
--- a/playwright/_impl/_page.py
+++ b/playwright/_impl/_page.py
@@ -111,7 +111,7 @@ class Page(ChannelOwner):
         self, parent: ChannelOwner, type: str, guid: str, initializer: Dict
     ) -> None:
         super().__init__(parent, type, guid, initializer)
-        self._browser_context: BrowserContext = None  # type: ignore
+        self._browser_context: BrowserContext = parent
         self.accessibility = Accessibility(self._channel)
         self.keyboard = Keyboard(self._channel)
         self.mouse = Mouse(self._channel)
@@ -126,7 +126,9 @@ class Page(ChannelOwner):
         self._bindings: Dict[str, Any] = {}
         self._routes: List[RouteHandlerEntry] = []
         self._owned_context: Optional["BrowserContext"] = None
-        self._timeout_settings: TimeoutSettings = TimeoutSettings(None)
+        self._timeout_settings: TimeoutSettings = TimeoutSettings(
+            self._browser_context._timeout_settings
+        )
         self._video: Optional[Video] = None
         self._opener = cast("Page", from_nullable_channel(initializer.get("opener")))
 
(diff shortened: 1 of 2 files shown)


playwright/_impl/_page.py
```diff
@@ -271,7 +269,8 @@ def _on_worker(self, worker: "Worker") -> None:
 
     def _on_close(self) -> None:
         self._is_closed = True
-        self._browser_context._pages.remove(self)
+        if self in self._browser_context._pages:
```
pavelfeldman, line 272: Is it ever false?

pavelfeldman (approved):

pavelfeldman (approved):

PR #691, opened 2021-05-13
feat: implemented context tracing

kumaraditya303:
Blocked https://github.com/microsoft/playwright-python/pull/682


mxschmitt:
By definition, this is a roll, so it would be good if you could do it also in #682.

kumaraditya303:
Ok, I am thinking of fixing that websocket error in that pr and then rebase this on top of it. But till now have to fix for that websocket error, will just skip it for now I think.

Issue #689, opened 2021-05-13
Adding extension causes extra unwanted blank page to open and extension's opened page becomes a 'ghost' page and is completely unreachable!


ivarsj10s:
On JavaScript adding an extension which opens 'welcome' page such as MetaMask => extension opens 1 new page and you can close this page and the default page.

On Playwright Python adding an extension which opens 'welcome' page such as MetaMask => extension opens 1 extra blank page (2 blank pages / total) and a new 'extension' _**'ghost'**_  page which is unnaccessible from Playwright Python and you cannot close it.

I briefly mentioned this issue on https://github.com/microsoft/playwright-python/issues/681 where the developer who got assigned fixed my main issue in the title but skimmed over this issue. The developer's given fix only fixed `browser.close()` raising errors => into => not raising errors, but not this issue. This issue is still 'at large' (_i.e. especially of a criminal or dangerous animal) at liberty; escaped or not yet captured._)

PROBLEM: this bug causes an unaccessible 'extension' page to just hang there and I cannot close it using code. It's not inside `browser.pages`. 

Illustrative image sets with descriptive captions:
1) Playwright JS
https://imgur.com/a/X13eydu

2) Playwright Python
https://imgur.com/a/mkINpWl

EDIT: adding source as by confusing from @pavelfeldman 's reply
```
import playwright
from playwright.sync_api import sync_playwright

EXT_1_DIR = "C:/Users/PC/Desktop/PW_PY/metamask_unpck"
USER_DIR = "C:/Users/PC/AppData/Local/Google/Chrome/User Data/Profile 2"

ARGS = [
    "--disable-extensions-except={}".format(EXT_1_DIR),
    "--load-extension={}".format(EXT_1_DIR)
]


with sync_playwright() as p:
    browser = p.chromium.launch_persistent_context(USER_DIR, headless=False, args=ARGS)
    page = browser.new_page()
    _ = input("Wait till Metamask loads? ")
    print("len_pages()", len(browser.pages))
    for pg in browser.pages:
        pg.close()

    _ = input("Close browser? ")
    browser.close()
```

EDIT: update to @pavelfeldman 's input.

new source
```
import playwright
from playwright.sync_api import sync_playwright

EXT_1_DIR = "C:/Users/PC/Desktop/PW_PY/metamask_unpck"
USER_DIR = "C:/Users/PC/AppData/Local/Google/Chrome/User Data/Profile 2"

ARGS = [
[...]

PR #698, opened 2021-05-17
docs: add note about time.sleep

mxschmitt:
https://github.com/microsoft/playwright-python/issues/689#issuecomment-841632670

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Headless execution is supported for all browsers on all platforms.
   - [Mobile and geolocation](#mobile-and-geolocation)
   - [Evaluate JS in browser](#evaluate-js-in-browser)
   - [Intercept network requests](#intercept-network-requests)
+- [Known issues](#known-issues)
 - [Documentation](#documentation)
 
 ## Usage - pip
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -298,6 +298,12 @@ asyncio.run(main())
 
 </details>
 
+## Known issues
```
kumaraditya303, line 301: `asyncio.sleep` should work here in async context

README.md
```diff
@@ -298,6 +298,12 @@ asyncio.run(main())
...
+
+### `time.sleep()` leads to outdated state
+
+We internally rely on asynchronous operations. When using `time.sleep(5)` they can't process internally and by that e.g. the internal storage for pages can't get updated. To fix this behavior you should use `page.wait_for_timeout(5000)` instead if you are using the synchronous variant. For the asynchronous variant you can also use `await asyncio.sleep(5)`.
```
kumaraditya303, line 305:
```suggestion
We internally rely on asynchronous operations. When using `time.sleep(5)` they can't be processed internally and by that e.g. the internal storage for pages can't get updated. To fix this behaviour you should use `page.wait_for_timeout(5000)` instead if you are using the synchronous api. For the asynchronous api you can use `await asyncio.sleep(5)`.
```

kumaraditya303 (approved):

README.md
```diff
@@ -298,6 +298,12 @@ asyncio.run(main())
 
 </details>
 
+## Known issues
+
+### `time.sleep()` leads to outdated state
+
+We internally rely on asynchronous operations. When using `time.sleep(5)` they can't be processed internally and by that e.g. the internal storage for pages can't get updated. To fix this behaviour you should use `page.wait_for_timeout(5000)` instead if you are using the synchronous api. For the asynchronous api you can use `await asyncio.sleep(5)`.
```
pavelfeldman, line 305: I would start with the solution:
===
Always use page.wait_for_timeout(5000) instead of thread.sleep(5).

It is better to not wait for timeout at all, but sometimes it is useful for debugging. In these cases, use our wait method instead of the system one.
===

pavelfeldman (approved):

PR #703, opened 2021-05-20
chore: added types for devices

kumaraditya303:
Screenshot (70)


playwright/_impl/_api_structures.py
```diff
@@ -119,3 +120,23 @@ class FilePayload(TypedDict):
     name: str
     mimeType: str
     buffer: bytes
+
+
+__all__ = [
```
mxschmitt, line 125: Why do we need this here?

kumaraditya303: Since here we are re-exporting the devices and if there isn't __all__ then flake8 would complain

mxschmitt (approved):
I would probably rename it to `_devices_generated.py` otherwise it looks good since it's automated generated.

scripts/expected_api_mismatch.txt
```diff
@@ -20,3 +20,5 @@ Parameter type mismatch in BrowserContext.cookies(return=): documented as List[{
 Method not implemented: Error.name
 Method not implemented: Error.stack
 Method not implemented: Error.message
+# Temporary
```
dgozman, line 23: This is not temporary. Let's change to `# Devices are properly typed`.

playwright/_impl/_api_structures.py
```diff
@@ -26,6 +26,7 @@
 
 # Explicitly mark optional params as such for the documentation
 # If there is at least one optional param, set total=False for better mypy handling.
+from playwright._impl._devices_generated import DeviceDescriptor, Devices
```
dgozman, line 29: Why do we need to import this here? I'd instead define types here, and import them in `_devices_generated.py`. This way we don't need the `__all__` hack. WDYT?

kumaraditya303: If you want to directly import it then you would need some custom handling in docs https://github.com/microsoft/playwright-python/blob/cfc1030a69d1e934cac579687a680eac53d4b9ee/scripts/generate_api.py#L129, also __all__ is not a hacky solution, it is there to not make falke8 complain about not exported imports

scripts/expected_api_mismatch.txt
kumaraditya303: Waiting for review, if wanted to add types to docs upstream, but would change comment instead

playwright/_impl/_api_structures.py
mxschmitt: If it's only flake8 there which is complaining then I would keep it as it was, since we don't have the flake8 linter in place and it was working before.

kumaraditya303: Are you sure that we don't have falke8 https://github.com/microsoft/playwright-python/blob/835cdce3f0497192748a806bfe537051f1dd18d2/.pre-commit-config.yaml#L20 ?

kumaraditya303: Ignored now

pavelfeldman:
I am not really sure this helps much. My understanding is that the device name is only mentioned in some test config, do you have a bug report or a user requesting this?

kumaraditya303:
https://github.com/microsoft/playwright/blob/c60974d92273ef0ee9f4af7a465a61f12a312784/types/types.d.ts#L11559 The devices name are generated as types in typescript dts too, why are you opposing adding it to python here, this will enhance the type stubs, also it is used directly not in some test config.

pavelfeldman:
Some of these descriptors are old, some others are stale. I am not comfortable making those names a part of the API that we need to maintain. I would encourage you to file an issue explaining the rationale of changes like this upfront and giving us a chance to provide the feedback. That way you don't waste time on something that is potentially not included into production! Keep up the great work!

PR #707, opened 2021-05-21
chore: simplified transport stopping

kumaraditya303:
Uses [`asyncio.Event`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event) synchronization primitive to simplify the code.

### DIFF
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -19,12 +19,13 @@ import os
 import sys
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import Callable, Dict, Optional
 
 import websockets
 from pyee import AsyncIOEventEmitter
 
 from playwright._impl._api_types import Error
+from playwright._impl._helper import ParsedMessagePayload
 
 
 # Sourced from: https://github.com/pytest-dev/pytest/blob/da01ee0a4bb0af780167ecd228ab3ad249511302/src/_pytest/faulthandler.py#L69-L77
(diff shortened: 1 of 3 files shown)


playwright/sync_api/_context_manager.py
```diff
@@ -37,6 +37,7 @@ def __enter__(self) -> SyncPlaywright:
             loop = asyncio.get_running_loop()
         except RuntimeError:
             loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
```
kumaraditya303, line 40: It is required because the [loop parameter in asyncio is deprecated](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event) and this set up the loop correctly when there is no event loop in the current thread correctly

mxschmitt:
This looks in my mind more confusing now for non-python-asyncio experts.

playwright/_impl/_transport.py
```diff
@@ -84,22 +84,21 @@ def __init__(
...
         self._driver_executable = driver_executable
 
     def request_stop(self) -> None:
-        self._stopped = True
+        self._stopped.set()
         self._output.close()
 
     async def wait_until_stopped(self) -> None:
-        await self._stopped_future
+        await self._stopped.wait()
```
kumaraditya303, line 95: Don't you think it is confusing about what does `_stopped` do and `_stopped_future` previously ?

kumaraditya303: With this change there is only `_stopped` which is a lot simple to reason about

playwright/_impl/_transport.py
```diff
@@ -154,20 +153,20 @@ def __init__(
...
         Transport.__init__(self, loop)
 
-        self._stopped = False
         self.ws_endpoint = ws_endpoint
         self.headers = headers
         self.slow_mo = slow_mo
 
     def request_stop(self) -> None:
-        self._stopped = True
+        self._stopped.set()
```
pavelfeldman, line 161: We should not notify those that wait for stop here, we do that when we actually stopped. Is there a bug for this? What are we fixting?

kumaraditya303: We do wait for the actual connection to close https://github.com/microsoft/playwright-python/blob/ef07845d4433467798dcabe29daefa737f8518f9/playwright/_impl/_transport.py#L170

kumaraditya303: Can you show a failing test with the current change which was passing previously to show your point better ?, fyi: this change does not break anything

kumaraditya303: @mxschmitt Your opinion here would be appreciated!

PR #711, opened 2021-05-23
chore: Kill windows subprocess using kill

kumaraditya303:
There is no special handling required for windows so the code was redundant and hence removed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -238,14 +238,7 @@ class RemoteServer:
         # Send the signal to all the process groups
         if self.process.poll() is not None:
             return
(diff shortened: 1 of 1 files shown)


mxschmitt:
Could you add a description why this change was done or link a bug report?

kumaraditya303:
@mxschmitt Can you explain why we need extra handling on windows to kill subprocess, as far as I know the kill function was not there in python 2.5 but we only use python 3.7+ and was added in python 2.6 and there is no need to kill using taskkill.

mxschmitt (approved):
tested it and seems working.

PR #715, opened 2021-05-24
feat(roll): Roll Playwright driver to 1.12.0-next-1621812078000

kumaraditya303:


playwright/_impl/_helper.py
```diff
@@ -50,6 +50,7 @@
 RouteHandler = Union[Callable[["Route"], Any], Callable[["Route", "Request"], Any]]
 
 ColorScheme = Literal["dark", "light", "no-preference"]
+ReducedMotion = Literal["no-preference", "reduce"]
```
mxschmitt, line 53:
```suggestion
ReducedMotion = Literal["reduce", "no-preference"]
```

mxschmitt (approved):

playwright/_impl/_helper.py
kumaraditya303: It causes linting error so let it be like the earlier one https://github.com/microsoft/playwright-python/pull/715/checks?check_run_id=2672816782

mxschmitt (approved):

PR #717, opened 2021-05-24
tests: cleanup tests

mxschmitt:
Applied:
- https://github.com/microsoft/playwright/pull/3742
- https://github.com/microsoft/playwright/pull/6444

First iteration for reducing flaky tests, will follow up with more.

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -338,7 +338,3 @@ def serialize_call_stack(stack_trace: traceback.StackSummary) -> List[Dict]:
         )
     stack.reverse()
     return stack
-
-
-def capture_call_stack() -> List[Dict]:
-    return serialize_call_stack(traceback.extract_stack())
--- a/tests/async/test_browsertype_connect_cdp.py
+++ b/tests/async/test_browsertype_connect_cdp.py
@@ -18,7 +18,7 @@ import pytest
 import requests
 
 from playwright.async_api import BrowserType
-from tests.server import Server, find_free_port, wait_for_port
+from tests.server import Server, find_free_port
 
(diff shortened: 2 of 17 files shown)


playwright/_impl/_connection.py
```diff
@@ -338,7 +338,3 @@ def serialize_call_stack(stack_trace: traceback.StackSummary) -> List[Dict]:
         )
     stack.reverse()
     return stack
-
-
-def capture_call_stack() -> List[Dict]:
```
mxschmitt, line 343: this was unused

tests/async/test_browsertype_connect_cdp.py
```diff
@@ -30,7 +30,6 @@ async def test_connect_to_an_existing_cdp_session(
     browser_server = await browser_type.launch(
         **launch_arguments, args=[f"--remote-debugging-port={port}"]
     )
-    wait_for_port(port)
```
mxschmitt, line 33: We upstream wait now if the --remote-debugging-port arg is passed until the websocket server is ready.

tests/async/test_tracing.py
```diff
@@ -28,6 +28,7 @@ async def test_browser_context_output_trace(
     page = await context.new_page()
     await page.goto(server.PREFIX + "/grid.html")
     await context.tracing.stop()
+    await page.wait_for_timeout(100)
```
dgozman, line 31: Do we have this upstream? Perhaps make it 1000?

dgozman (approved):

tests/async/test_tracing.py
mxschmitt: I tried to repro it upstream, but wasn't successful with it, will investigate further in a follow-up.

PR #742, opened 2021-06-06
chore: use `asyncio.get_running_loop`

kumaraditya303:


setup.py
```diff
@@ -28,7 +28,7 @@
     InWheel = None
 from wheel.bdist_wheel import bdist_wheel as BDistWheelCommand
 
-driver_version = "1.12.0-next-1622782349000"
+driver_version = "1.12.0-next-1622928816000"
```
mxschmitt, line 31: please do not include a roll inside this PR.

playwright/async_api/_context_manager.py
```diff
@@ -27,15 +27,15 @@ def __init__(self) -> None:
...
         self._connection = Connection(
             None,
             create_remote_object,
-            PipeTransport(asyncio.get_event_loop(), compute_driver_executable()),
+            PipeTransport(loop, compute_driver_executable()),
         )
-        loop = asyncio.get_running_loop()
         self._connection._loop = loop
         loop.create_task(self._connection.run())
-        playwright_future = asyncio.create_task(
```
mxschmitt, line 38: It does by default use the current running loop, so I wouldn't change `asyncio.create_task` to `loop.create_task` or do I miss something?

kumaraditya303: It does but this way it is consistent with `loop.create_task(self._connection.run())` or WDYT?

mxschmitt: convinced. SGTM. :+1:

mxschmitt (approved):

pavelfeldman:
Looks like it does not fix anything, but could break something!

kumaraditya303:
@pavelfeldman You should have look at the linked issue of the PR

PR #746, opened 2021-06-07
chore: added context managers

### LINKED ISSUE #730: [FEATURE] Context managers for page and context
Currently playwright does not provides context manager for page and contexts and if they are not closed, they will cause resource leak or memory leak, however python has context managers for managing resources which would auto close the resource once they are used. Playwright should have context managers for page and contexts.

TBD API:
```python

with sync_playwright() as p:
    for browser_type in [p.chromium, p.firefox, p.webkit]:
        with browser_type.launch() as browser:
            with browser.new_page() as page:
                page.screenshot(path=f"example-{browser_type.name}.png")

```

kumaraditya303:
Fixes #730
Added context manager for the following:
- Browser
- BrowserContext
- Page

Test the sync code with 
```python

with sync_playwright() as p:
    for browser_type in [p.chromium, p.firefox, p.webkit]:
        with browser_type.launch() as browser:
            with browser.new_context() as context:
                with context.new_page() as page:
                    page.screenshot(path=f"example-{browser_type.name}.png")
```

Test Async code with 

```python
import asyncio

async def main():
        for browser_type in [p.chromium, p.firefox, p.webkit]:
            browser = await browser_type.launch()
            async with browser:
                context = await browser.new_context()
                async with context:
                    page = await context.new_page()
                    async with page:
                        await page.goto("http://whatsmyuseragent.org/")
                        await page.screenshot(path=f"example-{browser_type.name}.png")


asyncio.run(main())

```

### DIFF
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import base64
 import json
 from pathlib import Path
--- a/playwright/_impl/_browser.py
+++ b/playwright/_impl/_browser.py
@@ -11,12 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import base64
 import json
 from pathlib import Path
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Dict, List, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Type, Union
 
 from playwright._impl._api_structures import (
     Geolocation,
@@ -188,6 +187,17 @@ class Browser(ChannelOwner):
     async def stop_tracing(self) -> bytes:
         encoded_binary = await self._channel.send("stopTracing")
         return base64.b64decode(encoded_binary)
+
+    async def __aenter__(self) -> "Browser":
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Type[BaseException],
+        exc_val: BaseException,
+        traceback: Any,
+    ) -> None:
+        await self.close()
 
 
 def normalize_context_params(is_sync: bool, params: Dict) -> None:
--- a/playwright/_impl/_browser_context.py
+++ b/playwright/_impl/_browser_context.py
@@ -15,9 +15,21 @@
 import asyncio
 import inspect
 import json
+import typing
 from pathlib import Path
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Set,
+    Type,
+    Union,
+    cast,
+)
 
 from playwright._impl._api_structures import Cookie, Geolocation, StorageState
 from playwright._impl._api_types import Error
@@ -374,3 +386,14 @@ class BrowserContext(ChannelOwner):
     @property
     def tracing(self) -> Tracing:
         return self._tracing
+
+    async def __aenter__(self) -> "BrowserContext":
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Type[BaseException],
+        exc_val: BaseException,
+        traceback: typing.Any,
+    ) -> None:
+        await self.close()
--- a/playwright/_impl/_page.py
+++ b/playwright/_impl/_page.py
@@ -18,7 +18,7 @@ import inspect
 import sys
 from pathlib import Path
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union, cast
 
 from playwright._impl._accessibility import Accessibility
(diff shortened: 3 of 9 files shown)


mxschmitt (approved):
LGTM, maybe we can even change our fixtures (browser, context, page) for sync/async instead of:
```py
    page = await context.new_page()
    yield page
    await page.close()
```

to 

```py
    with await context.new_page() as page:
        yield page
```

which will dogfood the change.

playwright/_impl/_browser.py
```diff
@@ -11,12 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import base64
```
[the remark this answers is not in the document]
kumaraditya303: Okay, Black sometimes add a extra line though :/

kumaraditya303:
> LGTM, maybe we can even change our fixtures (browser, context, page) for sync/async instead of:
> 
> ```python
>     page = await context.new_page()
>     yield page
>     await page.close()
> ```
> 
> to
> 
> ```python
>     with await context.new_page() as page:
>         yield page
> ```
> 
> which will dogfood the change.

Will follow up in a separate PR with some more changes from other PR too

tests/sync/test_context_manager.py
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+
+from playwright.sync_api import BrowserType
+
+
+def test_context_managers(browser_type: BrowserType, launch_arguments):
+    with browser_type.launch(**launch_arguments) as browser:
+        with browser.new_context() as context:
+            with context.new_page() as page:
+                page.goto("http://whatsmyuseragent.org/")
```
mxschmitt, line 22: We don't want to rely on external websites for the tests. A simple page.evaluate("1 + 1") would be enough.

kumaraditya303: Fixed

mxschmitt (approved):
So `browser.new_page` does also work with the context manager? Having documentation would be also great to have, so users can use it for small scripts.

kumaraditya303:
@mxschmitt will follow docs in a separate PR, new_page also works with context manager

playwright/_impl/_browser.py
```diff
@@ -189,6 +189,17 @@ async def stop_tracing(self) -> bytes:
         encoded_binary = await self._channel.send("stopTracing")
         return base64.b64decode(encoded_binary)
 
+    async def __aenter__(self) -> "Browser":
```
pavelfeldman, line 192: We never expose _impl to the users, so I don't think we should implement contexts for them.

playwright/_impl/_browser_context.py
```diff
@@ -374,3 +386,14 @@ async def new_cdp_session(self, page: Page) -> CDPSession:
     @property
     def tracing(self) -> Tracing:
         return self._tracing
+
+    async def __aenter__(self) -> "BrowserContext":
```
pavelfeldman, line 390: Same here

playwright/_impl/_page.py
```diff
@@ -869,6 +869,17 @@ def expect_worker(
     ) -> EventContextManagerImpl["Worker"]:
         return self.expect_event("worker", predicate, timeout)
 
+    async def __aenter__(self) -> "Page":
```
pavelfeldman, line 872: Same here

scripts/generate_async_api.py
```diff
@@ -58,7 +58,21 @@ def generate(t: Any) -> None:
         print(f"{prefix}{suffix}")
     for [name, value] in t.__dict__.items():
         if name.startswith("_"):
-            continue
+            if name in ["__aenter__", "__aexit__"]:
```
pavelfeldman, line 61: Instead of growing the generator, introduce `SyncBaseContextManager` and `AsyncBaseContextManager` that would implement enter/exit and only inherit Page, BrowserContext and Browser from them.

tests/async/test_context_manager.py
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from playwright.async_api import BrowserType
+
+
+async def test_context_managers(browser_type: BrowserType, launch_arguments):
+    browser = await browser_type.launch(**launch_arguments)
+    async with browser:
```
pavelfeldman, line 20: Is `async with variable` idiomatic? I thought that the whole point of context manager was to never leak the value outside of the context.

pavelfeldman (changes requested):

playwright/_impl/_browser.py
kumaraditya303: Indeed we never expose _impl to users, users does not get any object from _impl, they always it is wrapped

playwright/_impl/_browser_context.py
kumaraditya303: Same answer

playwright/_impl/_page.py
kumaraditya303: Same answer

tests/async/test_context_manager.py
kumaraditya303: Nope, the use of context manager is to handle the resource and do some exit work after done, not to leak the value etc

scripts/generate_async_api.py
pavelfeldman: Were you going to address this or should I close the PR? Either way is fine!

kumaraditya303: I didn't understood how you wanted to do it, so closing

PR #748, opened 2021-06-09
feat: support dynamic browser name

### LINKED ISSUE #731: [FEATURE] Playwright does not support dynamic browser name
Playwright in typescript supports accessing the browser name with its string name like :
```ts
const playwright = require('playwright');

(async () => {
  for (const browserType of ['chromium', 'firefox', 'webkit']) {
    const browser = await playwright[browserType].launch();
    const context = await browser.newContext();
    const page = await context.newPage();
    await page.goto('http://whatsmyuseragent.org/');
    await page.screenshot({ path: `example-${browserType}.png` });
    await browser.close();
  }
})();
```
Whereas in python, it will errors as the playwright object is not subscriptable 
```console
TypeError: 'Playwright' object is not subscriptable 
```

Playwright should make the playwright subscriptable to be able to dynamically launch browser from string

TBD API:
```py

with sync_playwright() as p:
    for browser_type in ["chromium", "firefox", "webkit"]:
        browser = p[browser_type].launch()
        page = browser.new_page()
        page.screenshot(path=f"example-{browser_type.name}.png")
        browser.close()

```

kumaraditya303:
Fixes #731

### DIFF
--- a/playwright/async_api/_generated.py
+++ b/playwright/async_api/_generated.py
@@ -9930,6 +9930,15 @@ class Playwright(AsyncBase):
 
         return mapping.from_maybe_impl(self._impl_obj.stop())
 
+    def __getitem__(self, value: str) -> "BrowserType":
+        if value == "chromium":
+            return self.chromium
+        elif value == "firefox":
+            return self.firefox
+        elif value == "webkit":
+            return self.webkit
+        raise ValueError("Invalid browser " + value)
+
 
 mapping.register(PlaywrightImpl, Playwright)
 
--- a/playwright/sync_api/_generated.py
+++ b/playwright/sync_api/_generated.py
@@ -9872,6 +9872,15 @@ class Playwright(SyncBase):
 
         return mapping.from_maybe_impl(self._impl_obj.stop())
 
+    def __getitem__(self, value: str) -> "BrowserType":
+        if value == "chromium":
+            return self.chromium
+        elif value == "firefox":
+            return self.firefox
+        elif value == "webkit":
+            return self.webkit
+        raise ValueError("Invalid browser " + value)
+
 
 mapping.register(PlaywrightImpl, Playwright)
 
--- a/scripts/generate_async_api.py
+++ b/scripts/generate_async_api.py
@@ -115,7 +115,19 @@ def generate(t: Any) -> None:
                     f"""
         return {prefix}{arguments(value, len(prefix))}{suffix}"""
                 )
-
+    if class_name == "Playwright":
+        print(
+            """
+    def __getitem__(self, value: str) -> "BrowserType":
+        if value == "chromium":
+            return self.chromium
+        elif value == "firefox":
+            return self.firefox
+        elif value == "webkit":
+            return self.webkit
+        raise ValueError("Invalid browser "+value)
+            """
+        )
     print("")
     print(f"mapping.register({class_name}Impl, {class_name})")
 
--- a/scripts/generate_sync_api.py
+++ b/scripts/generate_sync_api.py
@@ -114,7 +114,19 @@ def generate(t: Any) -> None:
                     f"""
         return {prefix}{arguments(value, len(prefix))}{suffix}"""
                 )
-
+    if class_name == "Playwright":
+        print(
+            """
+    def __getitem__(self, value: str) -> "BrowserType":
+        if value == "chromium":
+            return self.chromium
+        elif value == "firefox":
+            return self.firefox
+        elif value == "webkit":
+            return self.webkit
+        raise ValueError("Invalid browser "+value)
+            """
+        )
     print("")
     print(f"mapping.register({class_name}Impl, {class_name})")
 


playwright/_impl/_playwright.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
```
mxschmitt, line 15:
```suggestion
```

playwright/_impl/_playwright.py
```diff
@@ -40,6 +41,15 @@ def __init__(
...
         }
 
+    def __getitem__(self, value: str) -> BrowserType:
+        if value == "chromium":
+            return self.chromium
+        elif value == "firefox":
+            return self.firefox
+        elif value == "webkit":
+            return self.webkit
+        raise ValueError(f"Invalid browser {value!r}")
```
mxschmitt, line 51: out of curiosity, what does the !r do again?

playwright/_impl/_playwright.py
```diff
@@ -40,6 +41,15 @@ def __init__(
             for device in initializer["deviceDescriptors"]
         }
 
+    def __getitem__(self, value: str) -> BrowserType:
+        if value == "chromium":
+            return self.chromium
+        elif value == "firefox":
+            return self.firefox
+        elif value == "webkit":
```
mxschmitt, line 49:
```suggestion
        if value == "firefox":
            return self.firefox
        if value == "webkit":
```

tests/assets/client.py
```diff
@@ -12,22 +12,22 @@
...
-    for browser_type in [playwright.chromium, playwright.firefox, playwright.webkit]:
-        browser = browser_type.launch()
-        page = browser.new_page()
-        page.goto("https://example.com")
-        here = Path(__file__).parent.resolve()
-        page.screenshot(path=here / f"{browser_type.name}.png")
-        page.close()
-        browser.close()
+def main() -> None:
+    with sync_playwright() as p:
```
mxschmitt, line 22: having additionally a normal test-case in the browser tests for async and sync would be good to have.

mxschmitt (approved):
LGTM, just nits

playwright/_impl/_playwright.py
kumaraditya303: Prints the browser name in quotes which looks nice in traceback

tests/assets/client.py
kumaraditya303: Okay

playwright/_impl/_playwright.py
kumaraditya303: else if should be fine I guess ?

mxschmitt: yeah its more a personal preference.

tests/async/test_dynamic_browser_name.py
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from playwright.sync_api import sync_playwright
+
+
+def test_dynamic_browser_name(browser_name, launch_arguments):
+    with sync_playwright() as p:
+        browser = p[browser_name].launch(**launch_arguments)
+        page = browser.new_page()
+        page.goto("https://github.com")
```
mxschmitt, line 21: We don't want to rely on external websites for the test. A simple assert page.evaluate("1 + 1") == 2 would be enough.

kumaraditya303: Fixed

playwright/_impl/_playwright.py
```diff
@@ -40,6 +40,15 @@ def __init__(
             for device in initializer["deviceDescriptors"]
         }
 
+    def __getitem__(self, value: str) -> BrowserType:
```
pavelfeldman, line 43: I would inline this exact snippet when generating Playwright (both sync and async), as a single hard-coded blob. That way we don't grow complexity in the generator.

pavelfeldman:
Where did the trailing coma come from in all the generated files?

kumaraditya303:
Black formatted it

pavelfeldman:
> Black formatted it

I don't see where you are changing the Black settings.

kumaraditya303:
> > Black formatted it
> 
> I don't see where you are changing the Black settings.

Even I don't know, https://test-black.readthedocs.io/en/style-guide/style_guide/trailing_commas.html explains it

kumaraditya303:
@pavelfeldman As you can see here in the diff black is adding the comma https://github.com/microsoft/playwright-python/pull/748/checks?check_run_id=2914309332 and with the diff the CI passes hence not wating more time Can you merge this ?

pavelfeldman:
This looks like a Black bug: if you replace `f"Invalid browser {value!r}"` with `"Invalid browser " + value` it goes back to normal.

pavelfeldman (approved):

PR #764, opened 2021-06-19
chore: bump dev deps

kumaraditya303:


pavelfeldman:
Looks like I already merge something that included this.

PR #781, opened 2021-06-28
test: unflake test_should_emit_frame_events test

mxschmitt:
Its not 1:1 the same as upstream:

https://github.com/microsoft/playwright/blob/409aeaa4438585f44484bfc9d43e7d51f9b5c22a/tests/web-socket.spec.ts#L52-L73

### DIFF
--- a/tests/async/test_websocket.py
+++ b/tests/async/test_websocket.py
@@ -55,30 +55,32 @@ async def test_should_emit_close_events(page, ws_server):
 
 
 async def test_should_emit_frame_events(page, ws_server):
-    sent = []
-    received = []
+    log = []
+    socke_close_future = asyncio.Future()
 
     def on_web_socket(ws):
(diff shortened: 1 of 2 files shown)


tests/async/test_websocket.py
```diff
@@ -55,30 +55,32 @@ async def test_should_emit_close_events(page, ws_server):
...
-    if not ws.is_closed():
-        await ws.wait_for_event("close")
-
-    assert sent == ["echo-text"]
-    assert received == ["incoming", "text"]
+    await socke_close_future
+    assert log[0] == "open"
+    assert log[3] == "close"
+    log.sort()
+    assert ":".join(log) == "close:open:received<incoming>:sent<outgoing>"
```
kumaraditya303, line 83: nit: You should just assert the full log which would also assert its order

pavelfeldman (approved):

PR #790, opened 2021-07-05
fix: improve typing for SyncBase and AsyncBase

kumaraditya303:


playwright/_impl/_async_base.py
```diff
@@ -15,19 +15,19 @@
...
 
 
 T = TypeVar("T")
-Self = TypeVar("Self", bound="AsyncBase")
+Self = TypeVar("Self", bound="AsyncContextManager")
 
 
 class AsyncEventInfo(Generic[T]):
-    def __init__(self, future: asyncio.Future) -> None:
+    def __init__(self, future: "asyncio.Future[T]") -> None:
```
pavelfeldman, line 30: I don't think Future is generic in 3.7, 3.8.

pavelfeldman: Ah, I see you are using quotes to work around

kumaraditya303: Any other changes required here ?

PR #800, opened 2021-07-10
chore: show code coverage in CI

kumaraditya303:


kumaraditya303:
@mxschmitt Ready for merge!

mxschmitt:
What value does this give us? Having the % per file is not very useful. I think running it locally should be enough there you can also inspect which lines are not covered.

Feel free to create tests for uncovered lines (I know there are a bunch of them).

kumaraditya303:
This would give a timeline of test coverage, also it consume time in running several tests locally, it also gives us a rough idea where the code coverage is less in which files.
Any reason to not do this ?

pavelfeldman:
Line coverage does not give us anything - we can still miss passing one argument over the channel and while status would be happy, tool would be broken. Would not want this false positive on the dashboard.

Issue #513, opened 2021-02-17
[Feature]: Make it compatible with Pyinstaller


IHateTomLrge:
Hi. 
I'm bundling my app using PyInstaller.
But I'm facing an issue.
When I use the arg: --console. it works well, the browser opens itself when I call it....
But when I use the arg: --windowed. I'm not getting error (no timeout, that's why) but so we are sure that it finds the browser, but the "headless" mode to False isn't working. to explains briefly, I open the browser to solve captchas. so I need it to be "headless=False" but even If I use this arg, it doesn't shows up, without getting any error (if I set timeout to 3000 It will raise error of selector not found)

nubonics:
Were you able to get the an executable with playwright? if so, would you mind sharing how you did so?

Try writing the page content to a file right before you are calling this selector [make sure you account for page navigation when doing so if needed], and viewing it in your browser. Chances are, it is not going to the page you expect, and that is why it cannot find the selector. I very well could be wrong though.

tomlarge:
> Were you able to get the an executable with playwright? if so, would you mind sharing how you did so?
> 
> Try writing the page content to a file right before you are calling this selector [make sure you account for page navigation when doing so if needed], and viewing it in your browser. Chances are, it is not going to the page you expect, and that is why it cannot find the selector. I very well could be wrong though.

No, you didn't understand I think. My prob doesn't comes from the page, like, it works fine, it select everything I told the script to fill/click on....
The issue comes from the headless mode, I never shows up, even I set the Headless to False, it will create an headless browser. And I need it to works on non headless browser.

nubonics:
So what your saying is you cant get out of headless mode? 
So your headless flag is not working? As playwright uses headless by default, so it seems you are not passing the correct flag to disable headless mode.
 If this is not the case, what are your reproducible steps? However, if you are passing the paramater correctly, I suggest starting over, create a new virtual environment, copy the code you had before, install playwright and pyinstaller, and try again, however, I think this has more to do with not passing the flags correctly, as this does not happen on my machine(s).

tomlarge:
> So what your saying is you cant get out of headless mode?
> So your headless flag is not working? As playwright uses headless by default, so it seems you are not passing the correct flag to disable headless mode.
[...]

PR #812, opened 2021-07-19
feat: added hooks for pyinstaller

kumaraditya303:
For review instructions:

- Create a file `main.py`
- add this to the file 
```python
import os

with sync_playwright() as p:
    browser = p.chromium.launch()
    page = browser.new_page()
    page.screenshot(path="example.png")
    browser.close()

```

- If you want to bundle browsers then set `PLAYWRIGHT_BROWSERS_PATH=0` to include bowser's inside the binary and then download the browsers 
- installer pyinstaller 
```sh
pip install pyinstaller
```

- Run pyinstaller on the file like 
```sh
pyinstaller -F main.py
```

- Binary would be placed in `dist/` relative to current dir and test it

### DIFF
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -105,6 +105,10 @@ class PipeTransport(Transport):
             creationflags = subprocess.CREATE_NO_WINDOW
 
         try:
+            env = os.environ.copy()
+            if getattr(sys, "frozen", False):
+                env["PLAYWRIGHT_BROWSERS_PATH"] = "0"
+
             self._proc = proc = await asyncio.create_subprocess_exec(
                 str(self._driver_executable),
                 "run-driver",
@@ -113,6 +117,7 @@ class PipeTransport(Transport):
                 stderr=_get_stderr_fileno(),
                 limit=32768,
                 creationflags=creationflags,
+                env=env,
             )
         except Exception as exc:
             self.on_error_future.set_exception(exc)
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -105,6 +105,7 @@ class PipeTransport(Transport):
             creationflags = subprocess.CREATE_NO_WINDOW
 
         try:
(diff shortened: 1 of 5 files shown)


mxschmitt:
10+ individual users requested it, so makes sense to add support for it.

Fixes #513

mxschmitt (approved):
LGTM, tested with the following test case:
1. `PLAYWRIGHT_BROWSERS_PATH=0 playwright install chromium`
2. `pyinstaller -F test.py`
3. clean browsers to make sure we are not using the wrong ones
4. `PLAYWRIGHT_BROWSERS_PATH=0 ./dist/test`

kumaraditya303:
cc: @pavelfeldman

mxschmitt:
We can land if you somehow can set the env var for the driver automatically. Like detect somehow if its pyinstaller and then set it.

kumaraditya303:
> We can land if you somehow can set the env var for the driver automatically. Like detect somehow if its pyinstaller and then set it.
> 
> Had an offline discussion with him about it.

Done!

playwright/_impl/_transport.py
```diff
@@ -105,6 +105,10 @@ async def run(self) -> None:
             creationflags = subprocess.CREATE_NO_WINDOW
 
         try:
+            env = os.environ.copy()
+            if getattr(sys, "frozen", False):
```
mxschmitt, line 109: nit: comment that its for Pyinstaller.

mxschmitt (approved):
LGTM

playwright/_impl/_transport.py
```diff
@@ -105,6 +105,10 @@ async def run(self) -> None:
             creationflags = subprocess.CREATE_NO_WINDOW
 
         try:
+            env = os.environ.copy()
```
kumaraditya303, line 108:
```suggestion
            # For pyinstaller
            env = os.environ.copy()
```

PR #819, opened 2021-07-23
fix: shutdown driver properly

kumaraditya303:


kumaraditya303:
@mxschmitt Ready for review

playwright/sync_api/_context_manager.py
```diff
@@ -45,8 +45,11 @@ def __enter__(self) -> SyncPlaywright:
             )
 
         def greenlet_main() -> None:
-            loop.run_until_complete(self._connection.run_as_sync())
-
+            try:
```
kumaraditya303, line 48: For review info: 

Here we catch exc and force close the driver and wait for it and then re raise the exc to propagate it to the top of the stack

PR #827, opened 2021-07-29
chore: support client side event logging

mxschmitt:


playwright/_impl/_wait_helper.py
```diff
@@ -121,10 +126,37 @@ def listener(event_data: Any = None) -> None:
...
 
     return asyncio.create_task(throw())
+
+
+def format_log_recording(log: List[str]) -> str:
+    if not log:
+        return ""
+    header = " logs "
+    header_length = 60
+    left_length = math.floor((header_length - len(header)) / 2)
```
yury-s, line 159: we don't have math.floor upstream is it actually required here?

yury-s (approved):

playwright/_impl/_wait_helper.py
mxschmitt: Upstream we do `"o".repeat(2.9)` and String.prototype.repeat does floor internally.

PR #832, opened 2021-08-03
feat(roll): roll Playwright 1.14.0-next-1627933604000

mxschmitt:


playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+
+class Locator:
+    def __init__(self, frame: "Frame", selector: str) -> None:
+        self._frame = frame
+        self._selector = selector
+        self._loop = frame._loop
+        self._dispatcher_fiber = frame._connection._dispatcher_fiber
+
+    def __repr__(self) -> str:
+        return f"<Locator frame='{self._frame._url}' selector='{self._selector}'>"
```
kumaraditya303, line 61:
```suggestion
        return f"<Locator frame={self._frame!} selector={self._selector!r}>"
```
Better to use repr of frame

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+        self,
+        task: Callable[[ElementHandle, Any], Awaitable[T]],
+        timeout: float = None,
+        **kwargs: Any,
+    ) -> T:
+        timeout = self._frame.page._timeout_settings.timeout(timeout)
+        deadline = (monotonic_time() + timeout) if timeout else 0
+        handle = await self.element_handle(timeout=timeout)
+        if not handle:
+            raise Error(f"Could not resolve {self._selector} to DOM Element")
```
kumaraditya303, line 73:
```suggestion
            raise Error(f"Could not resolve {self._selector!r} to DOM Element")
```

playwright/_impl/_transport.py
```diff
@@ -74,7 +74,7 @@ def serialize_message(self, message: Dict) -> bytes:
             print("\x1b[32mSEND>\x1b[0m", json.dumps(message, indent=2))
         return msg.encode()
 
-    def deserialize_message(self, data: bytes) -> ParsedMessagePayload:
+    def deserialize_message(self, data: Union[str, bytes]) -> ParsedMessagePayload:
```
kumaraditya303, line 77: Better to use `AnyStr`

tests/async/test_tap.py
```diff
@@ -190,6 +190,34 @@ async def test_should_wait_until_an_element_is_visible_to_tap_it(page):
     assert await div.text_content() == "clicked"
 
 
+async def test_locators_tap(page: Page):
+    await page.set_content(
+        """"
```
kumaraditya303, line 195:
```suggestion
        """
```

playwright/_impl/_locator.py
mxschmitt: Its always a string, so no need IMO.

playwright/_impl/_transport.py
mxschmitt: I like to be explicit. json.loads accepts `str|byte`

playwright/_impl/_locator.py
kumaraditya303: It is a `str` but `!r` will put in quotes which will be better to spot IMO

playwright/_impl/_transport.py
kumaraditya303: Yup Zen of Python

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+        )
+
+    async def element_handles(self) -> List[ElementHandle]:
+        return await self._frame.query_selector_all(self._selector)
+
+    @property
+    def first(self) -> "Locator":
+        return Locator(self._frame, f"{self._selector} >>  _nth=first")
+
+    @property
```
dgozman, line 204: Why is this a property? Feels strange that first/last are different from nth().

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+        self,
+        timeout: float = None,
+        type: Literal["jpeg", "png"] = None,
+        path: Union[str, pathlib.Path] = None,
+        quality: int = None,
+        omitBackground: bool = None,
+    ) -> bytes:
+        params = locals_to_params(locals())
+        return await self._with_element(
+            lambda h, o: h.screenshot(), strict=True, **params
```
dgozman, line 335: You should pass arguments to h.screenshot()

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+    def __init__(self, frame: "Frame", selector: str) -> None:
+        self._frame = frame
+        self._selector = selector
+        self._loop = frame._loop
+        self._dispatcher_fiber = frame._connection._dispatcher_fiber
+
+    def __repr__(self) -> str:
+        return f"<Locator frame={self._frame!r} selector={self._selector!r}>"
+
+    async def _with_element(
```
dgozman, line 63: I'd suggest the following signature to make it easier for callsites and readers:
```py
async def _with_element(
  self,
  task: Callable[[ElementHandle, int], Awaitable[T]],
  timeout: float = None
) -> T:
```
Basically, you pass updated timeout to the task, so it can do whatever it wants.

Also, I wonder why the result is `T` and not some `Promise<T>` equivalent?

mxschmitt: I like this idea!

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+        timeout = self._frame.page._timeout_settings.timeout(timeout)
+        deadline = (monotonic_time() + timeout) if timeout else 0
+        handle = await self.element_handle(timeout=timeout)
+        if not handle:
+            raise Error(f"Could not resolve {self._selector} to DOM Element")
+        options = {
+            **kwargs,
+            "timeout": (deadline - monotonic_time()) if deadline else 0,
+        }
+        if "strict" in options:
```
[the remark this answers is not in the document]
mxschmitt: with the new approach its not needed anymore. (Reason why we had it before: In python you can't pass additional keyword arguments which don't exist, in JS we have option objects, they don't care)

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+                options,
+            )
+        finally:
+            await handle.dispose()
+
+    async def bounding_box(self, timeout: float = None) -> Optional[FloatRect]:
+        params = locals_to_params(locals())
+        return await self._with_element(
+            lambda e, o: e.bounding_box(),
+            strict=True,
```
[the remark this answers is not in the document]
mxschmitt: This was previously handled via the `**kwargs` which gets all the other not-named arguments. With the new implementation this is also not needed anymore.

playwright/_impl/_locator.py
```diff
@@ -0,0 +1,454 @@
...
+        force: bool = None,
+        noWaitAfter: bool = None,
+        trial: bool = None,
+    ) -> None:
+        params = locals_to_params(locals())
+        return await self._frame.check(self._selector, strict=True, **params)
+
+    async def click(
+        self,
+        modifiers: List[KeyboardModifier] = None,
```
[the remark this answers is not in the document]
mxschmitt: - We have verification scripts which ensure the right parameter are used, with the correct types etc.
- We add another generated layer for sync and async on top. So the user never see's this code. The generated layer includes docs, underscore naming etc.

playwright/_impl/_locator.py
mxschmitt: We currently have the following logic which we use everywhere in Python to decide if its a method or property.
```py
        doc_is_property = (
            not method.get("async") and not len(method["args"]) and "type" in method
        )
        if method["name"].startswith("is_") or method["name"].startswith("as_"):
            doc_is_property = False
        if doc_is_property != is_property:
            self.errors.add(f"Method vs property mismatch: {fqname}")
            return
```
-> it needs to be sync, has no arguments, not start with `is` or `as`. Then it's a property. I would keep it instead of adding more edge-cases.

playwright/_impl/_locator.py
mxschmitt: Good catch!

pavelfeldman (approved):

PR #846, opened 2021-08-14
test: improve ref count test

kumaraditya303:


tests/test_reference_count_async.py
```diff
@@ -42,11 +41,8 @@ async def test_memory_objects() -> None:
 
     gc.collect()
 
-    df_dicts = pd.DataFrame()
-    df_dicts["dicts"] = objgraph.by_type("dict")
-    df_dicts["pw_types"] = df_dicts["dicts"].apply(lambda x: x.get("_type"))
+    pw_types = {d.get("_type") for d in objgraph.by_type("dict")}
```
mxschmitt, line 44:
```suggestion
    dict_objects = {}
    for o in objgraph.by_type("dict"):
        name = o.get("_type")
        if not name:
            continue
        if name not in dict_objects:
            dict_objects[name] = 0
        dict_objects[name] += 1

    assert "Dialog" not in dict_objects
    assert "Request" not in dict_objects
    assert "Route" not in dict_objects
```

This keeps the previous format and helps us when working with the code again, since the amount is helpful.

mxschmitt (changes requested):

mxschmitt (approved):

PR #847, opened 2021-08-14
chore: test on python 3.10

kumaraditya303:
Python 3.10rc is out and hence we should add it to CI to prevent breakage when the final is out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,22 +68,15 @@ jobs:
         - os: ubuntu-latest
           python-version: '3.10-dev'
           browser: chromium
-        # TODO: Enable windows once https://github.com/python-pillow/Pillow/issues/5569 is fixed
-        # - os: windows-latest
-        #   python-version: '3.10-dev'
-        #   browser: chromium
+        - os: windows-latest
+          python-version: '3.10-dev'
+          browser: chromium
         - os: macos-latest
           python-version: '3.10-dev'
           browser: chromium
         - os: macos-11.0
           python-version: '3.10-dev'
           browser: chromium
-        - os: macos-11.0
-          python-version: '3.10-dev'
-          browser: firefox
-        - os: macos-11.0
-          python-version: '3.10-dev'
-          browser: webkit
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,21 +20,20 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.9
-    - name: Install dependencies
+    - name: Install dependencies & browsers
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip wheel
         pip install -r local-requirements.txt
         pip install -e .
         python setup.py bdist_wheel
-        python -m playwright install-deps
-    - name: Install browsers
-      run: python -m playwright install
+        python -m playwright install --with-deps
     - name: Lint
       uses: pre-commit/action@v2.0.3
     - name: Generate APIs
       run: bash scripts/update_api.sh
     - name: Verify generated API is up to date
       run: git diff --exit-code
+
   build:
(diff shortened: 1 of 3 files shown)


tests/sync/test_browsertype_connect.py
```diff
@@ -12,13 +12,18 @@
...
 
+import sys
 import time
 
 import pytest
 
 from playwright.sync_api import BrowserType, Error
 from tests.server import Server
 
+pytestmark = pytest.mark.skipif(
```
mxschmitt, line 23: why a warning and what gives us this warning?

tests/async/test_browsertype_connect.py
```diff
@@ -12,11 +12,17 @@
...
 # limitations under the License.
 
+import sys
+
 import pytest
 
 from playwright.async_api import BrowserType, Error
 from tests.server import Server
 
+pytestmark = pytest.mark.skipif(
```
mxschmitt, line 22: what kind of warning is that?

tests/test_reference_count_async.py
```diff
@@ -42,11 +41,8 @@ async def test_memory_objects() -> None:
 
     gc.collect()
 
-    df_dicts = pd.DataFrame()
-    df_dicts["dicts"] = objgraph.by_type("dict")
-    df_dicts["pw_types"] = df_dicts["dicts"].apply(lambda x: x.get("_type"))
+    pw_types = {d.get("_type") for d in objgraph.by_type("dict")}
```
mxschmitt, line 44: a PR should only contain a single change, please rebase to master.

.github/workflows/ci.yml
```diff
@@ -44,44 +43,24 @@ jobs:
...
-        python-version: [3.7, 3.8]
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-11.0]
+        python-version: [3.7, 3.8, 3.9, '3.10-dev']
         browser: [chromium, firefox, webkit]
-        include:
-        - os: ubuntu-latest
-          python-version: 3.9
-          browser: chromium
-        - os: windows-latest
-          python-version: 3.9
```
mxschmitt, line 55: The reason for manually including it previously was that we ended up having too many jobs. 58 is quite a lot...

mxschmitt (changes requested):

tests/test_reference_count_async.py
kumaraditya303: It was still a draft

tests/sync/test_browsertype_connect.py
kumaraditya303: https://github.com/aaugustin/websockets/issues/1029

tests/async/test_browsertype_connect.py
kumaraditya303: https://github.com/aaugustin/websockets/issues/1029

.github/workflows/ci.yml
kumaraditya303: Now it would be even more :)

mxschmitt:
let's wait until 3.10 is released. Also a few more changes are related:
- add labels in setup.py
- build the python 3.10 conda package

kumaraditya303:
See https://discuss.python.org/t/python-3-10-0rc1-is-now-available/9982, no need to wait for final 3.10 release as libraries should support it from rc itself. 

> add labels in setup.py

Will add.

> build the python 3.10 conda package

Can be done later as conda does not support 3.10 currently.

> wait that a new version of the websocket package was released

A new version is only required if you are using it on 3.10 hence there is a python version marker 
`websockets @ https://github.com/aaugustin/websockets/archive/6edb363af07365867b4c372405b6ab3177a57830.tar.gz; python_version >= '3.10'` , **We are just testing here not using 3.10** and the marker will be removed after a new version is released.

kumaraditya303:
Waiting for https://github.com/microsoft/playwright/commit/0fd5078b2b94eb2d7e51e18181d8d7b95f956d72 to land in python

mxschmitt (approved):

Issue #786, opened 2021-07-02
[FEATURE] Generate Event based overloads for event listeners


kumaraditya303:
Like in the ts playwright, playwright generates event based overloads for the methods, the same in the python is possible with `@overload` and would help in type checking and also provide excellent autocompletions , @mxschmitt Already agree with the  proposal.

/cc : @pavelfeldman 

<img width="486" alt="Screenshot (84)" src="https://user-images.githubusercontent.com/59607654/124219358-dc341c00-db19-11eb-92b1-ef34dd3fdaa1.png">

pavelfeldman:
Sounds good to me

PR #851, opened 2021-08-19
support overloads and add overloads to `wait_for_selector` for when it returns `None`

DetachHead:
- optional arguments are now only automatically converted to keyword arguments if the method has no overloads, because otherwise it can invalidate overloads. it's expected that you explicitly specify what arguments should be keywords, like in the example below

- overloads must be defined using `@api_overload` in order for the generate scripts to be able to see them at runtime.
  ```py
  from playwright._impl._overload import api_overload
  
  @api_overload
  async def wait_for_selector(
      self,
      selector: str,
      *,
      timeout: float = None,
      state: Literal["attached", "visible"] = None,
      strict: bool = None,
  ) -> ElementHandle:
      pass

  @api_overload  # type: ignore[no-redef]
  async def wait_for_selector(
      self,
      selector: str,
      *,
      timeout: float = None,
      state: Literal["detached", "hidden"],
      strict: bool = None,
  ) -> None:
      pass

  async def wait_for_selector(  # type: ignore[no-redef]
      self,
      selector: str,
      *,
      timeout: float = None,
      state: Literal["attached", "detached", "hidden", "visible"] = None,
      strict: bool = None,
  ) -> Optional[ElementHandle]:
      ...
  ```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ _repo_version.py
 coverage.xml
 junit/
 htmldocs/
+.venv
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,12 +49,62 @@ pre-commit run --all-files
 
 For more details look at the [CI configuration](./blob/master/.github/workflows/ci.yml).
 
-### Regenerating APIs
+### APIs
+
+#### Regenerating
 
 ```bash
 ./scripts/update_api.sh
 pre-commit run --all-files
 ```
+
+#### Differences between `_impl` and exposed API method signatures
+
+- optional arguments are automatically converted to keyword arguments, unless the method has overloads. for example:
+  ```py
+  def wait_for_selector(self, selector: str, timeout: float = None, state: str = None): ...
+  ```
+  becomes
+  ```py
+  def wait_for_selector(self, selector: str, *, timeout: float = None, state: str = None): ...
+  ```
+
+- overloads must be defined using `@api_overload` in order for the generate scripts to be able to see them at runtime.
+  ```py
+  from playwright._impl._overload import api_overload
+  
+  @api_overload
+  async def wait_for_selector(
+      self,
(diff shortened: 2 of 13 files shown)


mxschmitt:
Hi! Thank you for your contribution! This needs to be added to the generated sync and async APIs, since we expose them and not the implementation classes. See e.g. in `scripts/generate_async_api.py`.

DetachHead:
@mxschmitt i added support for overloads

mxschmitt:
Hey @DetachHead, sorry for the delay. Will talk later with the team about it. We will likely end up adding two more methods to not have to rely on overloads. This gives us the same behaviour across all our languages (.NET, Java, etc.).

- new: `{Page,ElementHandle,Locator,Frame}.wait_for_selector_hidden(state=Literal["detached", "hidden"]...))`
- new: `{Page,ElementHandle,Locator,Frame}.wait_for_selector_visible(state=Literal["visible", "attached"]...))`

kumaraditya303:
Relates https://github.com/microsoft/playwright-python/issues/786

mxschmitt:
Closing since its stale, waiting also for user reports.

DetachHead:
Why close it?

> Will talk later with the team about it. We will likely end up adding two more methods to not have to rely on overloads (since we're adding in the future languages which don't support overloads). This gives us the same behaviour across all our languages (.NET, Java, etc.).

did anything ever come of this?

DetachHead:
@mxschmitt `typing.get_overloads` is being introduced in python 3.11 https://docs.python.org/3.11/library/typing.html#typing.get_overloads

I can look at redoing this mr using it

KotlinIsland:
@mxschmitt  I think this is a good idea, can we see an update on this at all?

KotlinIsland:
Looks like some hard coded overloads were supported in #897

Issue #852, opened 2021-08-19
[API] Mark API classes with @final


kumaraditya303:
Since playwright exported classes are not supposed to be subclassed hence it should be marked with `@final` decorator to force mypy to not allow subclassing it.

See https://www.python.org/dev/peps/pep-0591/#the-final-decorator

mxschmitt:
Sounds good, do you want to add it?

kumaraditya303:
Yes, will create PR for it.

PR #856, opened 2021-08-20
chore: mark api classes with `@final`

kumaraditya303:
Closes https://github.com/microsoft/playwright-python/issues/852

### DIFF
--- a/scripts/generate_async_api.py
+++ b/scripts/generate_async_api.py
@@ -40,15 +40,16 @@ def generate(t: Any) -> None:
     class_name = short_name(t)
     base_class = t.__bases__[0].__name__
     if class_name in ["Page", "BrowserContext", "Browser"]:
-        base_sync_class = "AsyncContextManager"
+        base_async_class = "AsyncContextManager"
     elif base_class in ["ChannelOwner", "object"]:
-        base_sync_class = "AsyncBase"
+        base_async_class = "AsyncBase"
     else:
-        base_sync_class = base_class
-    print(f"class {class_name}({base_sync_class}):")
+        base_async_class = base_class
+    print("@final")
+    print(
+        f"class {class_name}({base_async_class}): {'# type: ignore' if base_async_class== 'JSHandle' else ''}"
+    )
     print("")
-    print(f"    def __init__(self, obj: {class_name}Impl):")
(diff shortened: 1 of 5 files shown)


scripts/generate_async_api.py
```diff
@@ -40,15 +40,16 @@ def generate(t: Any) -> None:
...
     else:
-        base_sync_class = base_class
-    print(f"class {class_name}({base_sync_class}):")
+        base_async_class = base_class
+    print("@final")
+    print(
+        f"class {class_name}({base_async_class}): {'# type: ignore' if base_async_class== 'JSHandle' else ''}"
+    )
     print("")
-    print(f"    def __init__(self, obj: {class_name}Impl):")
```
kumaraditya303, line 50: This was redundant so removed

scripts/generate_async_api.py
```diff
@@ -40,15 +40,16 @@ def generate(t: Any) -> None:
...
     elif base_class in ["ChannelOwner", "object"]:
-        base_sync_class = "AsyncBase"
+        base_async_class = "AsyncBase"
     else:
-        base_sync_class = base_class
-    print(f"class {class_name}({base_sync_class}):")
+        base_async_class = base_class
+    print("@final")
+    print(
+        f"class {class_name}({base_async_class}): {'# type: ignore' if base_async_class== 'JSHandle' else ''}"
```
kumaraditya303, line 50: `# type: ignore` is for `ElementHandle` as it inherits from `JSHandle`

mxschmitt: is the `@final` thing only for linting or does it actually throw when you try to extend from it again?

symonk: don't think its `runtime_checkable`, if it's only for mypy this is fine, _enforce_ it through other means (__init_subclass__? or metaclass I guess) but not sure if that's what you want to enforce

kumaraditya303:
> Since `@final` seems not common and does not provide a huge benefit we decided to not land this to reduce the complexity overhead.

Huge benefit here = better types, also `@final` was added on high user demand to typing hence it has a dedicated PEP

mxschmitt:
Closing to keep the complexity of the generator logic low and there was no user reports before. But still thank you for your contribution!

PR #880, opened 2021-08-30
chore: Added default user agent for playwright connect

Archish27:


playwright/__init__.py
```diff
@@ -17,3 +17,4 @@
 Firefox and WebKit with a single API. Playwright is built to enable cross-browser
 web automation that is ever-green, capable, reliable and fast.
 """
+__version__ = "1.14.0"
```
mxschmitt, line 20: we don't want to maintain another version, you can get the version from `from playwright._repo_version import version`

playwright/_impl/_browser_type.py
```diff
@@ -187,6 +189,9 @@ async def connect(
         if timeout is None:
             timeout = 30000
 
+        if headers is None:
+            headers = {"User-Agent": getUserAgent()}
```
mxschmitt, line 193: are you sure that you only want to add it when there are no headers given? I would do it on another layer - on the transport directly.

playwright/_impl/_browser_type.py
```diff
@@ -243,3 +248,7 @@ def normalize_launch_params(params: Dict) -> None:
         params["executablePath"] = str(Path(params["executablePath"]))
     if "downloadsPath" in params:
         params["downloadsPath"] = str(Path(params["downloadsPath"]))
+
+
+def getUserAgent() -> str:
+    return "Playwright/{} ({})".format(playwright.__version__, platform.platform())
```
mxschmitt, line 254: can we make this consistent across all the Playwright language bindings?

mxschmitt:
Closing since it seems stale.

PR #883, opened 2021-09-01
fix(1.14): do not trigger playwright install message for deps

mxschmitt:
When the OS dependencies are not satisfied, Playwright throws an exception with `npx playwright install-deps`. This will also trigger this if condition, but it should not. To prevent that for the release we add a space after it.

For 1.15 this logic is not relevant anymore, since we don't wrap it in Python anymore. We throw it from Node.js directly in the correct format.


mxschmitt:
Approved offline by Lusha.

PR #891, opened 2021-09-07
chore: drop 32bit Windows builds

mxschmitt:
https://github.com/microsoft/playwright/issues/8045

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,10 +63,10 @@ class PlaywrightBDistWheelCommand(BDistWheelCommand):
         platform_map = {
             "darwin": "mac",
             "linux": "linux",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -93,9 +93,7 @@ def run(self) -> None:
                 wheel = "macosx_10_13_x86_64.whl"
             if platform == "linux":
                 wheel = "manylinux1_x86_64.whl"
-            if platform == "win32":
-                wheel = "win32.whl"
-            if platform == "win32_x64":
+            if platform == "win":
```
kumaraditya303, line 96: This would break 32 bit python installations on 64 bit windows, better to alias it to 64 bit and 32 bit to same wheel

mxschmitt:
We'll revisit later.

PR #897, opened 2021-09-12
chore: generate events docs

kumaraditya303:
Closes https://github.com/microsoft/playwright-python/issues/786

### DIFF
--- a/playwright/async_api/_generated.py
+++ b/playwright/async_api/_generated.py
@@ -40,6 +40,7 @@ from playwright._impl._api_structures import (
     StorageState,
     ViewportSize,
 )
+from playwright._impl._api_types import Error
 from playwright._impl._async_base import (
     AsyncBase,
     AsyncContextManager,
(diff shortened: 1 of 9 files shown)


setup.cfg
```diff
@@ -14,7 +14,7 @@ warn_redundant_casts = True
 warn_unused_configs = True
 check_untyped_defs = True
 disallow_untyped_defs = True
-[mypy-tests/async.*]
```
mxschmitt, line 17: why this? The reason for that was that our sync tests are now 100% typed to ensure that our typings are good.

scripts/documentation_provider.py
```diff
@@ -63,11 +64,12 @@ def _patch_case(self) -> None:
                 self._add_link(member["kind"], clazz["name"], member_name, new_name)
 
                 if member["kind"] == "event":
-                    continue
+                    clazz["events"].append(member)
 
-                new_name = to_snake_case(new_name)
-                member["name"] = new_name
-                members[new_name] = member
+                if member["kind"] != "event":
```
mxschmitt, line 69:
```suggestion
                else:
```

scripts/documentation_provider.py
```diff
@@ -55,6 +55,7 @@ def _patch_case(self) -> None:
                 continue
             members = {}
             self.classes[clazz["name"]] = clazz
+            clazz["events"] = []
```
mxschmitt, line 58: lets do it like we do for the `members` variable. Initialise it after line 56 and set it in line 97.

scripts/documentation_provider.py
```diff
@@ -386,7 +421,8 @@ def inner_serialize_doc_type(self, type: Any, direction: str) -> str:
                     )
                 )
             return f"{{{', '.join(items)}}}"
-
+        if type_name == "String":
```
mxschmitt, line 424: lets drop this and add in line 428 "or type_name == "String" or e.g. lower().

mxschmitt:
nice work!
https://user-images.githubusercontent.com/17984549/133244167-4bd7e033-ad28-44c9-bee1-4fa5d6aeaf4a.mov

mxschmitt:
Can we have a fallback to the value which we had before? Otherwise we might break existing users which had the event name as a string. e.g. at the end of the on overloads in the Page.

```

    @typing.overload
    def on(
        self,
        event: str,
        f: typing.Callable[..., typing.Union[typing.Awaitable[None], None]],
    ) -> None:
        return super().on(event=event, f=f)
```

setup.cfg
kumaraditya303: That's because of event now emit the objects, feel free to fix the tests mypy

kumaraditya303:
> Can we have a fallback to the value which we had before? Otherwise we might break existing users which had the event name as a string. e.g. at the end of the on overloads in the Page.
> 
> ```
> 
>     @typing.overload
>     def on(
>         self,
>         event: str,
>         f: typing.Callable[..., typing.Union[typing.Awaitable[None], None]],
>     ) -> None:
>         return super().on(event=event, f=f)
> ```

This is non-breaking change moreover it is typing only change hence there is *no*  runtime change

playwright/async_api/_generated.py
```diff
@@ -693,6 +694,100 @@ async def handle(route, request):
...
+        event: Literal["socketerror"],
+        f: typing.Callable[["str"], "typing.Union[typing.Awaitable[None], None]"],
+    ) -> None:
+        """
+        Fired when the websocket has an error."""
+
+    def on(
+        self,
+        event: str,
+        f: typing.Callable[..., typing.Union[typing.Awaitable[None], None]],
```
kumaraditya303, line 740: See the fallback here

mxschmitt (approved):

PR #905, opened 2021-09-15
feat(roll): roll Playwright 1.15.0-next-1631655106000

mxschmitt:


tests/async/test_interception.py
```diff
@@ -204,7 +204,7 @@ async def test_page_route_should_work_with_redirect_inside_sync_XHR(page, server
 
 
 async def test_page_route_should_work_with_custom_referer_headers(page, server):
-    await page.set_extra_http_headers({"referer": server.EMPTY_PAGE})
+    await page.set_extra_http_headers({"Referer": server.EMPTY_PAGE})
```
dgozman, line 207: Why this change? Seems suspicious. Did we have the same change upstream?

dgozman (approved):

tests/async/test_interception.py
mxschmitt: Let me try to repro it upstream.

mxschmitt: fixed, was my wrong implementation!

Issue #873, opened 2021-08-28
[Question]: how to fail test on any unsuccessful status?


nedbat:
### Your question

I would like to be able to write tests that use `goto` and `click`, and have the test automatically fail if the navigation results in an unsuccessful status.

I tried this:

```python
@pytest.fixture(autouse=True)
def fail_on_failure(page):
    def _finished(response):
        assert response.status == 200, f"{response.request.method} {response.url} returned {response.status}"
    page.on("response", _finished)
```
My test had a `page.click` that produced a 500 error.  I see the assert exception in a traceback in my test results, but only because a subsequent assert failed.

I also tried:
```python
@pytest.fixture(autouse=True)
def fail_on_failure(page):
    def _finished(response):
        if response.status == 500:
            pytest.fail(f"{response.request.method} {response.url} returned {response.status}")
    page.on("response", _finished)
```
This caused pytest to freeze after my test with a Fail. The failure seems to be due to the `pytest.fail` (I removed the subsequent assert and the test still failed), but pytest becomes stuck.

Is there a way to systemically fail tests if any request fails?

nedbat:
BTW, for now I am using this helper:
```
def nav(meth, *args, **kwargs):
    with meth.__self__.expect_response("*") as response_info:
        meth(*args, **kwargs)
    response = response_info.value
    assert response.status == 200, f"{response.request.method} {response.url} returned {response.status}"
```
used like this in a test:
```
nav(page.goto, url)
nav(page.click, "input:has-text('Preview')")
```

mxschmitt:
Seems like the greenlets internally deadlock if you want to stop a test inside the reponse event:

```py
import pytest

@pytest.fixture(autouse=True)
def fail_on_failure(page):
    def _finished(response):
        pytest.fail("oh noo")
    page.on("response", _finished)

def test_abc(page):
    page.goto("https://httpstat.us/500")
```

This results in a failing test but it hangs in the teardown.

kumaraditya303:
@mxschmitt 
Reproducer with fix:

```py
import pytest

@pytest.fixture(autouse=True)
def fail_on_failure(page):
    def _finished(response):
        raise RuntimeError

    page.on("response", _finished)


def test_abc(page):
    page.goto("https://httpstat.us/500")

```

Upon running it, `RuntimeError` is just ignored.

**With fix**
```py
import pytest
from greenlet import greenlet


@pytest.fixture(autouse=True)
def fail_on_failure(page):
[...]

Issue #818, opened 2021-07-22
[BUG] Playwright should cleanly shutdown driver in sync_api 


kumaraditya303:
Platform : Windows 10 x64

Minimal reproducer:

```python
from playwright.sync_api import Playwright, sync_playwright


def run(playwright: Playwright) -> None:
    browser = playwright.chromium.launch(headless=False)
    context = browser.new_context()

    page = context.new_page()

    page.goto("https://github.com/")
    page.wait_for_timeout(10e5)
    page.close()

    context.close()
    browser.close()


with sync_playwright() as playwright:
    run(playwright)

```

Wait for the browser to open and then CTRL+C to force exit, playwright does not closes the driver properly and tries to write to non-existing pipes.

Stacktrace:

```console
Traceback (most recent call last):
  File "D:\playwright-python\main.py", line 19, in <module>
    run(playwright)
  File "D:\playwright-python\main.py", line 11, in run
    page.wait_for_timeout(10e5)
  File "D:\playwright-python\playwright\sync_api\_generated.py", line 7600, in wait_for_timeout
    self._sync(
  File "D:\playwright-python\playwright\_impl\_sync_base.py", line 109, in _sync
    self._dispatcher_fiber.switch()
  File "D:\playwright-python\playwright\sync_api\_context_manager.py", line 48, in greenlet_main
    loop.run_until_complete(self._connection.run_as_sync())
  File "c:\users\kumar aditya\appdata\local\programs\python\python39\lib\asyncio\base_events.py", line 629, in run_until_complete
    self.run_forever()
  File "c:\users\kumar aditya\appdata\local\programs\python\python39\lib\asyncio\windows_events.py", line 316, in run_forever
    super().run_forever()
  File "c:\users\kumar aditya\appdata\local\programs\python\python39\lib\asyncio\base_events.py", line 596, in run_forever
    self._run_once()
  File "c:\users\kumar aditya\appdata\local\programs\python\python39\lib\asyncio\base_events.py", line 1854, in _run_once
    event_list = self._selector.select(timeout)
  File "c:\users\kumar aditya\appdata\local\programs\python\python39\lib\asyncio\windows_events.py", line 434, in select
    self._poll(timeout)
  File "c:\users\kumar aditya\appdata\local\programs\python\python39\lib\asyncio\windows_events.py", line 783, in _poll
    status = _overlapped.GetQueuedCompletionStatus(self._iocp, ms)
KeyboardInterrupt
Task was destroyed but it is pending!
[...]

PR #917, opened 2021-09-24
fix: propagate dispatch error to current task

kumaraditya303:
Fixes https://github.com/microsoft/playwright-python/issues/873

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import asyncio
-import sys
 import traceback
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
@@ -47,9 +46,15 @@ class Channel(AsyncIOEventEmitter):
         if params is None:
             params = {}
         callback = self._connection._send_message_to_server(self._guid, method, params)
-
-        done, pending = await asyncio.wait(
-            {self._connection._transport.on_error_future, callback.future},
+        if self._connection._error:
+            error = self._connection._error
+            self._connection._error = None
+            raise error
+        done, _ = await asyncio.wait(
+            {
+                self._connection._transport.on_error_future,
+                callback.future,
+            },
             return_when=asyncio.FIRST_COMPLETED,
         )
         if not callback.future.done():
@@ -152,10 +157,10 @@ class Connection:
         self._callbacks: Dict[int, ProtocolCallback] = {}
         self._object_factory = object_factory
         self._is_sync = False
-        self._api_name = ""
         self._child_ws_connections: List["Connection"] = []
         self._loop = loop
         self._playwright_future: asyncio.Future["Playwright"] = loop.create_future()
+        self._error: Optional[BaseException] = None
 
     async def run_as_sync(self) -> None:
         self._is_sync = True
@@ -260,11 +265,10 @@ class Connection:
                     g.switch(self._replace_guids_with_channels(params))
             else:
                 object._channel.emit(method, self._replace_guids_with_channels(params))
-        except Exception:
-            print(
-                "Error dispatching the event",
-                "".join(traceback.format_exception(*sys.exc_info())),
-            )
+        except BaseException as exc:
+            print("Event Listener Error")
+            traceback.print_exc()
+            self._error = exc
 
     def _create_remote_object(
         self, parent: ChannelOwner, type: str, guid: str, initializer: Dict
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import sys
 import traceback
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
@@ -266,7 +267,7 @@ class Connection:
             else:
                 object._channel.emit(method, self._replace_guids_with_channels(params))
         except BaseException as exc:
-            print("Event Listener Error")
+            print("Error occured in event listener", file=sys.stderr)
             traceback.print_exc()
             self._error = exc
 
--- a/tests/common/test_events.py
+++ b/tests/common/test_events.py
@@ -1,0 +1,18 @@
+from typing import Dict
+
+import pytest
+
+from playwright.sync_api import sync_playwright
+
+
+def test_events(browser_name: str, launch_arguments: Dict) -> None:
+    with pytest.raises(Exception, match="fail"):
+
+        def fail() -> None:
+            raise Exception("fail")
+
+        with sync_playwright() as p:
+            with p[browser_name].launch(**launch_arguments) as browser:
+                with browser.new_page() as page:
+                    page.on("response", lambda _: fail())
+                    page.goto("https://example.com")


kumaraditya303:
cc @mxschmitt

mxschmitt:
Can we have a test for it?  The best would be for sync and async event emitters.

kumaraditya303:
> Can we have a test for it? The best would be for sync and async event emitters which fail without your change.

I have tried it locally, but how would you add a test without the pytest plugin ?

kumaraditya303:
> Can we have a test for it? The best would be for sync and async event emitters which fail without your change.

@mxschmitt Added test

mxschmitt:
When executing the test on my environment it leads to the following:

```txt
(env) ➜  playwright-python git:(master) ✗ pytest --browser chromium tests/common/test_events.py             
Test session starts (platform: darwin, Python 3.9.7, pytest 6.2.4, pytest-sugar 0.9.4)
cachedir: .pytest_cache
rootdir: /Users/max/development/playwright-python, configfile: setup.cfg
plugins: cov-2.12.1, sugar-0.9.4, asyncio-0.15.1, flaky-3.7.0, xdist-2.3.0, repeat-0.9.1, timeout-1.4.2, forked-1.3.0
collecting ... 
 tests/common/test_events.py::test_events[chromium] ✓                                                                                                                                                100% ██████████

Results (3.15s):
       1 passed
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x1060550d0>
  File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_subprocess.py", line 126, in __del__
    self.close()
  File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_subprocess.py", line 104, in close
    proto.pipe.close()
  File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/unix_events.py", line 536, in close
    self._close(None)
  File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/unix_events.py", line 560, in _close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_events.py", line 746, in call_soon
    self._check_closed()
  File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_events.py", line 510, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

kumaraditya303:
That's because of this issue https://github.com/microsoft/playwright-python/issues/818 which out of scope of this PR

kumaraditya303:
Also @mxschmitt that is just a warning as you can see the the test has passed so I would like to push forward this PR the other issue can be fixed by one from tbe team because my solution was rejected but no other fix has been proposed yet

kumaraditya303:
> When executing the test on my environment it leads to the following:
> 
> ```
> (env) ➜  playwright-python git:(master) ✗ pytest --browser chromium tests/common/test_events.py             
> Test session starts (platform: darwin, Python 3.9.7, pytest 6.2.4, pytest-sugar 0.9.4)
> cachedir: .pytest_cache
> rootdir: /Users/max/development/playwright-python, configfile: setup.cfg
> plugins: cov-2.12.1, sugar-0.9.4, asyncio-0.15.1, flaky-3.7.0, xdist-2.3.0, repeat-0.9.1, timeout-1.4.2, forked-1.3.0
> collecting ... 
>  tests/common/test_events.py::test_events[chromium] ✓                                                                                                                                                100% ██████████
> 
> Results (3.15s):
>        1 passed
> Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x1060550d0>
> Traceback (most recent call last):
>   File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_subprocess.py", line 126, in __del__
>     self.close()
>   File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_subprocess.py", line 104, in close
>     proto.pipe.close()
>   File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/unix_events.py", line 536, in close
>     self._close(None)
>   File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/unix_events.py", line 560, in _close
>     self._loop.call_soon(self._call_connection_lost, exc)
>   File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_events.py", line 746, in call_soon
>     self._check_closed()
>   File "/Users/max/.pyenv/versions/3.9.7/lib/python3.9/asyncio/base_events.py", line 510, in _check_closed
>     raise RuntimeError('Event loop is closed')
> RuntimeError: Event loop is closed
> ```

@mxschmitt Fixed in latest commit

playwright/_impl/_connection.py
```diff
@@ -260,11 +263,8 @@ def _dispatch(self, msg: ParsedMessagePayload) -> None:
                     g.switch(self._replace_guids_with_channels(params))
             else:
                 object._channel.emit(method, self._replace_guids_with_channels(params))
-        except Exception:
-            print(
```
mxschmitt, line 264: lets keep the print statement, since not every event emit is followed by a Playwright call IMO.

playwright/_impl/_connection.py
```diff
@@ -48,8 +47,12 @@ async def inner_send(
             params = {}
         callback = self._connection._send_message_to_server(self._guid, method, params)
 
-        done, pending = await asyncio.wait(
-            {self._connection._transport.on_error_future, callback.future},
+        done, _ = await asyncio.wait(
```
mxschmitt, line 50: if self._error:
  e = self._error
  self._error = None
  throw e
should also do it then.

kumaraditya303: Fixed

playwright/_impl/_connection.py
kumaraditya303: Fixed

mxschmitt: Lets re-add the print statement please, being more verbose here is better.

kumaraditya303: That would clutter up the console in Pytest, I don't think print is needed as the exception is fixed is now correctly handled

mxschmitt: Imagine you have an event handler, which throws but you never do something Playwright related afterwards. With your current impl this would then never show the error from my understanding.

kumaraditya303: Added

playwright/_impl/_connection.py
```diff
@@ -260,11 +265,9 @@ def _dispatch(self, msg: ParsedMessagePayload) -> None:
                     g.switch(self._replace_guids_with_channels(params))
             else:
                 object._channel.emit(method, self._replace_guids_with_channels(params))
-        except Exception:
-            print(
-                "Error dispatching the event",
-                "".join(traceback.format_exception(*sys.exc_info())),
-            )
+        except BaseException as exc:
```
mxschmitt, line 268: Please re-add how it previously was. With the current was of logging it the stack does not get logged. So the user never know in which of his event listeners an exception got thrown. Consider the following example:

```py
import sys
import traceback
def main():
    def sub():
        def sub_sub():
            raise Exception("abc")
        sub_sub()
    sub()
try:
    main()
except Exception as exc:
    print(f"Event Listerner Error : {exc}")
    print(20 * "-")
    print(
            "Error dispatching the event",
            "".join(traceback.format_exception(*sys.exc_info())),
        )
```

kumaraditya303: Will do tomorrow :)

kumaraditya303: @mxschmitt fixed

playwright/_impl/_connection.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import asyncio
-import sys
 import traceback
```
mxschmitt, line 16:
```suggestion
import sys
import traceback
```

playwright/_impl/_connection.py
```diff
@@ -260,11 +265,10 @@ def _dispatch(self, msg: ParsedMessagePayload) -> None:
                     g.switch(self._replace_guids_with_channels(params))
             else:
                 object._channel.emit(method, self._replace_guids_with_channels(params))
-        except Exception:
-            print(
-                "Error dispatching the event",
-                "".join(traceback.format_exception(*sys.exc_info())),
-            )
+        except BaseException as exc:
+            print("Event Listener Error")
```
mxschmitt, line 269:
```suggestion
            print("Playwright event could not be dispatched:", file=sys.stderr)
```

mxschmitt (approved):

playwright/_impl/_connection.py
kumaraditya303: This message is misleading because the event was already dispatched then the exc was raised

kumaraditya303: Fixed

PR #925, opened 2021-09-30
chore: roll to 1.16.0-next-1632960932000

kumaraditya303:
- Roll to 1.16.0-next-1632960932000
- Some typing fixes

### DIFF
--- a/playwright/_impl/_helper.py
+++ b/playwright/_impl/_helper.py
@@ -67,7 +67,7 @@ class ErrorPayload(TypedDict, total=False):
     message: str
     name: str
     stack: str
-    value: Any
+    value: Optional[Any]
 
 
 class ContinueParameters(TypedDict, total=False):
@@ -159,7 +159,9 @@ class TimeoutSettings:
 
 
 def serialize_error(ex: Exception, tb: Optional[TracebackType]) -> ErrorPayload:
-    return dict(message=str(ex), name="Error", stack="".join(traceback.format_tb(tb)))
+    return ErrorPayload(
+        message=str(ex), name="Error", stack="".join(traceback.format_tb(tb))
+    )
 
 
 def parse_error(error: ErrorPayload) -> Error:
--- a/playwright/_impl/_locator.py
+++ b/playwright/_impl/_locator.py
@@ -24,7 +24,6 @@ from typing import (
     Optional,
     TypeVar,
     Union,
-    cast,
 )
 
 from playwright._impl._api_structures import FilePayload, FloatRect, Position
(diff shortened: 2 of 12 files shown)


kumaraditya303:
cc @mxschmitt

playwright/_impl/_helper.py
```diff
@@ -67,7 +67,7 @@ class ErrorPayload(TypedDict, total=False):
     message: str
     name: str
     stack: str
-    value: Any
+    value: Optional[Any]
```
mxschmitt, line 70: Doesn't total=False already say that all its properties are optional / can be None?

playwright/_impl/_locator.py
```diff
@@ -173,12 +172,11 @@ async def element_handle(
         timeout: float = None,
     ) -> ElementHandle:
         params = locals_to_params(locals())
-        return cast(
-            ElementHandle,
-            await self._frame.wait_for_selector(
-                self._selector, strict=True, state="attached", **params
-            ),
+        handle = await self._frame.wait_for_selector(
```
mxschmitt, line 175: I would keep the cast here instead of an additional assert. It gets ensured on the server side already.

playwright/_impl/_transport.py
```diff
@@ -23,6 +23,7 @@
 from typing import Callable, Dict, Optional, Union
 
 import websockets
+import websockets.exceptions
```
mxschmitt, line 26: how did this result in a linting error? since the CI was green I'm confused.

playwright/sync_api/__init__.py
```diff
@@ -74,9 +74,7 @@ def sync_playwright() -> PlaywrightContextManager:
 
 
 __all__ = [
-    "async_playwright",
```
mxschmitt, line 77: What is the reasoning that this gets removed?

playwright/_impl/_helper.py
kumaraditya303: Nope it means the keys can be missing but that they are non

playwright/_impl/_transport.py
kumaraditya303: Try hovering over websockets.exceptions it is missing import which is allowed in the mypy config, with this you will get intellisense :)

playwright/_impl/_locator.py
kumaraditya303: Since it is assured on server, an assert is much better and cheaper instruction as compared to a redundant func call, and also friendly to type checkers

playwright/sync_api/__init__.py
kumaraditya303: `async_playwright` should not be included in the sync package

playwright/_impl/_locator.py
mxschmitt: You are right.  From a small experiment assert is 4 times faster compared to a cast.

mxschmitt (approved):

Issue #950, opened 2021-10-14
Switch to driver based browsertype.connect


kumaraditya303:
There will be PR for each:
- [ ] Remove connect api for now #951
- [ ] Refactor connection and transport #952
- [ ] Implement JSON pipe
- [ ] Remove redundant code
- [ ] Some optimizations if necessary / cleanup

PR #951, opened 2021-10-14
chore: remove connect api

kumaraditya303:
Ref https://github.com/microsoft/playwright-python/issues/950


kumaraditya303:
@mxschmitt merge this first

mxschmitt:
Lets instead of merging this one, keep it inside a commit in your PR. So the Git History does not get polluted.

PR #998, opened 2021-10-31
chore: use requests to download drivers

kumaraditya303:
This avoids the use of `curl` which is not available on all platforms by default., on windows cmd.exe it is not available

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     rev: v0.910-1
     hooks:
       - id: mypy
-        additional_dependencies: [types-pyOpenSSL==20.0.6]
+        additional_dependencies: [types-pyOpenSSL==20.0.6, types-requests==2.25.11]
   - repo: https://github.com/pycqa/flake8
     rev: 4.0.1
     hooks:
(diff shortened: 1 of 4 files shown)


carlosperezm:
I guess it does not really make a big difference for this, but have you can considered using [httpx](https://github.com/encode/httpx/) instead of requests. It is more modern and apparently better maintained these days

kumaraditya303:
> I guess it does not really make a big difference for this, but have you can considered using [httpx](https://github.com/encode/httpx/) instead of requests. It is more modern and apparently better maintained these days

Ya, I know about httpx but the dep on requests was already there and we also need to support conda which AFAIK does not has a httpx package

x0day:
this PR ignore default HTTP_PROXY/HTTPS_PROXY environment var.

kumaraditya303:
> this PR ignore default HTTP_PROXY/HTTPS_PROXY environment var.

https://2.python-requests.org/en/master/user/advanced/#proxies

kumaraditya303:
cc @mxschmitt

mxschmitt:
why not: `urllib.request.urlretrieve(url, "driver/" + zip_file)`?

kumaraditya303:
First, requests was already a test dep, second requests has SSL on all platforms whereas urllib doesn't Sent from my Galaxy

kumaraditya303:
@mxschmitt fixed merge conflict

PR #1006, opened 2021-11-05
feat: conda python 3.10

kumaraditya303:


kumaraditya303:
cc @mxschmitt merge

mxschmitt:
Without this change its not possible for Python 3.10 conda users to use Playwright? Can't we publish a single package per OS instead of also per Python version?

kumaraditya303:
> Without this change its not possible for Python 3.10 conda users to use Playwright? Can't we publish a single package per OS instead of also per Python version?

Yes, the latter is out of scope of this PR. Feel free to do it.

mxschmitt:
Do you have some doc pointers how this can be accomplished? Thanks in advance.

kumaraditya303:
> Do you have some doc pointers how this can be accomplished? Thanks in advance.

@mxschmitt  This will be useful https://www.anaconda.com/blog/condas-new-noarch-packages, Can you merge this so that playwright can be installed on 3.10 and improve later on ?

Issue #1005, opened 2021-11-04
FEATURE: Provide all the class properties in repr for debugging


kumaraditya303:
When debugging, better reprs are better hence reprs should have all the class properties

PR #1018, opened 2021-11-10
chore:  frame locator repr

kumaraditya303:
Closes https://github.com/microsoft/playwright-python/issues/1005

### DIFF
--- a/playwright/_impl/_locator.py
+++ b/playwright/_impl/_locator.py
@@ -503,3 +503,6 @@ class FrameLocator:
 
     def nth(self, index: int) -> "FrameLocator":
         return FrameLocator(self._frame, f"{self._frame_selector} >> nth={index}")
+
+    def __repr__(self) -> str:
+        return f"<FrameLocator frame={self._frame!r} selector={self._frame_selector!r}>"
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):

playwright/_impl/_page.py
```diff
@@ -199,7 +199,7 @@ def __init__(
         )
 
     def __repr__(self) -> str:
-        return f"<Page url={self.url!r}>"
+        return f"<Page url={self.url!r}> frame={self.main_frame!r}"
```
mxschmitt, line 202: I think the frame does not provide much value here.

kumaraditya303: Frame has name which can be useful.

tests/sync/test_sync.py
```diff
@@ -43,7 +43,10 @@ def test_sync_query_selector(page: Page) -> None:
 
 def test_page_repr(page: Page) -> None:
     page.goto("https://example.com")
-    assert repr(page) == f"<Page url={page.url!r}>"
+    assert (
+        repr(page)
+        == "<Page url='https://example.com/'> frame=<Frame name= url='https://example.com/'>"
```
mxschmitt, line 48: In my opinion is the page's main frame name always empty, thats why I think we don't have to add the frame to the Page repr.

kumaraditya303: If it's always empty then why does that even exists ?

mxschmitt: Only sub-frames of a page can have named frames. The main frame's name is always an empty string.

mxschmitt (approved):

PR #1036, opened 2021-11-17
feat: added web-first assertions

### LINKED ISSUE #988: Web Assertions for Python binding
https://github.com/microsoft/playwright/issues/8365 for Python

mxschmitt:
Fixes #988

### DIFF
--- a/playwright/_impl/_assertions.py
+++ b/playwright/_impl/_assertions.py
@@ -1,0 +1,413 @@
+# Copyright (c) Microsoft Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+from typing import Any, List, Pattern, Union
+from urllib.parse import urljoin
+
+from playwright._impl._api_structures import ExpectedTextValue, FrameExpectOptions
+from playwright._impl._locator import Locator
+from playwright._impl._page import Page
+
+
+class AssertionsBase:
+    def __init__(self, locator: Locator, is_not: bool = False) -> None:
+        self._actual_locator = locator
+        self._loop = locator._loop
+        self._is_not = is_not
+
+    async def _expect_impl(
+        self,
+        expression: str,
+        expect_options: FrameExpectOptions,
+        expected: Any,
+        message: str,
+    ) -> None:
+        expect_options["isNot"] = self._is_not
+        if expect_options.get("timeout") is None:
+            expect_options["timeout"] = 5_000
+        if expect_options["isNot"]:
+            message = message.replace("expected to", "expected not to")
+        if "useInnerText" in expect_options and expect_options["useInnerText"] is None:
+            del expect_options["useInnerText"]
+        result = await self._actual_locator._expect(expression, expect_options)
+        if result["matches"] == self._is_not:
+            log = "".join(result.get("log", ""))
+            if log.strip():
+                log = "\nCall log:\n" + log
+            if expected is None:
+                raise AssertionError(f"{message} {log}")
+            raise AssertionError(f"{message} {log}")
+
+
+class PageAssertions(AssertionsBase):
+    def __init__(self, page: Page, is_not: bool = False) -> None:
+        super().__init__(page.locator(":root"), is_not)
+        self._actual_page = page
+
+    async def has_title(
+        self, title_or_reg_exp: Union[Pattern, str], timeout: float = None
+    ) -> None:
+        expected_values = to_expected_text_values(
+            [title_or_reg_exp], normalize_white_space=True
+        )
+        await self._expect_impl(
+            "to.have.title",
+            FrameExpectOptions(expectedText=expected_values, timeout=timeout),
+            title_or_reg_exp,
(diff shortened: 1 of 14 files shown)


playwright/_impl/_assertions.py
```diff
@@ -0,0 +1,539 @@
...
+            message = message.replace("expected to", "expected not to")
+        if "useInnerText" in expect_options and expect_options["useInnerText"] is None:
+            del expect_options["useInnerText"]
+        result = await self._actual_locator._expect(expression, expect_options)
+        if result["matches"] == self._is_not:
+            log = "".join(result.get("log", ""))
+            if log.strip():
+                log = "\nCall log:\n" + log
+            if expected is None:
+                raise AssertionError(f"{message} {log}")
```
dgozman, line 50: What's the difference between this and next lines?

playwright/_impl/_assertions.py
```diff
@@ -0,0 +1,539 @@
...
+        self, title_or_reg_exp: Union[Pattern, str], timeout: float = None
+    ) -> None:
+        expected_values = to_expected_text_values(
+            [title_or_reg_exp], normalize_white_space=True
+        )
+        await self._expect_impl(
+            "to.have.title",
+            FrameExpectOptions(expectedText=expected_values, timeout=timeout),
+            title_or_reg_exp,
+            "Page title expected to be",
```
dgozman, line 73: Could you please show an example of how the error looks like?

playwright/_impl/_assertions.py
```diff
@@ -0,0 +1,539 @@
...
+            raise AssertionError(f"{message} {log}")
+
+
+class PageAssertions(AssertionsBase):
+    def __init__(self, page: Page, is_not: bool = False) -> None:
+        super().__init__(page.locator(":root"), is_not)
+        self._actual_page = page
+
+    @property
+    def _not(self) -> "PageAssertions":
```
dgozman, line 60: Is this a public property?

playwright/_impl/_assertions.py
```diff
@@ -0,0 +1,539 @@
...
+            expected["regexFlags"] += "s"
+        if (pattern.flags & int(re.MULTILINE)) != 0:
+            expected["regexFlags"] += "m"
+        assert (
+            pattern.flags
+            & ~(
+                int(re.MULTILINE)
+                | int(re.IGNORECASE)
+                | int(re.DOTALL)
+                | int(re.UNICODE)
```
dgozman, line 514: JS has "u" flag - should we pass it as well?

playwright/_impl/_page.py
```diff
@@ -817,6 +818,10 @@ async def wait_for_function(
     def workers(self) -> List["Worker"]:
         return self._workers.copy()
 
+    @property
+    def request(self) -> "APIRequestContext":
```
dgozman, line 822: Seems unrelated.

playwright/async_api/__init__.py
```diff
@@ -76,7 +82,28 @@ def async_playwright() -> PlaywrightContextManager:
     return PlaywrightContextManager()
 
 
+@overload
+def expect(page_or_locator: Page) -> PageAssertions:
+    ...
```
dgozman, line 87: What is this? Does it mean "pass"? I am confused :)

tests/async/test_assertions.py
```diff
@@ -0,0 +1,293 @@
...
+# limitations under the License.
+
+import re
+from datetime import datetime
+
+import pytest
+
+from playwright.async_api import Browser, Page, expect
+from tests.server import Server
+
```
dgozman, line 22: This does not look like it covers all cases we have it ts, does it?

mxschmitt: This at least gave me 100% test-coverage, do you have any specific tests which you think are relevant for Python?

playwright/async_api/__init__.py
mxschmitt: This is how Python handles overloads. Since the return value when you pass a page is different compared when you pass a locator.

playwright/_impl/_page.py
mxschmitt: Since we roll in this PR, this got caught up by it.

playwright/_impl/_assertions.py
mxschmitt: We didn't do it in Java, thats why I decided to not doing it and in Python every regexp has /u by default.

playwright/_impl/_assertions.py
mxschmitt: It is not, its used as a helper to easily cal the not_* methods.

playwright/_impl/_assertions.py
mxschmitt: https://gist.github.com/mxschmitt/1af1ee657a4237fafb13588b0324ceb3

playwright/_impl/_assertions.py
mxschmitt: In the first one we could access the actual / received values to create a better error message: https://github.com/microsoft/playwright-java/blob/3b4d8dc955f62cca263b1c31c5791a9d9e5fe122/assertions/src/main/java/com/microsoft/playwright/impl/AssertionsBase.java#L66-L69

In this specific case we don't yet, let me dig into formatting.

kumaraditya303:
It would be better to have `expect(...).not.to_have_text` rather than `expect(...).not_to_have_text`.

mxschmitt:
> It would be better to have `expect(...).not.to_have_text` rather than `expect(...).not_to_have_text`.

Yes we also thought the same, but unfortunately `not` is a keyword in Python and thats why we can't use it as a method name. We thought about `not_` instead but it felt wrong to us.

kumaraditya303:
How about `expect_not(...).to_have_text` which is very explicit and concise ?

dgozman (approved):
Looks good to me.

PR #1043, opened 2021-11-19
fix: setup.py --all only put local driver if arch matches

mxschmitt:
Before this CL if you have used setup.py -all it resulted in that the local driver got overridden by the last matching downloaded driver. Only Linux or MacOS this would be then the wrong architecture in most cases.

drive-by: had to rename it a bit, because it conflicted otherwise with the imported platform module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,11 +14,13 @@
 
 import glob
 import os
+import platform
 import shutil
 import subprocess
 import sys
 import zipfile
 from pathlib import Path
+from typing import Dict, List
 
 from setuptools import find_packages, setup
 
@@ -40,6 +42,23 @@ def extractall(zip: zipfile.ZipFile, path: str) -> None:
             os.chmod(extracted_path, attr)
 
 
+def download_driver(zip_name: str) -> None:
+    zip_file = f"playwright-{driver_version}-{zip_name}.zip"
+    if os.path.exists("driver/" + zip_file):
+        return
+    url = "https://playwright.azureedge.net/builds/driver/"
+    if (
+        "-alpha" in driver_version
+        or "-beta" in driver_version
+        or "-next" in driver_version
+    ):
+        url = url + "next/"
+    url = url + zip_file
+    print(f"Fetching {url}")
+    # Don't replace this with urllib - Python won't have certificates to do SSL on all platforms.
+    subprocess.check_call(["curl", url, "-o", "driver/" + zip_file])
+
+
 class PlaywrightBDistWheelCommand(BDistWheelCommand):
     user_options = BDistWheelCommand.user_options + [
         ("all", "a", "create wheels for all platforms")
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -59,57 +60,73 @@ def run(self) -> None:
...
                 "win32": [
                     {
                         "zip_name": "win32_x64",
                         "wheels": ["win32.whl", "win_amd64.whl"],
+                        "machine": "amd64",
                     }
                 ],
             }
-            platforms = [*platform_map.values()]
+            pw_platforms = pw_platform_map.items()
```
pavelfeldman, line 101: Why did this change?

setup.py
```diff
@@ -119,12 +136,15 @@ def run(self) -> None:
         base_wheel_location = glob.glob(os.path.join(self.dist_dir, "*.whl"))[0]
         without_platform = base_wheel_location[:-7]
 
-        for platform in platforms:
-            for arch in platform:
+        for pw_architecture, pw_platform in pw_platforms:
+            for arch in pw_platform:
```
pavelfeldman, line 140: Looks like "machine" is "arch", and this was "platform"?

setup.py
```diff
@@ -119,12 +136,15 @@ def run(self) -> None:
...
                 zip_file = f"driver/playwright-{driver_version}-{arch['zip_name']}.zip"
                 with zipfile.ZipFile(zip_file, "r") as zip:
                     extractall(zip, f"driver/{arch['zip_name']}")
-                if platform_map[sys.platform][0] in platform:
+                if (
+                    arch["machine"] == platform.machine().lower()
+                    and pw_architecture == sys.platform
```
pavelfeldman, line 146: Confused, you can't compare platform to architecture.

kumaraditya303 (approved):

setup.py
```diff
@@ -14,13 +14,16 @@
...
+import platform
 import shutil
 import subprocess
 import sys
 import zipfile
 from pathlib import Path
+from typing import List
 
 from setuptools import find_packages, setup
+from typing_extensions import TypedDict
```
kumaraditya303, line 26: This would break the build when this isn't installed.

setup.py
```diff
@@ -57,96 +68,113 @@ def run(self) -> None:
...
-                    url = url + zip_file
-                    print(f"Fetching {url}")
-                    # Don't replace this with urllib - Python won't have certificates to do SSL on all platforms.
-                    subprocess.check_call(["curl", url, "-o", "driver/" + zip_file])
-        base_wheel_location = glob.glob(os.path.join(self.dist_dir, "*.whl"))[0]
+            wheel_bundles = list(
+                filter(
+                    lambda bundle: bundle["machine"] == platform.machine()
+                    and bundle["platform"] == sys.platform
+                    and bundle["dev_build"] is True,
```
kumaraditya303, line 131:
```suggestion
                    and bundle["dev_build"],
```

setup.py
```diff
@@ -57,96 +68,113 @@ def run(self) -> None:
...
+            zip_file = (
+                f"driver/playwright-{driver_version}-{wheel_bundle['zip_name']}.zip"
+            )
+            with zipfile.ZipFile(zip_file, "r") as zip:
+                extractall(zip, f"driver/{wheel_bundle['zip_name']}")
+            # Driver for local development
+            if (
+                wheel_bundle["machine"] == platform.machine().lower()
+                and wheel_bundle["platform"] == sys.platform
+                and wheel_bundle["dev_build"] is True
```
kumaraditya303, line 161:
```suggestion
                and wheel_bundle["dev_build"]
```

kumaraditya303 (changes requested):

setup.py
mxschmitt: reverted

pavelfeldman (approved):

PR #1086, opened 2021-12-27
Added an if statement that checks if response is in params else adds …

axelande:
Had some issues with a keyError that "respons" was not in parms, it happen very seldom but was really annoying.

### DIFF
--- a/playwright/_impl/_browser_context.py
+++ b/playwright/_impl/_browser_context.py
@@ -137,7 +137,7 @@ class BrowserContext(ChannelOwner):
             "requestFinished",
             lambda params: self._on_request_finished(
                 from_channel(params["request"]),
-                from_nullable_channel(params["response"]),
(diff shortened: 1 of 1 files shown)


playwright/_impl/_browser_context.py
```diff
@@ -137,7 +137,7 @@ def __init__(
             "requestFinished",
             lambda params: self._on_request_finished(
                 from_channel(params["request"]),
-                from_nullable_channel(params["response"]),
+                from_nullable_channel(params["response"]) if "response" in params else False,
```
mxschmitt, line 140:
```suggestion
                from_nullable_channel(params.get("response")),
```

mxschmitt (approved):

mxschmitt:
Seems like we forgot to add it when we backported https://github.com/microsoft/playwright/pull/8569

Issue #1098, opened 2022-01-12
[BUG] Playwright fails if you suppress stderr


AlberLC:
**Context:**
- Playwright Version: 1.17.2
- Operating System: Windows 11 - 64 bits
- Python version: 3.10
- Browser: Chromium

### System:
 - OS: Windows 10 10.0.22000
 - Memory: 8.05 GB / 15.92 GB
### Binaries:
 - Node: 16.13.1 - C:\Program Files\nodejs\node.EXE
 - npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
### Languages:
 - Bash: 4.4.19 - C:\Windows\system32\bash.EXE

## **Minimal example**
```
import asyncio
import contextlib
import os

from playwright.async_api import async_playwright


async def a():
    with open(os.devnull, 'w') as fnull:
        with contextlib.redirect_stderr(fnull):
            await asyncio.sleep(0)


async def b():
    try:
        async with async_playwright():
            print('PLEASE WORK!')
    except ValueError as e:
        print(f':(   {e}')


async def main():
    await asyncio.gather(a(), a(), b())


if __name__ == '__main__':
    asyncio.run(main())

```
## **Expected output**
`PLEASE WORK!`

## **Actual output**
`:(   I/O operation on closed file`

## **Description**
Playwright fails if you suppress stderr.

I have detected it in a telegram bot that I have running with asyncio. If you mix tasks that affect stderr, this error occurs and playwright doesn't work or will not work until the program is completely restarted.

I am ashamed to recognize the amount of hours that I just spent to discover how to reproduce the bug in a minimal example (I have not slept today ._. ).

PR #1100, opened 2022-01-12
Fix: Playwright fails if you suppress stderr (#1098)

AlberLC:


mxschmitt:
Hello, thank you for your contribution! I think it would be better to check if the stderr is closed, and if so, return `None`. Something like that:

```python
        if sys.stderr.closed:
            return None
```

This should lead to the same behaviour but is more explicit.

mxschmitt (approved):

PR #1107, opened 2022-01-16
chore: provide driver with the host-lang version

### LINKED ISSUE #1103: internal: setup env variables when launching driver
1.18 driver assumes the following environment variables are set:

```
PW_CLI_TARGET_LANG=python
PW_CLI_TARGET_LANG_VERSION=<version>
PW_CLI_DISPLAY_VERSION=CLI_VERSION (of the language)
```

`version` should not be too granular to avoid noise; IMO major + minor seems to be good for Python.

mxschmitt:
Fixes #1103

### DIFF
--- a/playwright/__main__.py
+++ b/playwright/__main__.py
@@ -12,20 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import subprocess
 import sys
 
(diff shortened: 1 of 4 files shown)


aslushnikov (approved):
lgtm % test that the default user agent indeed includes Python + version (see Java for the test example - https://github.com/microsoft/playwright-java/pull/774)

PR #1117, opened 2022-01-20
Fixing a pyee DeprecationWarning

shadowwolf899:
pyee.AsyncIOEventEmitter was moved to pyee.asyncio.AsyncIOEventEmitter, so this PR just fixes two imports

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -24,7 +24,7 @@ requirements:
   run:
     - python
     - greenlet >=0.4
-    - pyee >=8.0.1
+    - pyee >=9.0.0
     - websockets >=8.1
     - typing_extensions # [py<39]
 test:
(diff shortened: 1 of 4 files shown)


mxschmitt:
I think we should also bump pyee to `>=9.0.0` inside `setup.py` and `meta.yaml` otherwise it would break.

shadowwolf899:
Went ahead and updated those files

mxschmitt:
It seems like version 9.0 is not available on conda-forge yet: https://github.com/conda-forge/pyee-feedstock/ So we need to look into this one as well. I can try to dig into it tomorrow.

mxschmitt:
I opened https://github.com/conda-forge/pyee-feedstock/pull/15 to fix the pyee conda-forge package. Since this is only about a deprecation warning, we can wait for that a few days.

shadowwolf899:
Thanks! And, definitely no urgency needed. Just wanted to try and get this warning fixed since it comes up every time I run my test scripts

Issue #1148, opened 2022-02-03
[BUG] Pyinstaller Windowed application has no stderr


mickmcc:
**Context:**
- Playwright Version: 1.18.2
- Operating System: Windows 10
- Python version: 3.9.2
- Browser: Chromium (probably All)
- Extra: Pyinstaller 4.8

**Code Snippet**

main.py:
```
from playwright.sync_api import sync_playwright

START_URL = 'https://playwright.dev/'

if __name__ == "__main__":

    exit_code = 1
    with sync_playwright() as p:
        browser = p.chromium.launch_persistent_context(
                                    headless=False,
                                    devtools=False,
                                    user_data_dir='',
                                    no_viewport=True)
        page = browser.new_page()
        try:
            page.wait_for_timeout(10 * 1000)
            page.goto(START_URL,
                      timeout=None)
            page.wait_for_timeout(10 * 1000)
            page.screenshot(path="chromium-screenshot.png")
            exit_code = 0
        except TimeoutError:
            print('Timeout occurred waiting for navigation to: ' + START_URL)
        browser.close()

    sys.exit(exit_code)
```

requirements.txt:
```
playwright==1.18.2
pyinstaller==4.8
```

pyinstaller commands (Batch) - Exception raised:
```
set PLAYWRIGHT_BROWSERS_PATH=0
playwright install chromium
pyinstaller -F --clean --windowed main.py
```

pyinstaller commands (Batch) - No Exception:
```
set PLAYWRIGHT_BROWSERS_PATH=0
playwright install chromium
pyinstaller -F --clean --console main.py
```

**Describe the bug**

Pyinstaller can be used to package applications in two forms:
1.  Console application.  Console window created and output can be written to `sys.stdout`, `sys.stderr`
2.  Windowed application.  No console created, only a GUI window.  No `sys.stdout` or `sys.stderr`.

If a 'windowed' application is created, an exception occurs in Playwright as it tries to access `stderr` but this does not exist.
The use case is an GUI application (Tkinter) which uses Playwright to embed a browser session.

Exception stack trace:
```
    Traceback (most recent call last):
    File "main.py", line 20, in <module>
    File "playwright\sync_api\_context_manager.py", line 70, in __enter__
    File "playwright\sync_api\_context_manager.py", line 48, in greenlet_main
    File "asyncio\base_events.py", line 642, in run_until_complete
    File "playwright\_impl\_connection.py", line 172, in run_as_sync
[...]

PR #1154, opened 2022-02-08
fix: handle pyinstaller app with no stderr (#1148)

mickmcc:
Proposed fix for issue #1148.

When playwright is running in an application which was packaged by pyinstaller, `stderr` can be of type `NullWriter` which has no attribute `closed`.  Moved access to the attribute `closed` inside the following `try` clause to handle the `AttributeError`.
Re-run unit tests after the change an no additional failures seen.

### DIFF
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -37,7 +37,7 @@ def _get_stderr_fileno() -> Optional[int]:
     try:
         # when using pythonw, sys.stderr is None.
         # when Pyinstaller is used, there is no closed attribute because Pyinstaller monkey-patches it with a NullWriter class
-        if sys.stderr is None or not hasattr(sys.stderr, 'closed'):
+        if sys.stderr is None or not hasattr(sys.stderr, "closed"):
             return None
         if sys.stderr.closed:
             return None
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -35,7 +35,8 @@ from playwright._impl._helper import ParsedMessagePayload
 # Sourced from: https://github.com/pytest-dev/pytest/blob/da01ee0a4bb0af780167ecd228ab3ad249511302/src/_pytest/faulthandler.py#L69-L77
 def _get_stderr_fileno() -> Optional[int]:
     try:
-        # handle pythonw, pyinstaller: no stderr
+        # when using pythonw, sys.stderr is None.
+        # when Pyinstaller is used, there is no closed attribute because Pyinstaller monkey-patches it with a NullWriter class
         if sys.stderr is None or not hasattr(sys.stderr, 'closed'):
             return None
         if sys.stderr.closed:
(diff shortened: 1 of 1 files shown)


playwright/_impl/_transport.py
```diff
@@ -34,12 +34,15 @@
...
-    if sys.stderr.closed:
-        return None
-
     try:
+        if sys.stderr.closed:
+            return None
+
         return sys.stderr.fileno()
     except (AttributeError, io.UnsupportedOperation):
+        # pyinstaller apps may not have stderr
```
mxschmitt, line 43: Isn't it pythonw specific?

mxschmitt (approved):

playwright/_impl/_transport.py
mickmcc: Yes, you're right.  I was focused only on pyinstaller but it is specific to pythonw.
I will update the comment.  Would you prefer this scenario to be handled differently in the method?

mxschmitt: It seems alright, but we could also check above if it's a nullwriter which seems cleaner.

mickmcc: I've updated the branch as `pythonw` and `pyinstaller` scenarios are different.

- In case of `pythonw`, `sys.stderr ` is `None`
- In case of `pyinstaller`, `sys.stderr` is an instance of `NullWriter` class from the `pyinstaller` package.

Moved the check for these two situations to early in the `try` clause.  Don't explicitly check for `NullWriter` class as this is from another package so added a generic check if `sys.stderr` has attribute `closed`.

playwright/_impl/_transport.py
```diff
@@ -34,10 +34,13 @@
 
 # Sourced from: https://github.com/pytest-dev/pytest/blob/da01ee0a4bb0af780167ecd228ab3ad249511302/src/_pytest/faulthandler.py#L69-L77
 def _get_stderr_fileno() -> Optional[int]:
-    if sys.stderr.closed:
-        return None
-
     try:
+        # handle pythonw, pyinstaller: no stderr
```
mxschmitt, line 38:
```suggestion
        # when using pythonw, sys.stderr is None.
        # when Pyinstaller is used, there is no closed attribute because Pyinstaller monkey-patches it with a NullWriter class
```

playwright/_impl/_transport.py
```diff
@@ -34,10 +34,14 @@
 
 # Sourced from: https://github.com/pytest-dev/pytest/blob/da01ee0a4bb0af780167ecd228ab3ad249511302/src/_pytest/faulthandler.py#L69-L77
 def _get_stderr_fileno() -> Optional[int]:
-    if sys.stderr.closed:
-        return None
-
     try:
+        # when using pythonw, sys.stderr is None.
+        # when Pyinstaller is used, there is no closed attribute because Pyinstaller monkey-patches it with a NullWriter class
+        if sys.stderr is None or not hasattr(sys.stderr, 'closed'):
```
mxschmitt, line 40:
```suggestion
        if sys.stderr is None or not hasattr(sys.stderr, "closed"):
```

Issue #1134, opened 2022-01-27
PSA: next driver roll will change environment variable names


aslushnikov:
See https://github.com/microsoft/playwright/pull/11661

To apply changes:
```
$ git grep -l PW_CLI_TARGET_LANG | xargs sed -i -e 's/PW_CLI_TARGET_LANG/PW_LANG_NAME/g
$ git grep -l PW_CLI_TARGET_LANG | xargs sed -i -e 's/PW_CLI_TARGET_LANG_VERSION/PW_LANG_NAME_VERSION/g
```

Issue #1120, opened 2022-01-22
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/ab9d5a0dc4744fbfad6be2ed219b81977b6aa97d (chore(tracing): add tracing to APIRequestContext (#11502))
  - [x] https://github.com/microsoft/playwright/commit/f29a673b4c37f1c132ef180938dc20cee0211d3d (fix: do not mute route handler exceptions (#11555))
  - [x] https://github.com/microsoft/playwright/commit/4d42f6a1c4bb182dae5bf614357fc8c47d1c4527 (chore: remove unused _waitingForObject from Connection (#11474))
  - [x] https://github.com/microsoft/playwright/commit/64e7557fb9052a14b02161a9e341e7b350d5c004 (fix: falsey behavior in route.continue, page.post, testInfo.attach (#11421))
  - [x] https://github.com/microsoft/playwright/commit/800b813d4b95a6243f73d7950d5d455b23afe3c4 (chore(test runner): prepare to per-fixture timeout (#11605))
  - [x] https://github.com/microsoft/playwright/commit/687a16b848b9fa7c2a9b92e6b4a9dc96388ba0b1 (feat(locator): introduce locator.page() getter (#11630))
  - [x] https://github.com/microsoft/playwright/commit/f587a43932119fb50e3b81700b7f5eb89b266628 (feat(locator): "has" option (#11411))
  - [x] https://github.com/microsoft/playwright/commit/7912c515a32d8cebaa94c5e815d4a03366ec4e1b (fix(fixtures): account for default options being undefined (#11916))
  - [x] https://github.com/microsoft/playwright/commit/dbd124d84b734f30ad2f57256f492c72b9b58819 (chore: disable page.pause() when JS debugger is attached (#11926))
  - [x] https://github.com/microsoft/playwright/commit/9ce8572c788549f32d32b08554a17aac07feca4f (chore: don't run debugger on pause() (#11948))
  - [x] https://github.com/microsoft/playwright/commit/706c8970314a730efc71a99b3fa2446ea907626c (feat(fixtures): respect tracing config for APIRequestContext (#11954))
  - [x] https://github.com/microsoft/playwright/commit/ae7c52154fca922b0aa6bc5146c8a156d950372b (fix(fixtures): make sure connected browser respects context options (#11990))
[...]

PR #1157, opened 2022-02-11
chore: roll Playwright to 1.19.0

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1120
Fixes https://github.com/microsoft/playwright-python/issues/1134

### DIFF
--- a/playwright/_impl/_network.py
+++ b/playwright/_impl/_network.py
@@ -51,6 +51,7 @@ from playwright._impl._helper import ContinueParameters, locals_to_params
 from playwright._impl._wait_helper import WaitHelper
 
 if TYPE_CHECKING:  # pragma: no cover
+    from playwright._impl._fetch import APIResponse
     from playwright._impl._frame import Frame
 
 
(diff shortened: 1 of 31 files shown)


playwright/_impl/_network.py
```diff
@@ -200,8 +201,25 @@ async def fulfill(
...
         params = locals_to_params(locals())
+        if response:
+            del params["response"]
+            params["status"] = (
+                params["status"] if params.get("status") else response.status
+            )
+            params["headers"] = (
+                params["headers"] if params.get("headers") else response.headers
+            )
+            from playwright._impl._fetch import APIResponse
```
mxschmitt, line 215: this prevents the circular dependency which would otherwise come up.

tests/server.py
```diff
@@ -101,7 +102,9 @@ def process(self) -> None:
                 path = uri.path
 
                 if request_subscribers.get(path):
-                    request_subscribers[path].set_result(request)
+                    loop.call_soon_threadsafe(
```
mxschmitt, line 105: TIL! we call it from another thread into the testing main thread. It was racing and test_frame_wait_for_nav_should_fail_when_frame_detaches was failing then.

pavelfeldman (approved):

PR #1171, opened 2022-02-24
chore: added Docker image publishing

mxschmitt:


utils/docker/publish_docker.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+
+set -e
+set +x
+
+trap "cd $(pwd -P)" EXIT
+cd "$(dirname "$0")"
+
+MCR_IMAGE_NAME="playwright-python"
+PW_VERSION=$(python -c "from playwright._repo_version import version;print(version)")
```
aslushnikov, line 10: Where does "playwright" come from? Do I need to be inside some venv for this to work? If so, sounds fragile! Any chance we can extract version to a separate file and just `cat` it here?

utils/docker/publish_docker.sh
```diff
@@ -0,0 +1,118 @@
...
+        IMAGE_NAMES="${IMAGE_NAMES} ${BASE_IMAGE_TAG}-$3"
+    fi
+    docker manifest create "${BASE_IMAGE_TAG}" $IMAGE_NAMES
+    docker manifest push "${BASE_IMAGE_TAG}"
+  done
+}
+
+publish_docker_images_with_arch_suffix focal amd64
+publish_docker_images_with_arch_suffix focal arm64
+publish_docker_manifest focal amd64 arm64
```
aslushnikov, line 118: I see you don't publish bionic. Do we have a strong signal that we can do so? I'd keep it for now to minimize disturbance of our clients.

utils/docker/build.sh
```diff
@@ -0,0 +1,43 @@
...
+  echo "  docker run --rm -it playwright:localbuild-focal /bin/bash"
+  echo ""
+  echo "NOTE: this requires on Playwright dependencies to be installed with 'npm install'"
+  echo "      and Playwright itself being built with 'npm run build'"
+  echo ""
+  exit 0
+fi
+
+function cleanup() {
+  rm -f "playwright-core.tar.gz"
```
aslushnikov, line 20: seems to be unrelated.

utils/docker/build.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ ($1 == '--help') || ($1 == '-h') || ($1 == '') || ($2 == '') ]]; then
+  echo "usage: $(basename $0) {--arm64,--amd64} {bionic,focal} playwright:localbuild-focal"
+  echo
+  echo "Build Playwright docker image and tag it as 'playwright:localbuild-focal'."
+  echo "Once image is built, you can run it with"
+  echo ""
+  echo "  docker run --rm -it playwright:localbuild-focal /bin/bash"
+  echo ""
+  echo "NOTE: this requires on Playwright dependencies to be installed with 'npm install'"
```
aslushnikov, line 13: this doesn't sound right?

utils/docker/Dockerfile.focal
```diff
@@ -0,0 +1,42 @@
...
+ARG DEBIAN_FRONTEND=noninteractive
+ARG TZ=America/Los_Angeles
+
+# === INSTALL Python ===
+
+RUN apt-get update && \
+    # Install Python
+    apt-get install -y python3 python3-distutils curl && \
+    update-alternatives --install /usr/bin/python python /usr/bin/python3 1 && \
+    curl -sSL https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
```
aslushnikov, line 12: why different lines here?

We currently do the following:
https://github.com/microsoft/playwright/blob/a6fecb5d4b273c934069781c65ae5d50a5007a38/utils/docker/Dockerfile.focal#L16-L20

aslushnikov (approved):

utils/docker/publish_docker.sh
mxschmitt: Does not depend on venv. We install it globally on gha which is common practise.

utils/docker/publish_docker.sh
mxschmitt: I’d drop bionic anyway in 2 months when ubuntu 22 ltd comes out.

utils/docker/Dockerfile.focal
mxschmitt: The apt pip is very very old and does not work with our arm afaik.

Issue #1163, opened 2022-02-14
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/5a0445b8dac2febe9702904d816327769e10886e (feat(remote): let client enable/disable sock proxy (#12086))
  - [x] https://github.com/microsoft/playwright/commit/363b8a697095ccd0ea0dd3a1c1ade136aa555b70 (feat: support `mask` option in screenshot methods (#12072))
  - [x] https://github.com/microsoft/playwright/commit/396d9201458a48ee37cf1e4e51acb51a23df53b1 (feat(test-runner): implement expect(pageOrLocator).toHaveScreenshot (#12242))
  - [x] https://github.com/microsoft/playwright/commit/61a6cdde704b659ba2a8e450db48127c00f64a5b (feat(api): expose locator.highlight (#12420))

PR #1179, opened 2022-03-03
chore: roll Playwright to 1.20.0-alpha-mar-3-2022

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1163


playwright/_impl/_element_handle.py
```diff
@@ -269,10 +270,24 @@ async def screenshot(
         path: Union[str, Path] = None,
         quality: int = None,
         omitBackground: bool = None,
+        disableAnimations: bool = None,
```
pavelfeldman, line 273: You'll be hit by the review, but I think landing it as is still makes sense.

pavelfeldman (approved):

Issue #1152, opened 2022-01-26
[BUG]Exceptions are sometimes reported when the browser is not closed


liuyazhou1991:

aca:
Same here,
```
λ python3 run.py
Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 618225 is closed
<Response [200]>
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x7feccf255750>
Traceback (most recent call last):
  File "/usr/lib/python3.10/asyncio/base_subprocess.py", line 126, in __del__
  File "/usr/lib/python3.10/asyncio/base_subprocess.py", line 104, in close
  File "/usr/lib/python3.10/asyncio/unix_events.py", line 546, in close
  File "/usr/lib/python3.10/asyncio/unix_events.py", line 570, in _close
  File "/usr/lib/python3.10/asyncio/base_events.py", line 745, in call_soon
  File "/usr/lib/python3.10/asyncio/base_events.py", line 510, in _check_closed
RuntimeError: Event loop is closed

λ cat run.py
from playwright.sync_api import sync_playwright

def get_cookies(playwright):
    browser = playwright.chromium.connect_over_cdp("http://127.0.0.1:9222/")
    context = browser.contexts[0]
    cookies = context.cookies("https://google.com")
    return {cookie["name"]: cookie["value"] for cookie in cookies}

cookies =  {}
with sync_playwright() as playwright:
    cookies = get_cookies(playwright)

import requests
print(requests.get("https://google.com", cookies=cookies))
```

Issue #984, opened 2021-10-26
too many chrome processes left after the playwright browsing


XufengXufengXufeng:
![image](https://user-images.githubusercontent.com/31204817/138976976-0af979e8-18d5-4795-ae44-57f8ab1b95f2.png)
i use python api and always use the 'with' clause. but after a few web goto, my system filled with a lot of the chrome processes even though my program has long finished.

kumaraditya303:
@mxschmitt @pavelfeldman Any update on this,when will this be fixed ?

mxschmitt:
Closing since we were not able to reproduce it. Please create a new issue with Python/OS/ etc. and preferable some code which helps to repro it.

cpwan:
Hi, same issure here. 
I am using docker. I installed playwright 1.17.0 with 
```shell
pip install --upgrade pip
pip install playwright
playwright install
```

## Code 
```python
# debug.py
from playwright.sync_api import sync_playwright

with sync_playwright() as p:
    browser = p.chromium.launch(headless=True)
    page = browser.new_page()
    page.close()
    browser.close()
```
## Screenshot
![image](https://user-images.githubusercontent.com/22132868/145949947-261d009d-b82e-4a7b-979e-195e442bf1e2.png)

cpwan:
That was the problem with my container.  Specifying `--init` tag like `docker run --init xxxx` solves my problem.

Issue #939, opened 2021-10-10
How to disable [UnixSelectorEventLoop] warning


re1oaded:
`Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 1919221 is closed`
`Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 1919232 is closed`
`Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 1923783 is closed`
`Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 1923796 is closed`
`Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 1924099 is closed`

A large number of continuous warnings.
UnixSelectorEventLoop always appears on my console and affects the running log of the program I want to print.

I tried many methods and couldn't disable it.

mxschmitt:
Do you have a reproducible script which demonstrates it?

PR #966, opened 2021-10-19
fix: fix process handling bugs

kumaraditya303:
Fixes https://github.com/microsoft/playwright-python/issues/939
Fixes https://github.com/microsoft/playwright-python/issues/818
Fixes https://github.com/microsoft/playwright-python/issues/984

PR #1213, opened 2022-03-18
fix: enhance process close procedure

mxschmitt:
This fixes the following warning which go emitted:

```
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x000002674D05BF78>
  File "C:\Python\lib\asyncio\base_subprocess.py", line 127, in __del__
  File "C:\Python\lib\asyncio\base_subprocess.py", line 104, in close
  File "C:\Python\lib\asyncio\proactor_events.py", line 86, in close
  File "C:\Python\lib\asyncio\base_events.py", line 694, in call_soon
  File "C:\Python\lib\asyncio\base_events.py", line 480, in _check_closed
RuntimeError: Event loop is closed
```

and the following warning which got emitted:

`Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 1923783 is closed`

Fixes https://github.com/microsoft/playwright-python/issues/1152
Fixes https://github.com/microsoft/playwright-python/issues/939
Fixes https://github.com/microsoft/playwright-python/issues/818
Fixes https://github.com/microsoft/playwright-python/issues/984
Fixes https://github.com/microsoft/playwright-python/issues/1254
Relates #966

Repros:

```py

for i in range(300):
    print(i)
        pass
```

```py


def run(playwright):
    browser = playwright.chromium.connect_over_cdp("http://localhost:9222")
    page = browser.new_page()
    page.goto("http://example.com/")


for i in range(300):
    print(i)
        run(playwright)
```


playwright/__main__.py
```diff
@@ -19,9 +19,9 @@
 
 
 def main() -> None:
-    driver_executable = compute_driver_executable()
+    (driver_executable, cli_entrypoint) = compute_driver_executable()
```
pavelfeldman, line 22: Looks unreated

playwright/_impl/_py37ThreadedChildWatcher.py
```diff
@@ -0,0 +1,377 @@
+# This is the LICENSE from the Python project:
```
pavelfeldman, line 1: This does not look like a compatible permissive license we would be able to use.

playwright/_impl/_connection.py
```diff
@@ -180,12 +181,13 @@ async def init() -> None:
             self.playwright_future.set_result(await self._root_object.initialize())
 
         await self._transport.connect()
-        self._loop.create_task(init())
+        self._init_task = self._loop.create_task(init())
```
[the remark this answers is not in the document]
mxschmitt: This is still in use when cleanup is called.

pavelfeldman (approved):

Issue #1208, opened 2022-03-17
[BUG]In response bytes in a non-'UTF-8' encoding format, Apiresponse.Text() will have a decoded error.


Zrincet:
**Context:**
- Playwright Version: 1.20
- Operating System:Windows
- Python version: 3.8
- Browser: Chromium
- Extra: None

**Code Snippet**

```py
def run(playwright):
    browser = playwright.chromium.launch(
        executable_path="C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe"
    )
    page = browser.new_page()
    response = page.request.get("https://qq.com")
    print(response.text())
    browser.close()

with sync_playwright() as playwright:
    run(playwright)
```
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcc in position 50: invalid continuation byte`

**Describe the bug**

In response bytes in a non-'UTF-8' encoding format, Apiresponse.Text() will have a decoded error.
https://github.com/microsoft/playwright-python/blob/417a5b04628c769446665010f9232ec0bfbc7581/playwright/_impl/_fetch.py#L391-L393

PR #1219, opened 2022-03-21
fix: ignore utf-8 decoding errors on APIResponsee

mxschmitt:
https://github.com/microsoft/playwright-python/issues/1208


yury-s (commented):
Can we throw in case the content is not utf-8 ?

Issue #1211, opened 2022-03-18
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/a8d80621b22b62fcae5135860fd666f05010e927 (feat(chromium): large file uploads (#12860))
  - [x] https://github.com/microsoft/playwright/commit/91408f2c5ed4b134165efeec5b86566d8319a455 (fix(storageState): provide nice error message for storage state issues (#13019))
  - [x] https://github.com/microsoft/playwright/commit/07c1a27732997a5a418ea8ec7d3794b49df7367b (chore(deps): prepare deps for monorepo (#13057))
  - [x] https://github.com/microsoft/playwright/commit/5734c18ef8d323cf1ba7335dc41866654b95ae41 (feat(route): add cors header in route.fulfill (#12943))
  - [x] https://github.com/microsoft/playwright/commit/8758cf8cbf7532fdbb4912a97df8d285fd654557 (chore: migrate html reporter to vite (#13116))
  - [x] https://github.com/microsoft/playwright/commit/7d7fe3c618464fd6d9d01afd4e679ba336ddd1ac (fix(route): remove cors option, compare origin (#13231))
  - [x] https://github.com/microsoft/playwright/commit/b0103566c980b35fbc5c0b10f253eb8442f08062 (fix(addInitScript): tolerate trailing comments (#13275))
  - [x] https://github.com/microsoft/playwright/commit/8232497c884e16d8846c6a8585c27691969c7739 (feat(electron): expose app process(), detach on exit (#13280))
  - [x] https://github.com/microsoft/playwright/commit/1bebc28aed3a9fa63972ffa3e83e1db0aa81b938 (chore: reset page/context for reuse in component tests (#13264))
  - [x] https://github.com/microsoft/playwright/commit/a09b8dfa36284c988e36ead2f2f423b244643444 (chore: unrelease 'fonts' screenshot option (#13300))
  - [x] https://github.com/microsoft/playwright/commit/1c3db811f57b7f798dbd789fa7defb0483cb8209 (chore: move to folder-based deps lists (#13361))
  - [x] https://github.com/microsoft/playwright/commit/5ae2017a5be9d465c7777170b8c62b1a71a504bc (chore: always import type (#13365))
  - [x] https://github.com/microsoft/playwright/commit/c80365dd43e5c14589d3274199431519e73fab9b (chore: distill deps after import types update (#13367))
[...]

PR #1255, opened 2022-04-13
feat: large file uploads

rwoll:
Ports:
- https://github.com/microsoft/playwright/commit/a8d80621b22b62fcae5135860fd666f05010e927 (feat(chromium): large file uploads (#12860))
- https://github.com/microsoft/playwright/commit/b0103566c980b35fbc5c0b10f253eb8442f08062 (fix(addInitScript): tolerate trailing comments (#13275))

Fixes https://github.com/microsoft/playwright-python/issues/1211


tests/async/test_browsertype_connect.py
```diff
@@ -241,3 +243,54 @@ async def handle_request(route: Route) -> None:
...
+        request.finish()
+    server.set_route("/upload", handler)
+
+    await page.click("input[type=submit]")
+    request = await file_upload
+    contents = request.args[b'file1'][0]
+    assert len(contents) == 200 * 1024 * 1024
+    assert contents[:1024] == data
+    assert contents[len(contents)-1024:] == data
+    match = re.search(rb'^.*Content-Disposition: form-data; name="(?P<name>.*)"; filename="(?P<filename>.*)".*$', request.post_body, re.MULTILINE)
```
rwoll, line 294: Let's consider using a proper multipart form parsing lib. This works for now, though!

tests/async/test_browsertype_connect.py
```diff
@@ -241,3 +243,54 @@ async def handle_request(route: Route) -> None:
...
+            return events;
+        }
+    """
+    )
+
+    await input.set_input_files(large_file_path)
+    assert await input.evaluate("e => e.files[0].name") == "200MB.zip"
+    assert await events.evaluate("e => e") == ["input", "change"]
+
+    file_upload = asyncio.Future()
```
rwoll, line 282: Can this be re-written without this future?

playwright/_impl/_writable_stream.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @staticmethod
+    async def copy(path: os.PathLike, stream: 'WritableStream'):
+        with open(path, 'rb') as f:
+            while True:
+                data = f.read(COPY_BUFSIZE)
+                if not data:
+                    break
+                await stream.send("write", {"binary": base64.b64encode(data).decode()})
+        await stream.send('close')
```
rwoll, line 40: This seems to work, but I don't think it's proper nor idiomatic w.r.t. to chunked async reading of a file and Channels!

tests/async/test_browsertype_connect.py
mxschmitt: twisted can't get the file name out of the box yeah, you could get the name and it's value via `len(request.args['file1'][0])` but it seems like your approach is fine for now.

playwright/_impl/_element_handle.py
```diff
@@ -190,8 +191,18 @@ async def set_input_files(
...
+        if not frame:
+            raise Error("Cannot set input files to detached element")
+        converted = await convert_input_files(files, frame.page.context)
+        if converted.get("files") is not None:
+            params["files"] = converted.get("files")
+            await self._channel.send("setInputFiles", params)
+        else:
+            await self._channel.send(
+                "setInputFilePaths",
+                locals_to_params({**params, **converted, "files": None}),
```
mxschmitt, line 204: lets try to avoid locals_to_params here and instead either pass all the values or do it like above.

playwright/_impl/_frame.py
```diff
@@ -598,8 +598,15 @@ async def set_input_files(
...
-        params["files"] = await normalize_file_payloads(files)
-        await self._channel.send("setInputFiles", params)
+        converted = await convert_input_files(files, self.page.context)
+        if converted.get("files") is not None:
+            params["files"] = converted.get("files")
+            await self._channel.send("setInputFiles", params)
+        else:
+            await self._channel.send(
+                "setInputFilePaths",
+                locals_to_params({**params, **converted, "files": None}),
```
mxschmitt, line 608: ditto

playwright/_impl/_page.py
```diff
@@ -83,6 +83,7 @@
     serialize_argument,
 )
 from playwright._impl._network import Request, Response, Route, serialize_headers
+from playwright._impl._set_input_files_helpers import convert_input_files
```
mxschmitt, line 86: looks unused

playwright/_impl/_set_input_files_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+
+from playwright._impl._api_structures import FilePayload
+
+SIZE_LIMIT_IN_BYTES = 50 * 1024 * 1024
+
+
+async def convert_input_files(
+    files: Union[str, Path, FilePayload, List[Union[str, Path]], List[FilePayload]],
+    context: "BrowserContext",
+):
```
mxschmitt, line 22: lets define an intermediate return structure like we have upstream, so we have some types which we can use.
```suggestion
) -> InputFilesList:
```

playwright/_impl/_set_input_files_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+    files = files if isinstance(files, list) else [files]
+
+    has_large_buffer = list(
+        filter(
+            lambda f: not isinstance(f, (str, Path))
+            and len(f.get("buffer", "")) > SIZE_LIMIT_IN_BYTES,
+            files,
+        )
+    )
```
mxschmitt, line 31:
```suggestion
    has_large_buffer = any([
            not isinstance(f, (str, Path))
            and len(f.get("buffer", "")) > SIZE_LIMIT_IN_BYTES,
            files,
        for f in files
    ])
```

style nit, should also work.

playwright/_impl/_set_input_files_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+    if has_large_buffer:
+        raise Error(
+            "Cannot set buffer larger than 50Mb, please write it to a file and pass its path instead."
+        )
+
+    has_large_file = list(
```
mxschmitt, line 37: seems like this does not check the upper file limits, lets do something like this instead?

`any([os.stat(f).st_size > SIZE_LIMIT_IN_BYTES for f in files if isinstance(f, (str, Path))])`

playwright/_impl/_set_input_files_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+            for file in files:
+                stream: WritableStream = await context._channel.send(
+                    "createTempFile", {"name": os.path.basename(file)}
+                )
+                await WritableStream.copy(file, stream)
+                streams.append(stream)
+            return {"streams": streams}
+        return {"localPaths": list(map(lambda f: Path(f).absolute().resolve(), files))}
+
+    return {"files": await normalize_file_payloads(files)}
```
mxschmitt, line 55: I'd probably inline normalize_file_payloads now, since we don't use it elsewhere. e

playwright/_impl/_writable_stream.py
mxschmitt: lets ask dima or Yury to review this logic.

tests/async/test_browsertype_connect.py
```diff
@@ -241,3 +243,54 @@ async def handle_request(route: Route) -> None:
...
+
+    browser = await browser_type.connect(remote.ws_endpoint)
+    context = await browser.new_context()
+    page = await context.new_page()
+
+    await page.goto(server.PREFIX + "/input/fileupload.html")
+    large_file_path = tmp_path / "200MB.zip"
+    data = b'A' * 1024
+    with large_file_path.open("wb") as f:
+        for i in range(0, 200 * 1024 * 1024, len(data)):
```
mxschmitt, line 264: nit: lets have it like we have upstream. Instead of 204k for loop iterations only 50k!

tests/async/test_browsertype_connect.py
mxschmitt: Yes, you can use server.wait_for_request

tests/async/test_browsertype_connect.py
```diff
@@ -241,3 +243,54 @@ async def handle_request(route: Route) -> None:
...
+    def handler(request):
+        file_upload.set_result(request)
+        request.finish()
+    server.set_route("/upload", handler)
+
+    await page.click("input[type=submit]")
+    request = await file_upload
+    contents = request.args[b'file1'][0]
+    assert len(contents) == 200 * 1024 * 1024
+    assert contents[:1024] == data
```
mxschmitt, line 292: why is there a 1024 offset?

tests/async/test_input.py
```diff
@@ -284,3 +286,44 @@ async def _listen_for_wheel_events(page: Page, selector: str) -> None:
...
     )
+
+
+@pytest.mark.only_browser("chromium")
+async def test_should_upload_large_file(page, server, tmp_path):
+    await page.goto(server.PREFIX + "/input/fileupload.html")
+    large_file_path = tmp_path / "200MB.zip"
+    data = b'A' * 1024
+    with large_file_path.open("wb") as f:
+        for i in range(0, 200 * 1024 * 1024, len(data)):
```
mxschmitt, line 297: ditto, /4 -> less iterations like we have upstream

tests/async/test_input.py
```diff
@@ -284,3 +286,44 @@ async def _listen_for_wheel_events(page: Page, selector: str) -> None:
...
+
+    await input.set_input_files(large_file_path)
+    assert await input.evaluate("e => e.files[0].name") == "200MB.zip"
+    assert await events.evaluate("e => e") == ["input", "change"]
+
+    file_upload = asyncio.Future()
+    def handler(request):
+        file_upload.set_result(request)
+        request.finish()
+    server.set_route("/upload", handler)
```
mxschmitt, line 319: server.wait_for_request

tests/async/test_input.py
```diff
@@ -284,3 +286,44 @@ async def _listen_for_wheel_events(page: Page, selector: str) -> None:
...
+
+    await page.click("input[type=submit]")
+    request = await file_upload
+    contents = request.args[b'file1'][0]
+    assert len(contents) == 200 * 1024 * 1024
+    assert contents[:1024] == data
+    assert contents[len(contents)-1024:] == data
```
mxschmitt, line 326: ditto, 1024?

tests/async/test_browsertype_connect.py
rwoll: I think this is equivalent to upstream; the step param is `len(data)`

tests/async/test_browsertype_connect.py
rwoll: we're just checking the first and last 1024 bytes in the test; assuming those are correct the other 200MB should be good in between 😄

playwright/_impl/_page.py
rwoll: yes, will be removed when I lint everything, I hope 😄

tests/async/test_browsertype_connect.py
mxschmitt: gotcha

mxschmitt (approved):

PR #1260, opened 2022-04-14
test: enable large file upload tests on all browsers

rwoll:


mxschmitt:
We only run our tests on 11/12, don't we? if I saw it correctly

rwoll:
> We only run our tests on 11/12, don't we? if I saw it correctly

@mxschmitt Good catch—upadted.

Issue #1269, opened 2022-04-22
Wrong ProxySettings type hints


povilasb:
**Context:**

- Playwright Version: 1.21.0
- Operating System: Mac
- Python version: 3.10.1
- Browser: All

**Code Snippet**

```py
import asyncio

from playwright.async_api import async_playwright, ProxySettings


async def main():
    pw = await async_playwright().start()
    browser = await pw.firefox.launch()
    proxy_settings = ProxySettings(
        server="http://localhost:8080", username=None, password=None
    )
    ctx = await browser.new_context(proxy=proxy_settings, ignore_https_errors=True)

    page = await ctx.new_page()
    resp = await page.goto("https://httpbin.org/ip")
    print(resp)


asyncio.run(main())
```

**Describe the bug**

1. run [mitmproxy](https://mitmproxy.org/), or any other HTTP/HTTPS proxy.
2. Run the sample.
3. **playwright** crashes with
```sh
Traceback (most recent call last):
  File "/private/tmp/main.py", line 19, in <module>
    asyncio.run(main())
  File "/Users/povilas/.pyenv/versions/3.10.1/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/povilas/.pyenv/versions/3.10.1/lib/python3.10/asyncio/base_events.py", line 641, in run_until_complete
    return future.result()
  File "/private/tmp/main.py", line 12, in main
    ctx = await browser.new_context(proxy=proxy_settings, ignore_https_errors=True)
  File "/private/tmp/pyenv/lib/python3.10/site-packages/playwright/async_api/_generated.py", line 11317, in new_context
    await self._async(
  File "/private/tmp/pyenv/lib/python3.10/site-packages/playwright/_impl/_browser.py", line 117, in new_context
    channel = await self._channel.send("newContext", params)
  File "/private/tmp/pyenv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 39, in send
    return await self.inner_send(method, params, False)
  File "/private/tmp/pyenv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 63, in inner_send
    result = next(iter(done)).result()
playwright._impl._api_types.Error: proxy.username: expected string, got object
```

4. `ProxySettings` declaration is
```py
class ProxySettings(TypedDict, total=False):
    server: str
    bypass: Optional[str]
    username: Optional[str]
    password: Optional[str]
```

[...]

PR #1271, opened 2022-04-25
chore: fix proxy types and don't allow None

mxschmitt:
`total=False` implies that a property is either allowed to be defined or not. Having additionally `Optional[...]` on some was not correct documentation wise since we don't filter nested dict `None` values out.

Fixes https://github.com/microsoft/playwright-python/issues/1269


dgozman (approved):

KotlinIsland:
Why was this closed?

Issue #1279, opened 2022-05-03
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/54dd6d01e5cf4847b40ded0663e8c3e6a60cd939 (feat(locator): layout options (leftOf, rightOf, above, below, near) (#13821))
  - [x] https://github.com/microsoft/playwright/commit/c3cf7ee4a72d27bd0416e811a095ead1b1349c32 (feat(layout locators): remove maxDistance option (#14013))
  - [x] https://github.com/microsoft/playwright/commit/6931d89eafb5c611113ccf1ab177f63bb0699e8c (feat(locators): rename locator.that to locator.filter (#14025))
  - [x] ~~https://github.com/microsoft/playwright/commit/a052211dbf09d4d5cd4d5283a54e5570ca481130 (chore: do not reset internal bindings for reuse (#14019))~~ (does not apply to Python, for component testing)
  - [x] https://github.com/microsoft/playwright/commit/cf89a361818b031b763910598e8d5967d2de3b6e (chore: serialize circular objects (#14045))

PR #1293, opened 2022-05-10
feat: port layout selectors, filter, circular serialization

rwoll:
Roll driver in preperation for 1.22 release.

Ports the following changes:

- [x] https://github.com/microsoft/playwright/commit/54dd6d01e5cf4847b40ded0663e8c3e6a60cd939 (feat(locator): layout options (leftOf, rightOf, above, below, near) (#13821))
- [x] https://github.com/microsoft/playwright/commit/c3cf7ee4a72d27bd0416e811a095ead1b1349c32 (feat(layout locators): remove maxDistance option (#14013))
- [x] https://github.com/microsoft/playwright/commit/6931d89eafb5c611113ccf1ab177f63bb0699e8c (feat(locators): rename locator.that to locator.filter (#14025))
- [x] https://github.com/microsoft/playwright/commit/cf89a361818b031b763910598e8d5967d2de3b6e (chore: serialize circular objects (#14045))

Resolves https://github.com/microsoft/playwright-python/issues/1279


tests/async/test_queryselector.py
```diff
@@ -179,3 +179,398 @@ async def test_selectors_register_should_handle_errors(selectors, page: Page, ut
     with pytest.raises(Error) as exc:
         await selectors.register("css", dummy_selector_engine_script)
     assert exc.value.message == '"css" is a predefined selector engine'
+
+
+async def test_should_work_with_layout_selectors(page: Page) -> None:
```
rwoll, line 184: Ported from https://github.com/microsoft/playwright/blob/ba0cfaeb2df509e4be93bfc950a4f43a913a2955/tests/page/selectors-misc.spec.ts#L165.

We don't really need _all_ these in the python repo but all are easy enough to port.

playwright/_impl/_locator.py
```diff
@@ -86,6 +93,19 @@ def __init__(
                 raise Error('Inner "has" locator must belong to the same frame.')
             self._selector += " >> has=" + json.dumps(has._selector)
 
+        for inner in ["has", "left_of", "right_of", "above", "below", "near"]:
+            locator: "Locator" = cast("Locator", _params.get(inner))
```
mxschmitt, line 97:
```suggestion
            locator: Optional["Locator"] = cast("Locator", _params.get(inner))
```

playwright/_impl/_locator.py
```diff
@@ -86,6 +93,19 @@ def __init__(
                 raise Error('Inner "has" locator must belong to the same frame.')
             self._selector += " >> has=" + json.dumps(has._selector)
 
+        for inner in ["has", "left_of", "right_of", "above", "below", "near"]:
```
mxschmitt, line 96: we can remove line 91-87 now.

tests/async/test_locators.py
```diff
@@ -748,6 +748,13 @@ async def test_locator_should_enforce_same_frame_for_has_locator(
     assert (
         'Inner "has" locator must belong to the same frame.' in exc_info.value.message
     )
+    for option in ["has", "left_of", "right_of", "above", "below", "near"]:
```
mxschmitt, line 751: we can drop line 746-750 now.

tests/async/test_queryselector.py
mxschmitt: I did the same for .NET, sounds good!

mxschmitt (commented):
lgtm but waiting for the serializer changes.

rwoll:
> lgtm but waiting for the serializer changes.

Circular serializer port underway. Will merge into this PR once I debug and clean up some of the tests: https://github.com/rwoll/playwright-python/compare/ports...ports-circular-handling

NB: In Python there is an extra layer of impl mapping.

Currently debugging the following failure:

```
$ pytest tests/async/test_jshandle.py::test_jshandle_evaluate_should_work_for_circular_objects --browser=firefox
____________________ test_jshandle_evaluate_should_work_for_circular_objects[firefox] _____________________

page = <Page url='about:blank'>

    async def test_jshandle_evaluate_should_work_for_circular_objects(page):
        a = {"x": 1}
        a["y"] = a
>       await page.evaluate("x => x", a)

tests/async/test_jshandle.py:71: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
playwright/async_api/_generated.py:7170: in evaluate
    await self._async(
playwright/_impl/_page.py:409: in evaluate
    return await self._main_frame.evaluate(expression, arg)
playwright/_impl/_frame.py:248: in evaluate
    await self._channel.send(
playwright/_impl/_connection.py:39: in send
    return await self.inner_send(method, params, False)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <playwright._impl._connection.Channel object at 0x110a6c220>, method = 'evaluateExpression'
params = {'arg': {'handles': [], 'value': {'id': 4577179840, 'o': [{'k': 'x', 'v': {'n': 1}}, {'k': 'y', 'v': {'ref': 4577179840}}]}}, 'expression': 'x => x'}
return_as_dict = False

    async def inner_send(
        self, method: str, params: Optional[Dict], return_as_dict: bool
    ) -> Any:
        if params is None:
            params = {}
        if self._connection._error:
            error = self._connection._error
            self._connection._error = None
            raise error
        done, _ = await asyncio.wait(
            {
                self._connection._transport.on_error_future,
                callback.future,
            },
        )
        if not callback.future.done():
            callback.future.cancel()
>       result = next(iter(done)).result()
E       playwright._impl._api_types.Error: Unexpected value

playwright/_impl/_connection.py:63: Error
```

…so looks like somewhere I'm not sending driver an expected value for (at least) circular objects.

rwoll:
Fixed https://github.com/microsoft/playwright-python/pull/1293#issuecomment-1123034860. (Needed a more up-to-date driver.)

The bots have all sorts of other failures, so will need to look into them tomorrow. Some reproduce locally consistently, but others only repro locally 1-2 (of 100) times.

NB: There's still one more client port remaining, and we need to scrutinize use of Python's built-in `id`.

rwoll:
The one remaining port does not look related (as it is a component test context reuse bit).

However, the use of built-in `id` looks potentially problematic. The below fails 3/100 times, but when it fails it looks like the reference resolution is "broken"/mis-aligned:

```
$ pytest --count=100 tests/sync/test_accessibility.py::test_accessibility_should_work_on_a_menu 
…
E       AssertionError: assert {'children': [{'name': 'First Item', 'role': 'menuitem'},\n              [{'name': 'First Item', 'role': 'menuitem'},\n               {'name': 'Second Item', 'role': 'menuitem'},\n               {'name': 'Third Item', 'role': 'menuitem'}],\n              {'name': 'Third Item', 'role': 'menuitem'}],\n 'name': 'My Menu',\n 'role': 'menu'} == {'children': [{'name': 'First Item', 'role': 'menuitem'},\n              {'name': 'Second Item', 'role': 'menuitem'},\n              {'name': 'Third Item', 'role': 'menuitem'}],\n 'name': 'My Menu',\n 'role': 'menu'}
E         Common items:
E         {'name': 'My Menu', 'role': 'menu'}
E         Differing items:
E         {'children': [{'name': 'First Item', 'role': 'menuitem'}, [{'name': 'First Item', 'role': 'menuitem'}, {'name': 'Second Item', 'role': 'menuitem'}, {'name': 'Third Item', 'role': 'menuitem'}], {'name': 'Third Item', 'role': 'menuitem'}]} != {'children': [{'name': 'First Item', 'role': 'menuitem'}, {'name': 'Second Item', 'role': 'menuitem'}, {'name': 'Third Item', 'role': 'menuitem'}]}
E         Full diff:
E           {
E            'children': [{'name': 'First Item',
E                          'role': 'menuitem'},
E         +               [{'name': 'First Item',
E         +                 'role': 'menuitem'},
E         -               {'name': 'Second Item',
E         +                {'name': 'Second Item',
E         ? +
E         -                'role': 'menuitem'},
E         +                 'role': 'menuitem'},
E         ? +
E         +                {'name': 'Third Item',
E         +                 'role': 'menuitem'}],
E                         {'name': 'Third Item',
E                          'role': 'menuitem'}],
E            'name': 'My Menu',
E            'role': 'menu',
E           }

tests/sync/test_accessibility.py:422: AssertionError
```

playwright/_impl/_js_handle.py
```diff
@@ -117,30 +117,38 @@ def serialize_value(value: Any, handles: List[JSHandle], depth: int) -> Any:
     if isinstance(value, str):
         return {"s": value}
 
+    if value in visited:
+        return dict(ref=visited.lookup_id(value))
+
     if isinstance(value, list):
-        result = list(map(lambda a: serialize_value(a, handles, depth + 1), value))
-        return dict(a=result)
+        visited[value] = True
```
pavelfeldman, line 124: It looks like there are two different pattern for StrongDict use: in impl_to_api and here.

playwright/_impl/_visited_util.py
```diff
@@ -0,0 +1,26 @@
+from typing import Any, Dict, List
+
+
+class StrongDict:
```
pavelfeldman, line 4: How about we call it a 'Map' and use it as a map?
```py
K = TypeVar("K")
V = TypeVar("V")
class Map(Generic[K, V]):
```

playwright/_impl/_visited_util.py
```diff
@@ -0,0 +1,26 @@
...
+    """Like a Dict, but can use arbitrary values as keys."""
+
+    def __init__(self) -> None:
+        self._entries: Dict[int, Any] = {}
```
pavelfeldman, line 8: self._entries: Dict[int, Tuple[K, V]] = {}

playwright/_impl/_visited_util.py
```diff
@@ -0,0 +1,26 @@
...
+        # Python built-in |id| only guarantees uniqueness if objects have overlapping lifetimes,
+        # therefore we retain references to ensure overlapping lifetimes and validity of ids.
+        self._do_not_gc: List[Any] = []
```
pavelfeldman, line 11: Use that one dict above for simplicity.

playwright/_impl/_visited_util.py
```diff
@@ -0,0 +1,26 @@
...
+        return id(item) in self._entries
+
+    def __setitem__(self, idx: Any, value: Any) -> None:
+        self._do_not_gc.append(idx)
+        self._entries[id(idx)] = value
+
+    def __getitem__(self, obj: Any) -> Any:
+        return self._entries[id(obj)]
+
+    def lookup_id(self, obj: Any) -> int:
```
pavelfeldman, line 23: This would need to be a userland, so you'll manage the counter externally. Which is good because we don't know how big id() values are - could be too expensive to use them in transport.

playwright/_impl/_js_handle.py
rwoll: Yes, here's we're just using it essentially as a Set (but built in `set(…)` requires hash impl)

pavelfeldman (approved):

PR #1301, opened 2022-05-12
chore: bump dev deps

rwoll:


rwoll:
To fix complaints in https://github.com/microsoft/playwright-python/runs/6414439303?check_suite_focus=true

Issue #1324, opened 2022-05-25
[BUG] Can't register custom selectors using sync API


trogowski:
**Context:**
- Playwright Version: 1.22.0
- Operating System: Windows 11
- Python version: 3.10.2
- Browser: Chromium
- Extra: PDM version 1.14.1

**Code Snippet**

```python
from playwright.sync_api import sync_playwright as pw

max_length_selector = """
    {
      query(root, selector) {
        return root.querySelector(`[maxlength="${selector}"]`);
      },
      queryAll(root, selector) {
        return Array.from(root.querySelectorAll(`[maxlength="${selector}"]`));
      }
    }"""

with pw() as p:
    selectors = p.selectors
    selectors.register('max_length', max_length_selector)

    browser = p.chromium.launch(headless=False)
    page = browser.new_page()
    page.goto('https://google.com')

    search_button = page.locator('max_length=2048')
    print(search_button.count())

    browser.close()
```

**Describe the bug**

`p.selectors` property access throws following error:
```
Traceback (most recent call last):
  File "C:\x\dev\projects\sources\python-playground\web_scrapping\playwright_examples\07_custom_selector.py", line 18, in <module>
    selectors = p.selectors
  File "C:\x\dev\projects\sources\python-playground\web_scrapping\playwright_examples\__pypackages__\3.10\lib\playwright\sync_api\_generated.py", line 11939, in selectors
    return mapping.from_impl(self._impl_obj.selectors)
  File "C:\x\dev\projects\sources\python-playground\web_scrapping\playwright_examples\__pypackages__\3.10\lib\playwright\_impl\_impl_to_api_mapping.py", line 73, in from_impl
    result = self.from_maybe_impl(obj)
  File "C:\x\dev\projects\sources\python-playground\web_scrapping\playwright_examples\__pypackages__\3.10\lib\playwright\_impl\_impl_to_api_mapping.py", line 65, in from_maybe_impl
    api_instance = api_class(obj)
  File "C:\x\dev\projects\sources\python-playground\web_scrapping\playwright_examples\__pypackages__\3.10\lib\playwright\_impl\_sync_base.py", line 72, in __init__
    self._dispatcher_fiber = impl_obj._dispatcher_fiber
AttributeError: 'Selectors' object has no attribute '_dispatcher_fiber'
```
Here is the offending code in _sync_base.py
```python
class SyncBase(ImplWrapper):
    def __init__(self, impl_obj: Any) -> None:
[...]

PR #1325, opened 2022-05-26
fix: sync selectors

mxschmitt:
- pass Playwright dispatching fiber down to the Selector class instance
- port https://github.com/microsoft/playwright-python/blob/97c64908216aec0afe9d6a9a0277fecebc3b2ae3/tests/async/test_queryselector.py to the sync variant

Fixes https://github.com/microsoft/playwright-python/issues/1324


rwoll (approved):
LGTM—and thanks for porting the other tests!

PR #1341, opened 2022-06-01
chore: run the installation tests only with the actual tested browser

mxschmitt:


rwoll (approved):
Assuming bots look good, this looks like an improvement! 😄 LGTM!

PR #1342, opened 2022-06-01
devops: only install required OS dependencies

mxschmitt:
lets see if bots are passing.


rwoll (approved):
All the ❤️ for the Python! 🐍 LGTM

Issue #1350, opened 2022-06-08
[BUG]


miaoningzheng:
**Context:**
- Playwright Version: [1.18.0]
- Operating System: [mac]
- Python version: [3.7]
- Browser: [WebKit]

**Code Snippet**

First, I use node to launch a browser server
```node
const { webkit } = require('playwright'); 

(async () => {
    const browserServer = await webkit.launchServer({headless: false});
    const wsEndpoint = browserServer.wsEndpoint();
    console.log(wsEndpoint)
})();
```
Then I use node client to connect the browser server, it will work well
```node
const { webkit } = require('playwright');


function sleep(ms: number): Promise<void> {
    return new Promise(resolve => setTimeout(()=>{resolve()}, ms))
}


async function run() {
    try {
        const browser = await webkit.connect('ws://127.0.0.1:52558/529cb9771684409f33719e5f66f988ab')
        const page = await browser.newPage({})

        await page.goto('https://yslbeauty.buys.hk/', {timeout: 30000, waitUntil: 'load'});
        await sleep(3000);
        const content = await page.content();
        console.log(content.length);
        await page.close();
    } catch (e) {
        console.error(e)
    }
}


(
    async () => {
        await run()
    }
)()
```
However, if I use python client to connect the browser server, it will raise `Playwright connection closed` error (not every url raise this error)
```python
import asyncio
from playwright.async_api import async_playwright


async def run(playwright):
    webkit = playwright.webkit
    browser = await webkit.connect('ws://127.0.0.1:52558/529cb9771684409f33719e5f66f988ab')
    # browser = await webkit.launch(headless=False)
    page = await browser.new_page()
    await page.goto('https://yslbeauty.buys.hk/', timeout=30000, wait_until='load')
    await asyncio.sleep(3)
    content = await page.content()

    await page.close()
    await browser.close()


async def main():
    async with async_playwright() as playwright:
        await run(playwright)


if __name__ == '__main__':
    asyncio.run(main())
```

**Describe the bug**

When I use python client to connect the browser server and visit some websites, it will raise the error as follow：
```python
Traceback (most recent call last):
  File "/Users/zhengmiaoning/.pyenv/versions/3.8.0/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/Users/zhengmiaoning/.pyenv/versions/3.8.0/lib/python3.8/asyncio/base_events.py", line 608, in run_until_complete
    return future.result()
[...]

Issue #1177, opened 2022-03-01
[internal]: Rewrite BrowserType.connect to use server side connect


mxschmitt:
fixes #1071

PR #1351, opened 2022-06-08
fix(connect): when using a large payload

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1350

Relates https://github.com/microsoft/playwright-python/issues/1177


rwoll (approved):
LGTM! Do you anticipate any perf changes?

mxschmitt:
> LGTM! Do you anticipate any perf changes?

its aligned now the same as we have upstream. So probably better for larger payloads.

Issue #1308, opened 2022-05-17
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/fe0afd6b5c2aa15c5a8e379f16b9a80add047d1b (fix(toHaveProperty): serialize falsy arguments as well (#14232))
  - [x] https://github.com/microsoft/playwright/commit/4c2fc6b6ebf865346b2153611d9e04b3f2abf3ac (feat(browser): add Browser.BrowserType() method (#14468))
  - [x] ~https://github.com/microsoft/playwright/commit/176ab7e48bb73180db3740f50e167d258546206b (chore: make LocalUtils easily available on the client (#14717))~ (maybe unneeded)
  - [x] https://github.com/microsoft/playwright/commit/fccee89b4129d421b8a43764b2ffce02ce7c4747 (feat: add Response.fromServiceWorker flag (#14715))
  - [x] https://github.com/microsoft/playwright/commit/85d03714d9d61dde41a05f7c54acae788f4f84c0 (feat: add allow/block serviceWorkers option (#14714))
  - [x] https://github.com/microsoft/playwright/pull/14724#event-6774053878
  - [x] https://github.com/microsoft/playwright/commit/c7a28ac7e9eab63b40ce22fe96597410dbd08ed7 (fix(test runner): collect artifacts when calling `browser.close()` (#14797))
  - [x] https://github.com/microsoft/playwright/commit/26e22c9a89ef61cabff736073fac9afb267c3030 (fix(types): compilation error in TypeScript 4.8 (#14847))
  - [x] https://github.com/microsoft/playwright/commit/3974d1b3591fa58b644a264e8274d7582650fac5 (revert(#14797): fix(test runner): collect artifacts when calling `browser.close()` (#14846))
  - [x] https://github.com/microsoft/playwright/commit/a1324bd935d15b78d91351a9846126b46bc5bf88 (fix(route): support route w/ async handler & times (#14317))
  - [x] https://github.com/microsoft/playwright/commit/7a568a295266ff40f92393df573fabab26a7d58f (feat(route): chain routes (#14771))
  - [x] https://github.com/microsoft/playwright/commit/dcdd3c3cdb217a86f762a3f34b7acae5b5b039a2 (feat(route): explicitly fall back to the next handler (#14834))
[...]

PR #1357, opened 2022-06-14
chore(roll): roll Playwright to 1.23.0

rwoll:
TODO:
- https://github.com/microsoft/playwright-python/issues/1308

Ports:

- https://github.com/microsoft/playwright/commit/fccee89b4129d421b8a43764b2ffce02ce7c4747 (feat: add Response.fromServiceWorker flag (microsoft/playwright#14715))
- https://github.com/microsoft/playwright/commit/85d03714d9d61dde41a05f7c54acae788f4f84c0 (feat: add allow/block serviceWorkers option (microsoft/playwright#14714))
- https://github.com/microsoft/playwright/commit/4c2fc6b6ebf865346b2153611d9e04b3f2abf3ac (feat(browser): add Browser.BrowserType() method (microsoft/playwright#14468))
- https://github.com/microsoft/playwright/commit/fe0afd6b5c2aa15c5a8e379f16b9a80add047d1b (fix(toHaveProperty): serialize falsy arguments as well (microsoft/playwright#14232))
- https://github.com/microsoft/playwright/commit/978854b859cca8ea86d613fa65358ad8be4ffd81 (chore: move multi-select assertion to toHaveValues (microsoft/playwright#14595))
- https://github.com/microsoft/playwright/commit/d00efa0dfed58e4949d6f5574d8b0c75ad3afe12 (feat(expect): add ignoreCase option to toHaveText and toContainText (microsoft/playwright#14534))
- https://github.com/microsoft/playwright/commit/e00a26a11de55446fae5edc806a7236af725c3f8 (feat(route): fulfill with HARResponse (microsoft/playwright#14865))
- https://github.com/microsoft/playwright/commit/e3da3ebfa483bf970b31c991b30d20a99e164096 (feat(har): do not expose HAR types, remove HARResponse fulfill (#14992))

### DIFF
--- a/playwright/_impl/_api_structures.py
+++ b/playwright/_impl/_api_structures.py
@@ -179,6 +179,7 @@ class ExpectedTextValue(TypedDict, total=False):
     regexFlags: str
     matchSubstring: bool
     normalizeWhiteSpace: bool
+    ignoreCase: Optional[bool]
 
 
 class FrameExpectOptions(TypedDict, total=False):
--- a/playwright/_impl/_assertions.py
+++ b/playwright/_impl/_assertions.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, List, Pattern, Union
+from typing import Any, List, Optional, Pattern, Union
 from urllib.parse import urljoin
 
 from playwright._impl._api_structures import ExpectedTextValue, FrameExpectOptions
@@ -122,11 +122,15 @@ class LocatorAssertions(AssertionsBase):
         expected: Union[List[Union[Pattern, str]], Pattern, str],
         use_inner_text: bool = None,
         timeout: float = None,
+        ignore_case: bool = None,
     ) -> None:
         __tracebackhide__ = True
         if isinstance(expected, list):
             expected_text = to_expected_text_values(
-                expected, match_substring=True, normalize_white_space=True
+                expected,
+                match_substring=True,
+                normalize_white_space=True,
+                ignore_case=ignore_case,
             )
             await self._expect_impl(
                 "to.contain.text.array",
@@ -140,7 +144,10 @@ class LocatorAssertions(AssertionsBase):
             )
         else:
             expected_text = to_expected_text_values(
-                [expected], match_substring=True, normalize_white_space=True
+                [expected],
+                match_substring=True,
+                normalize_white_space=True,
+                ignore_case=ignore_case,
             )
             await self._expect_impl(
                 "to.have.text",
@@ -158,9 +165,10 @@ class LocatorAssertions(AssertionsBase):
         expected: Union[List[Union[Pattern, str]], Pattern, str],
         use_inner_text: bool = None,
         timeout: float = None,
+        ignore_case: bool = None,
     ) -> None:
         __tracebackhide__ = True
-        await self._not.to_contain_text(expected, use_inner_text, timeout)
+        await self._not.to_contain_text(expected, use_inner_text, timeout, ignore_case)
 
     async def to_have_attribute(
         self,
@@ -335,16 +343,41 @@ class LocatorAssertions(AssertionsBase):
         __tracebackhide__ = True
         await self._not.to_have_value(value, timeout)
 
+    async def to_have_values(
+        self,
+        values: List[Union[Pattern, str]],
+        timeout: float = None,
+    ) -> None:
+        __tracebackhide__ = True
+        expected_text = to_expected_text_values(values)
+        await self._expect_impl(
+            "to.have.values",
+            FrameExpectOptions(expectedText=expected_text, timeout=timeout),
+            values,
+            "Locator expected to have Values",
+        )
+
+    async def not_to_have_values(
+        self,
+        values: List[Union[Pattern, str]],
+        timeout: float = None,
+    ) -> None:
+        __tracebackhide__ = True
+        await self._not.to_have_values(values, timeout)
+
     async def to_have_text(
         self,
         expected: Union[List[Union[Pattern, str]], Pattern, str],
         use_inner_text: bool = None,
         timeout: float = None,
+        ignore_case: bool = None,
     ) -> None:
         __tracebackhide__ = True
         if isinstance(expected, list):
             expected_text = to_expected_text_values(
-                expected, normalize_white_space=True
+                expected,
+                normalize_white_space=True,
+                ignore_case=ignore_case,
             )
             await self._expect_impl(
                 "to.have.text.array",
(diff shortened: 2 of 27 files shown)


tests/sync/test_assertions.py
```diff
@@ -158,6 +158,81 @@ def test_assertions_locator_to_have_js_property(page: Page, server: Server) -> N
...
+    locator = page.locator("div")
+    with pytest.raises(AssertionError):
+        expect(locator).to_have_js_property("foo", True, timeout=500)
+
+
+def test_pass_undefined(page: Page) -> None:
+    page.set_content("<div></div>")
+    locator = page.locator("div")
+    expect(locator).to_have_js_property(
+        "foo", None
```
rwoll, line 225: What should this test be in Python? In the Node version, this is `undefined` instead of `None`.

As is, this fails like:

```
    async def test_pass_undefined(page: Page) -> None:
            await page.set_content('<div></div>')
            locator = page.locator('div')
>           await expect(locator).to_have_js_property('foo', None) # Python does not have an undefined
E           AssertionError: Locator expected to have JS Property
E           Actual value: None 
E           Call log:
E           locator_assertions.to_have_js_property with timeout 5000ms
E           waiting for selector "div"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
E             selector resolved to <div></div>
E             unexpected value "undefined"
```

playwright/_impl/_api_structures.py
```diff
@@ -195,3 +196,150 @@ class FrameExpectResult(TypedDict):
     matches: bool
     received: Any
     log: List[str]
+
+
+# see http://www.softwareishard.com/blog/har-12-spec/
```
rwoll, line 201: Something is not quite right as I'm getting when running  `./scripts/update_api.sh` (yet I don't see how this is different from our use of `ResourceTiming` typed dict:

```
(env) rwoll:playwright-python rosswollman$ ./scripts/update_api.sh
Generating playwright/sync_api/_generated.py
Parameter type mismatch in Route.fulfill(response=): documented as Union[APIResponse, HARResponse, NoneType], code has Union[APIResponse, {status: float, statusText: str, httpVersion: str, cookies: List[{name: str, value: str, path: Union[str, NoneType], domain: Union[str, NoneType], expires: Union[str, NoneType], httpOnly: Union[bool, NoneType], secure: Union[bool, NoneType], sameSite: Union[str, NoneType], comment: Union[str, NoneType]}], headers: List[{name: str, value: str, comment: Union[str, NoneType]}], content: {size: float, compression: Union[float, NoneType], mimeType: str, text: Union[str, NoneType], encoding: Union[str, NoneType], comment: Union[str, NoneType]}, redirectURL: str, headersSize: float, bodySize: float, comment: Union[str, NoneType]}, NoneType]
```

rwoll: struct ported from: https://github.com/microsoft/playwright/commit/e00a26a11de55446fae5edc806a7236af725c3f8?diff=unified#diff-3d5c3176db8e25c20b575634893a81a93ca1781de32879502bafc68b462be7afR1

tests/sync/test_assertions.py
mxschmitt: lets talk today in the standup about it.

rwoll: Conclusion: 

- null in Python/.NET should mean undefined on the server side.
- ignore the null case for now, we could add some server side change to handle both cases when undefined is passed in the future.

i.e.

- if you pass null it should assert undefined in the browser
- follow up accepting null and undefined in a separate patch post 1.23

playwright/_impl/_network.py
```diff
@@ -201,24 +207,56 @@ async def fulfill(
...
-                params["headers"] if params.get("headers") else response.headers
-            )
             from playwright._impl._fetch import APIResponse
 
-            if body is None and path is None and isinstance(response, APIResponse):
-                if response._request._connection is self._connection:
-                    params["fetchResponseUid"] = response._fetch_uid
-                else:
-                    body = await response.body()
+            if isinstance(response, APIResponse):
```
rwoll, line 218: This isn't quite correct. This is checking for `playwright._impl._fetch.APIResponse` but the type that ends up here is `playwright.async_api._generated.APIResponse` (or `playwright.sync_api._generated.APIResponse`) based on debugging `test_should_fulfill_with_any_response`.

We can't import the generated types (and the generated types don't subclass `playwright._impl._fetch.APIResponse`), so I could switch this code to be:

```python
if isinstance(response, dict):
   # fulfilling from HAR response
else:
   # fulfilling from some kind of APIResponse
```

@mxschmitt do you have a recommendation, or am I missing something?

playwright/_impl/_assertions.py
```diff
@@ -568,33 +602,46 @@ async def not_to_be_ok(self) -> None:
...
+    ignore_case: Optional[bool] = None,
 ) -> ExpectedTextValue:
     expected = ExpectedTextValue(
         regexSource=pattern.pattern,
         regexFlags=escape_regex_flags(pattern),
         matchSubstring=match_substring,
         normalizeWhiteSpace=normalize_white_space,
+        ignoreCase=ignore_case,
     )
+    if expected["ignoreCase"] is None:
```
mxschmitt, line 617: is the problem that the protocol will yell at you if you pass None -> null?

playwright/_impl/_network.py
```diff
@@ -80,21 +92,22 @@ def __init__(
...
         self._all_headers_future: Optional[asyncio.Future[RawHeaders]] = None
+        self._fallback_overrides: FallbackOverrides = FallbackOverrides()
 
     def __repr__(self) -> str:
         return f"<Request url={self.url!r} method={self.method!r}>"
 
     @property
     def url(self) -> str:
-        return self._initializer["url"]
+        return self._fallback_overrides.get("url") or self._initializer["url"]
```
mxschmitt, line 102:
```suggestion
        return self._fallback_overrides.get("url", self._initializer["url"])
```

playwright/_impl/_browser_context.py
```diff
@@ -96,8 +96,11 @@ def __init__(
         )
         self._channel.on(
             "route",
-            lambda params: self._on_route(
-                from_channel(params.get("route")), from_channel(params.get("request"))
+            lambda params: asyncio.create_task(
```
rwoll, line 99: I think this will cause problems for the sync variant.

rwoll:
We're going to land this in pieces, so closing this for now in favor of some upcoming CLs.

Issue #1220, opened 2022-03-21
[BUG] Hide Page.WaitForEventInfo from trace viewer


mxschmitt:
<img width="452" alt="image" src="https://user-images.githubusercontent.com/17984549/159305395-a0ff6009-26ae-4a30-8c4d-0347b23bc1f1.png">

PR #1361, opened 2022-06-15
chore: rework apiName api call handling in connection

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1220


playwright/_impl/_connection.py
```diff
@@ -36,10 +49,21 @@ def __init__(self, connection: "Connection", guid: str) -> None:
...
+        )
 
     async def send_return_as_dict(self, method: str, params: Dict = None) -> Any:
-        return await self.inner_send(method, params, True)
+        return await self._connection.wrap_api_call(
+            lambda: self.inner_send(method, params, True)
+        )
+
+    def send_no_reply(self, method: str, params: Dict = None) -> None:
+        self._connection.wrap_api_call_sync(
```
pavelfeldman, line 62: Why do we need a separate method for this?

playwright/_impl/_connection.py
```diff
@@ -337,6 +354,40 @@ def _replace_guids_with_channels(self, payload: Any) -> Any:
             return result
         return payload
 
+    async def wrap_api_call(
+        self, cb: Callable[[], Coroutine], is_internal: bool = False
+    ) -> Any:
+        zones = self._api_zone.get()
```
pavelfeldman, line 360: Why do we use array of zones?

playwright/_impl/_connection.py
```diff
@@ -337,6 +354,40 @@ def _replace_guids_with_channels(self, payload: Any) -> Any:
...
+        if zones[0] is not None:
+            result = cb()
+            return (
+                await cast(Coroutine, result) if asyncio.iscoroutine(result) else result
+            )
+        task = asyncio.current_task(self._loop)
+        st: List[inspect.FrameInfo] = getattr(task, "__pw_stack__", inspect.stack())
```
pavelfeldman, line 367: This assumes there were no async hops on the way from API call to wrap_api_call. I believe this is the case, but I'm not sure we should rely upon in. Consider leaving _async wrapper in place as a safer option.

pavelfeldman (approved):

PR #1375, opened 2022-06-24
chore: allow forking control flow in route

pavelfeldman:


playwright/_impl/_browser_context.py
```diff
@@ -262,7 +262,12 @@ async def route(
     ) -> None:
         self._routes.insert(
             0,
-            RouteHandler(URLMatcher(self._options.get("baseURL"), url), handler, times),
+            RouteHandler(
+                URLMatcher(self._options.get("baseURL"), url),
+                handler,
+                True if self._dispatcher_fiber else False,
```
rwoll, line 268:
```suggestion
                bool(self._dispatcher_fiber),
```

playwright/_impl/_helper.py
```diff
@@ -208,23 +210,34 @@ def __init__(
...
+            result = cast(
+                Callable[["Route", "Request"], Union[Coroutine, Any]], self.handler
+            )(route, request)
+            if inspect.iscoroutine(result):
+                asyncio.create_task(result)
+
+        # As with event handlers, each route handler is a potentially blocking context
+        # so it needs a fiber.
+        if self._is_sync:
+            g = greenlet(impl)
```
rwoll, line 237: Ah…so this is the key bit here to prevent blocking where we were getting stuck before?

PR #1382, opened 2022-06-25
chore: port record_har_* options (content, mode, url_filter)

rwoll:
This is part 3/n of the 1.23 port.

Relates #1308, #1374, #1376.

Ports:

- [x] https://github.com/microsoft/playwright/commit/fdcdd58d7fb128de2a26828eca4afa78c70047a8 (feat(har): introduce urlFilter (#14693))
- [x] https://github.com/microsoft/playwright/commit/c349c1d57f6a60813aa433fcfc39e0bce4c164c5 (feat: newContext.har (#14892))
- [x] https://github.com/microsoft/playwright/commit/245c33a5d44e6acd93b532e1166587fa716c560d (feat(har): allow storing content as separate files (#14934))
- [x] https://github.com/microsoft/playwright/commit/be64e9ce66bdd1e5b40c79b85a77b61a0c93b00a (chore(har): attach resources for .zip hars (#14938))
- [x] https://github.com/microsoft/playwright/commit/7bd72716f960ed11b48447c22f49a3b4ee453eb5 (feat(har): introduce the slim mode (#15053))

PR #1376, opened 2022-06-24
chore: port route chaining, fallback, async, times

rwoll:
This is part 2/n of the 1.23 port.

Relates #1308, #1374

Ports:

  - [x] https://github.com/microsoft/playwright/commit/a1324bd935d15b78d91351a9846126b46bc5bf88 (fix(route): support route w/ async handler & times (#14317))
  - [x] https://github.com/microsoft/playwright/commit/7a568a295266ff40f92393df573fabab26a7d58f (feat(route): chain routes (#14771))
  - [x] https://github.com/microsoft/playwright/commit/dcdd3c3cdb217a86f762a3f34b7acae5b5b039a2 (feat(route): explicitly fall back to the next handler (#14834))
  - [x] https://github.com/microsoft/playwright/commit/9cf068ad061f8b134dae98c8f000fe7b1b5214f0 (feat(fallback): allow falling back w/ overrides (#14849))
  - [x] https://github.com/microsoft/playwright/commit/48f98673ef423f34d960f9641d7cc16176b6ed5c (chore: remove stray fallback overrides check)
  - [x] https://github.com/microsoft/playwright/commit/ae6f48c4b8bb55db693dfdfd27592d9206e0d46d (fix(route): match against updated url while chaining (#15112))

PR #1374, opened 2022-06-23
chore(roll): roll Playwright to 1.23.0-beta-1656026605000

rwoll:
This is part 1/n of the 1.23 port.

Relates #1308.

Ports:

  - [x] https://github.com/microsoft/playwright/commit/fe0afd6b5c2aa15c5a8e379f16b9a80add047d1b (fix(toHaveProperty): serialize falsy arguments as well (#14232))
  - [x] https://github.com/microsoft/playwright/commit/4c2fc6b6ebf865346b2153611d9e04b3f2abf3ac (feat(browser): add Browser.BrowserType() method (#14468))
  - [x] https://github.com/microsoft/playwright/commit/fccee89b4129d421b8a43764b2ffce02ce7c4747 (feat: add Response.fromServiceWorker flag (#14715))
  - [x] https://github.com/microsoft/playwright/commit/85d03714d9d61dde41a05f7c54acae788f4f84c0 (feat: add allow/block serviceWorkers option (#14714))
  - [x] https://github.com/microsoft/playwright/commit/c7a28ac7e9eab63b40ce22fe96597410dbd08ed7 (fix(test runner): collect artifacts when calling `browser.close()` (#14797))
  - [x] https://github.com/microsoft/playwright/commit/cdb862767f408fa6d364179d449f562565b4a793 (fix(page): "load" event should fire before "waitForLoadState" resolves (#14897))
  - [x] https://github.com/microsoft/playwright/commit/7f026dd64c7d1d2b540ab9a4132b318c51f242ac Language override behavior changed upstream in https://github.com/WebKit/WebKit/commit/039ebd9

PR #1384, opened 2022-06-26
chore: port routeFromHar

rwoll:
> ⚠️ **DO NOT REVIEW (yet)**: The latest commit is the actual (small-ish) change, but it depends on some of the prior PRs to land first! Once those land, I'll drop the landed commits out of this PR.

This is part 5/n of the 1.23 port.

Relates #1308, #1374, #1376, #1382, #1383

Ports:

…


playwright/_impl/_har_router.py
```diff
@@ -0,0 +1,104 @@
...
+                status=response.get("status"),
+                headers={
+                    v["name"]: v["value"]
+                    for v in cast(HeadersArray, response.get("headers", []))
+                },
+                body=base64.b64decode(body),
+            )
+            return
+
+        if action == "error":
```
mxschmitt, line 78: the nesting level here seems off!

rwoll: > ⚠️ DO NOT REVIEW (yet):

😄 it is off—but already fixed in a local commit! 🎉

playwright/_impl/_local_utils.py
```diff
@@ -26,3 +28,45 @@ def __init__(
...
+                        level: {
+                            "phase": "log",
+                            "message": message,
+                        },
+                    },
+                ),
+                True,
+            )
+        except Exception:
+            pass
```
rwoll, line 72: @mxschmitt Please advise on the recommended way for logging in this repo. This is based off of the code in the `_wait_helper.py`, but I suspect `method` cannot be an arbitrary string. As of now, I'm calling it like so elsewhere:

```
          self._local_utils.log(
                "debug",
                "api",
                f"HAR: {route.request.url} redirected to {redirect_url}",
            )
```

but I suspect it's wrong even though I see some output when piping through grep.

mxschmitt: We don't have logging via RPC, lets just not log or use the logging module? This way seems wrong and I doubt it will work.

mxschmitt (approved):

PR #1394, opened 2022-07-01
Test Python 3.11 beta5 in CI

tonybaloney:
Python 3.11 RC1 is now a month away, this PR expands the test suite to include 3.11-dev


tonybaloney:
FYI, the current setup spec _doesn't_ infer that 3.11 isn't supported

https://github.com/microsoft/playwright-python/blob/main/setup.py#L228

mxschmitt:
Yeah its blocked by an upstream greenlet issue: https://github.com/python-greenlet/greenlet/issues/305

PR #1408, opened 2022-07-05
test: clean up tracing tests

mxschmitt:


rwoll (approved):
LGTM—I think. So the change is, instead of checking a subset of events, ensure we have no extra by explicitly matching on all actions? This ensure the tests catches extra/unintended events showing up?

If so, please put in commit message/PR description the impetus for this change.

Issue #1402, opened 2022-07-04
[Question]: Task was destroyed but it is pending! in route.abort()


cyberfuhrer:
### Your question

Why am I getting an error in route.abort()?
Example:
```
from playwright.sync_api import sync_playwright

def handle_route(route):
    if route.request.resource_type == "image":
        response = page.request.fetch(route.request)
	if int(response.headers['content-length']) > 500:
	    route.abort()
	else:
	    route.continue_()
    else:
        route.continue_()

playwright = sync_playwright().start()
browser = playwright.chromium.launch()
context = browser.new_context()
context.route('**/*', handle_route)
page = context.new_page()
page.goto(url='https://www.pinterest.com/', timeout=10000)
page.wait_for_timeout(3000)
page.close()
context.close()
browser.close()
```
```
Task was destroyed but it is pending!
task: <Task pending coro=<BrowserContext._on_route() running at /usr/local/lib/python3.7/dist-packages/playwright/_impl/_browser_context.py:187> wait_for=<_GatheringFuture pending cb=[<TaskWakeupMethWrapper object at 0x7f2c168870d8>()]>>
...
```
Playwright version 1.23.0

rwoll:
Thanks for the report—your code looks good and this is likely a leak introduced by the new requests handling code. It will be addressed, but it should be safe to ignore the warning for now.

rwoll:
Re-opening. The fix we landed might be non-exhaustive, and I'm not longer convinced it's the correct fix.

ja8zyjits:
@rwoll should we add a `self._background_task_tracker.close()` in the `page._on_close()` method?

rwoll:
> @rwoll shouldn't we add a `self._background_task_tracker.close()` in the `page._on_close()` method too?

I don't think that's exhaustive. There's a bunch of places where we call create_task and don't await, so it looks like the proper fix is to loop over the pending tasks when we go to exit the Python PW context manager cancelling. I'm currently finishing off some other work, but I will be looping back to this shortly. (To my knowledge, the current release has a leak and annoying log line sometimes, but is not functionally broken/blocking folks. If this is incorrect, please let me know.)

Thanks for your patience!

ja8zyjits:
> I don't think that's exhaustive.


At present the issue is mostly at the Page class only, so we need not worry about other areas. If we could clean up at the end of page then that must fix the issue, according to my understanding. Cleaning from the Python PW context manager will be good, but that shall be something that we can look into later.




[...]

PR #1430, opened 2022-07-13
fix(tasks): cancel all the pending tasks as we exit own loop

pavelfeldman:
Fixes https://github.com/microsoft/playwright-python/issues/1402


dgozman (approved):

rwoll:
Thanks! I'll follow up with a test addition.

ja8zyjits:
I dont think closing all the background tasks would be the right thing to do. Since we might have background async tasks of the remaining tests running.
This will cancel everything and make the running program hang.

PR #1436, opened 2022-07-14
chore: updated requirements

GabrielGavrilov:
Recently I've re-installed Python from a clean slate, and when I was setting up playwright-python for contributing, I've noticed that two packages weren't on the local-requirements.txt

So I've updated the requirements file to include those two packages.


mxschmitt:
These dependencies are listed inside the setup.py. So you need to do `python install -e .` to get them.

See here: https://github.com/microsoft/playwright-python/blob/main/CONTRIBUTING.md

PR #1440, opened 2022-07-18
chore: constrain Pattern type

### LINKED ISSUE #1415: [BUG] (🐞) `typing.Pattern` used in type hints are not supplied type parameters
https://github.com/microsoft/playwright-python/blob/c8d8f4a6859ac210fa6b6f01eb636fc11065abb9/playwright/sync_api/_generated.py#L3060-L3066
```py
page.expect_navigation(
  re.compile("abc")  # error: expression contains Any
)
```
`typing.Pattern` expects one type parameter, usually `AnyStr`.

rwoll:
Fixes #1415.

Most of this change was generated. The actual changed code (aside from find and replace) is in `//scripts/documentation_provider.py`.

NB: Using bytes in patterns with our code will currently explode, so
this constrains to hint to users to use str patterns:

```py
# OK
await expect(page).to_have_url(re.compile(r".*/grid\.html"))
# BAD
await expect(page).to_have_url(re.compile(rb".*/grid\.html"))
```

This change opts to constrain the type rather than allow for bytes, too. If a user wants to use bytes, they can encode/decode explicitly.

### DIFF
--- a/scripts/documentation_provider.py
+++ b/scripts/documentation_provider.py
@@ -372,8 +372,7 @@ class DocumentationProvider:
             args = get_args(value)
             return f"Callable[{', '.join(list(map(lambda a: self.serialize_python_type(a), args)))}]"
         if str(origin) == "<class 're.Pattern'>":
-            args = get_args(value)
-            return f"Pattern[{', '.join(list(map(lambda a: self.serialize_python_type(a), args)))}]"
+            return "Pattern[str]"
         if str(origin) == "typing.Literal":
             args = get_args(value)
             if len(args) == 1:
--- a/scripts/documentation_provider.py
+++ b/scripts/documentation_provider.py
@@ -371,6 +371,9 @@ class DocumentationProvider:
         if str(origin) == "<class 'collections.abc.Callable'>":
             args = get_args(value)
             return f"Callable[{', '.join(list(map(lambda a: self.serialize_python_type(a), args)))}]"
+        if str(origin) == "<class 're.Pattern'>":
+            args = get_args(value)
+            return f"Pattern[{', '.join(list(map(lambda a: self.serialize_python_type(a), args)))}]"
         if str(origin) == "typing.Literal":
             args = get_args(value)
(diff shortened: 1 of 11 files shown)


scripts/documentation_provider.py
```diff
@@ -371,6 +371,9 @@ def serialize_python_type(self, value: Any) -> str:
         if str(origin) == "<class 'collections.abc.Callable'>":
             args = get_args(value)
             return f"Callable[{', '.join(list(map(lambda a: self.serialize_python_type(a), args)))}]"
+        if str(origin) == "<class 're.Pattern'>":
```
mxschmitt, line 374: for what is this and why does the line 376 needs to be that complex? Feels like its always returning Pattern[str] anyway.

rwoll: > for what is this

Without this, we'll get a bunch of errors like:

```
Parameter type mismatch in BrowserContext.route_from_har(url=): documented as Union[Pattern[str], str, NoneType], code has Union[str, typing.Pattern[str], NoneType]
```

Notice the extra `typing.` prefix. Types with arguments (like `Callable`) are cleaned of the `typing` prefix above, while the simple cases are handled with:

```
        match = re.match(r"^typing\.(\w+)$", str_value)
        if match:
            return match.group(1)
```

further up.


Line 375:

`f"Pattern[{', '.join(list(map(lambda a: self.serialize_python_type(a), args)))}]"`

could be replaced with:

`"Pattern[str]"`

but then if we go to add `bytes` (although at that point `AnyStr` should be used), it's nice to have the multi-arg support like the surround lines are already doing.

We could also just remove this altogether and on the second line of the function do:

```
    def serialize_python_type(self, value: Any) -> str:
        str_value = str(value)
        str_value = str_value.replace('tying.Pattern', 'Pattern') # simple replace!
```

What's your preference?

mxschmitt (approved):

KotlinIsland:
Cool, just wondering why you use a custom `NoneType` instead of `None`?

PR #1474, opened 2022-08-01
chore: prevent save_as from hanging in sync mode

### LINKED ISSUE #1462: [BUG] Execution hangs when trying to save video or delete video before calling page.close()
**Context:**
- Playwright Version: 1.23
- Operating System: Windows
- Python: 3.9
- Browser: All

**Code Snippet**

```from playwright.sync_api import Playwright, sync_playwright


def run(playwright: Playwright) -> None:
    browser = playwright.chromium.launch(headless=False)
    context = browser.new_context(
            viewport={"width": 1920, "height": 1080},
            record_video_dir="temp_videos/",
            record_video_size={"width": 1920, "height": 1080})

    # Open new page
    page = context.new_page()

    # ---------------------
    # page.video.save_as("test.webm")
    # OR
    # page.video.delete()
    context.close()
    browser.close()


    run(playwright)
```

**Describe the bug**

Execution will hang, no stack trace will be produced when user tries to save video or delete video before closing the page (page.close)

Uncomment line 15 or 17 to reproduce

The docs for save_as suggest that it should be possible:
"Saves the video to a user-specified path. It is safe to call this method while the video is still in progress, or after the page has closed. "

Still in progress suggests that I do not need to page.close() first

rwoll:
Resolves #1462.

Docs: https://github.com/microsoft/playwright/pull/16120

### DIFF
--- a/playwright/_impl/_video.py
+++ b/playwright/_impl/_video.py
@@ -55,6 +55,10 @@ class Video:
         return artifact.absolute_path
 
     async def save_as(self, path: Union[str, pathlib.Path]) -> None:
+        if self._page._connection._is_sync and not self._page._is_closed:
+            raise Error(
+                "Page is not yet closed. Close the page prior to calling save_as"
+            )
         artifact = await self._artifact_future
         if not artifact:
             raise Error("Page did not produce any video frames")
--- a/tests/async/test_video.py
+++ b/tests/async/test_video.py
@@ -60,4 +60,5 @@ async def test_should_not_error_if_page_not_closed_before_save_as(
     saved = page.video.save_as(out_path)
     await page.close()
     await saved
-    os.path.exists(out_path)
+    await page.context.close()
+    assert os.path.exists(out_path)
--- a/tests/async/test_video.py
+++ b/tests/async/test_video.py
@@ -49,3 +49,15 @@ async def test_short_video_should_throw_persistent_context(
 
     path = await page.video.path()
     assert str(tmpdir) in str(path)
+
+
+async def test_should_not_error_if_page_not_closed_before_save_as(
+    browser, tmpdir, server
+):
+    page = await browser.new_page(record_video_dir=tmpdir)
+    await page.goto(server.PREFIX + "/grid.html")
+    out_path = tmpdir / "some-video.webm"
+    saved = page.video.save_as(out_path)
+    await page.close()
+    await saved
+    os.path.exists(out_path)
--- a/tests/sync/test_video.py
+++ b/tests/sync/test_video.py
@@ -106,3 +106,8 @@ def test_should_error_if_page_not_closed_before_save_as(
         assert video
         video.save_as(out_path)
     assert "Page is not yet closed. Close the page prior to calling save_as" in str(err)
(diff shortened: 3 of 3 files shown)


playwright/_impl/_video.py
```diff
@@ -55,6 +55,10 @@ async def path(self) -> pathlib.Path:
         return artifact.absolute_path
 
     async def save_as(self, path: Union[str, pathlib.Path]) -> None:
+        if self._page._connection._is_sync and not self._page._is_closed:
+            raise Error(
+                "Page is not yet closed. Close the page prior to calling save_as"
+            )
```
rwoll, line 61: Maybe this should be in all artifacts, not just video

playwright/_impl/_video.py
```diff
@@ -55,6 +55,10 @@ async def path(self) -> pathlib.Path:
         return artifact.absolute_path
 
     async def save_as(self, path: Union[str, pathlib.Path]) -> None:
+        if self._page._connection._is_sync and not self._page._is_closed:
+            raise Error(
+                "Page is not yet closed. Close the page prior to calling save_as"
```
mxschmitt, line 60:
```suggestion
                "Page is not yet closed. Close the page prior to calling Video.save_as"
```

tests/async/test_video.py
```diff
@@ -49,3 +49,15 @@ async def test_short_video_should_throw_persistent_context(
...
+async def test_should_not_error_if_page_not_closed_before_save_as(
+    browser, tmpdir, server
+):
+    page = await browser.new_page(record_video_dir=tmpdir)
+    await page.goto(server.PREFIX + "/grid.html")
+    out_path = tmpdir / "some-video.webm"
+    saved = page.video.save_as(out_path)
+    await page.close()
+    await saved
+    os.path.exists(out_path)
```
mxschmitt, line 63:
```suggestion
    assert os.path.exists(out_path)
```

tests/sync/test_video.py
```diff
@@ -91,3 +93,16 @@ def test_record_video_can_get_video_path_immediately(
...
+    browser: Browser, tmpdir: Path, server: Server
+) -> None:
+    page = browser.new_page(record_video_dir=tmpdir)
+    page.goto(server.PREFIX + "/grid.html")
+    out_path = tmpdir / "some-video.webm"
+    with pytest.raises(Error) as err:
+        video = page.video
+        assert video
+        video.save_as(out_path)
+    assert "Page is not yet closed. Close the page prior to calling save_as" in str(err)
```
mxschmitt, line 108: who closes the context?

rwoll: fixed

tests/async/test_video.py
rwoll: updated (also added corresponding check in the error case, and then an assurance we can still save after)

playwright/_impl/_video.py
rwoll: Left as is—this looks funny since it's not clear what `Video` is in this context. (In this case, just a type.)

Most will be calling it like `page.video.save_as(…)` or `video.save_as(…)`.

PR #1483, opened 2022-08-07
`pre-commit` version upgrades

symonk:
Hi, small patch to bump pre-commit hook versions and silence the current `mypy` complaints about the sync api base.

Thanks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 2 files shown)


rwoll:
Thanks—I've triggered the CI to run and check things are working as expected.

rwoll:
It looks like there's a new mypy error showing up in the Lint job:

```
scripts/documentation_provider.py:208: error: Assignment expressions are only supported in Python 3.8 and greater
```

I don't think this is related to your change, since it also shows up in other PRs from today. I cannot re-produce the error locally (or in a fresh GitHub Codespace). @mxschmitt any ideas? Seems like it's an FP

mxschmitt (approved):

PR #1486, opened 2022-08-10
chore: update dev dependencies

mxschmitt:
Update dev dependencies using [pur](https://pypi.org/project/pur/) and `pre-commit autoupdate`


local-requirements.txt
```diff
@@ -1,24 +1,24 @@
...
 pytest-repeat==0.9.1
 pytest-timeout==2.1.0
 pytest-xdist==2.5.0
-requests==2.27.1
+requests==2.28.1
 service_identity==21.1.0
-setuptools==60.9.3
-twine==3.8.0
+setuptools==63.4.2
+twine==4.0.1
```
mxschmitt, line 21: Should be fine, they just dropped Python 3.6 support see [here](https://twine.readthedocs.io/en/stable/changelog.html#twine-4-0-0-2022-03-31).

rwoll:
Let's hold off on merging any dep changes until after 1.25 release. It looks like Lint is failing, but will be fixed by https://github.com/microsoft/playwright-python/pull/1483/files#diff-291261f4050ff453e2c661edf96c77af65a779b75815708d924809ee49d5dc64R96. 

However, will https://github.com/microsoft/playwright-python/pull/1483#issuecomment-1209993623 still be there?

scripts/documentation_provider.py
```diff
@@ -205,7 +205,8 @@ def print_entry(
 
     def print_events(self, class_name: str) -> None:
         clazz = self.classes[class_name]
-        if events := clazz["events"]:
+        events = clazz["events"]
+        if events:
```
rwoll, line 209: Why can we no longer use `:=` here? This code only runs in 3.8+. I cannot reproduce it locally, and in CI mypy complains, but this seems like a FP and I cannot reproduce locally.

It's unideal the the pre-commit run locally vs. CI produces different results. Please file an issue to sort that out as this change seems to be hiding a larger issue with our linting.

rwoll (commented):
Please re-base on main or give check the box for me to push changes to this PR (I tried re-basing myself, but GitHub denied the push to your PR).

scripts/documentation_provider.py
rwoll: We chatted offline. We're still not sure—and while not satisfying—since this seems to be a one off, this is good workaround for the lint FP for now. 

If more discrepancies arise, we can spend more time digging in!

rwoll (approved):
Assuming bots pass, LGTM. Thanks!

Issue #1504, opened 2022-08-16
[internal] move to `inlcude`, instead of `exclude` for setup.py


rwoll:
Avoid accidentally publishing unintended packages.

A potential follow up for 1.26 release since this has its own risks, too. It's independent of the outcome of https://github.com/microsoft/playwright-python/issues/1500.

PR #1510, opened 2022-08-17
chore(build): have allow-list in setup.py included packages

mxschmitt:
Before it was the following:

```
>>> find_packages(exclude=["tests*"])
['playwright', 'playwright._impl', 'playwright.async_api', 'playwright.sync_api', 'playwright._impl.__pyinstaller']
```

after reducing it to the shown in the PR, it was still fine with my local integration tests (build wheel, install from wheel).

Fixes https://github.com/microsoft/playwright-python/issues/1504.


rwoll (approved):
Provisional LGTM. Please diff site directory before and after this change to esnsure the whole distributed/installed file hierarchy is identical before v. after.

Issue #1395, opened 2022-07-01
Python 3.11 support


mxschmitt:
Blocked by https://github.com/python-greenlet/greenlet/issues/305

Python 3.11 is currently in development and going to be released on October 3, 2022.

kumaraditya303:
The upstream issue [^1] is not a blocker for playwright as it does not uses C API of `greenlet`. There are pre-release wheels available for 3.11 [^2] which works on 3.11+, all that is needed is to update the requirement to allow for `>=` in playwright.

[^1]: https://github.com/python-greenlet/greenlet/issues/305

[^2]: https://pypi.org/project/greenlet/2.0.0a2/#files

mxschmitt:
Good catch! But I think would be nice to wait for the official release of 2.0 instead of relying on pre-releasees.

PR #1533, opened 2022-08-25
chore: add support for Python 3.11

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1395

### DIFF
--- a/scripts/generate_api.py
+++ b/scripts/generate_api.py
@@ -13,17 +13,20 @@
 # limitations under the License.
 
 import re
+import sys
 from types import FunctionType
 from typing import (  # type: ignore
     Any,
+    Dict,
     List,
     Match,
+    Optional,
     Union,
     cast,
     get_args,
     get_origin,
-    get_type_hints,
 )
+from typing import get_type_hints as typing_get_type_hints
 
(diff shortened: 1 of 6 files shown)


scripts/generate_api.py
```diff
@@ -287,3 +290,34 @@ def return_value(value: Any) -> List[str]:
...
+
+# Python 3.11+ does not treat default args with None as Optional anymore, this wrapper will still wrap them.
+# https://github.com/python/cpython/issues/90353
+def get_type_hints(func: Any, globalns: Any) -> Dict[str, Any]:
+    original_value = typing_get_type_hints(func, globalns)
+    if sys.version_info < (3, 11):
+        return original_value
+    for key, value in _get_defaults(func).items():
+        if original_value[key] is not Optional:
+            original_value[key] = Optional[original_value[key]]
```
rwoll, line 302: Could this overly apply optional? I think there should be a check for `value`, too, before wrapping in optional? It generates no diff, so looks fine as is for our use case, but seems like this is no longer equivalent to the built-in typing_get_type_hints < 3.11

mxschmitt: you are right, added the additional if check as well.

PR #1568, opened 2022-09-27
Add release notes link to PyPI page

adamchainz:
Gives you a nice link in the left "Project links" section like on e.g. https://pypi.org/project/django-cors-headers/ . Will make it easier for those upgrading the package to find the release notes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -205,6 +205,9 @@ setup(
     long_description=Path("README.md").read_text(encoding="utf-8"),
     long_description_content_type="text/markdown",
     url="https://github.com/Microsoft/playwright-python",
+    project_urls={
+        "Release notes": "https://playwright.dev/python/docs/release-notes",
+    },
     packages=["playwright"],
     include_package_data=True,
     install_requires=[


setup.py
```diff
@@ -205,6 +205,9 @@ def _download_and_extract_local_driver(
     long_description=Path("README.md").read_text(encoding="utf-8"),
     long_description_content_type="text/markdown",
     url="https://github.com/Microsoft/playwright-python",
+    project_urls={
+        "Release notes": "https://playwright.dev/python/docs/release-notes",
```
mxschmitt, line 209:
```suggestion
        "Release notes": "https://github.com/microsoft/playwright-python/releases",
```

mxschmitt (approved):

setup.py
adamchainz: personally I'd prefer to read the release notes within the docs site, so they are formatted the same as the rest of the docs that they link to

mxschmitt: We were thinking about stopping mirroring the release-notes in the docs, since they are already on GitHub and having them in the docs is a manual effort. Or at least having them mirrored into playwright.dev.

adamchainz: okay, well you do you, you're free to push to my PR

PR #1576, opened 2022-10-05
chore: roll to Playwright ToT

pavelfeldman:


playwright/_impl/_selectors.py
```diff
@@ -46,6 +47,9 @@ async def register(
             await channel._channel.send("register", params)
         self._registrations.append(params)
 
+    def set_test_id_attribute(self, attribute_name: str) -> None:
+        set_test_id_attribute_name(attribute_name)
```
mxschmitt, line 51:
```suggestion
        Locator.set_test_id_attribute_name(attribute_name)
```

and inside Locators class:

```python
    @staticmethod
    def set_test_id_attribute_name(date):
        ...
```

pavelfeldman: I'll align from the other side!

Issue #1584, opened 2022-10-09
[BUG] `wait_for_function()` ignores `polling` parameter


johnhawkinson:
**Context:**
- Playwright Version: 1.27.0
- Operating System: Mac
- Python version: 3.8
- Browser: All

**Describe the bug**

Playwright's python bindings appear to ignore the `polling` parameter (or more properly, fail to pass it properly to the playwright driver). At least, assuming I am properly passing it, which I think I am, although I have not found examples of its use, so perhaps I'm missing something obvious.

For instance, this code should produce an error, because `polling` must be either a floating point number (milliseconds) or the string `raf`:

```py
from playwright.sync_api import sync_playwright
import time

with sync_playwright() as p:
    browser = p.chromium.launch(headless=False)
    page = browser.new_page()
    page.goto("http://playwright.dev")
    jsh = page.wait_for_function("""
function() { 
  console.log("poll", Date.now());
  return document.querySelector('foobar');
}
        """, polling="bogus")
    print(page.title())
    browser.close()
```

**Compare with Javascript bindings**

For instance, the equivalent Node.js script gives the proper error:

```js
const playwright = require('playwright');

(async () => {
  // Setup
  const browser = await playwright.chromium.launch({headless: false});
  const page = await browser.newPage();

  await page.goto('https://playwright.dev');
  const watchDog = page.waitForFunction(() => { 
    console.log("poll", Date.now());
    return document.querySelector('foobar');
  }, null, { polling: "bogus" } );

})()
```

which properly returns

```sh
jhawk@lrr wff3b_test % node wff3b.js     
/Users/jhawk/src/wff3b_test/node_modules/playwright-core/lib/utils/index.js:100
  if (!value) throw new Error(message || 'Assertion error');
                    ^

Error: Unknown polling option: bogus
    at assert (/Users/jhawk/src/wff3b_test/node_modules/playwright-core/lib/utils/index.js:100:21)
    at Frame.waitForFunction (/Users/jhawk/src/wff3b_test/node_modules/playwright-core/lib/client/frame.js:582:64)
    at Page.waitForFunction (/Users/jhawk/src/wff3b_test/node_modules/playwright-core/lib/client/page.js:760:28)
    at /Users/jhawk/src/wff3b_test/wff3b.js:9:25

Node.js v18.10.0
```

**An alternative example**

For a more useful example, consider this:

```py
from playwright.sync_api import sync_playwright
import time

with sync_playwright() as p:
[...]

PR #1586, opened 2022-10-10
fix: wait_for_function with polling interval

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1584

### DIFF
--- a/playwright/_impl/_frame.py
+++ b/playwright/_impl/_frame.py
@@ -746,8 +746,12 @@ class Frame(ChannelOwner):
         timeout: float = None,
         polling: Union[float, Literal["raf"]] = None,
     ) -> JSHandle:
+        if isinstance(polling, str) and polling != "raf":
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

johnhawkinson:
I must admit I'm a little confused; initially I thought that this parameter checking occurred in the playwright driver, but I see now I was looking at the Javascript client (`playwright-core/src/client/frame.ts`) rather than the driver server (`playwright-core/src/server/frames.ts`?). If that's even the right terminology.

It's a bit disconcerting to realize there does not seem to be any mechanism, generally speaking, to ensure that parameters are consistently handled across different frontends. It suggests to me there ought to be some kind of automated testing framework that generates parameters from a spec (maybe being the server spec or the javascript frontend?) and validates that all those parameters exist in all of the frontends. Sounds like a huge task, maybe impractical, I dunno.

Anyhow, thank you for this fix.

Issue #1602, opened 2022-10-21
[BUG] CDPSession.on() causes memory leak


ViktorRubenko:
**Context:**
- Playwright Version: 1.27.0
- Operating System: Windows 7 / Ubuntu 22.04
- Python version: 3.9 / 3.10
- Browser: Chromium

```
import asyncio
from playwright.async_api import async_playwright

async def handle_event(event):
    event["response"]["url"]

async def test():
    with open("urls") as f:
        urls = f.read().splitlines()
    for url in urls:
        async with async_playwright() as pw:
            browser = await pw.chromium.launch(headless=False)
            context = await browser.new_context(
                accept_downloads=False, ignore_https_errors=True
            )
            page = await context.new_page()

            client = await page.context.new_cdp_session(page)
            await client.send("Network.enable")
            client.on(
                "Network.responseReceived",
                lambda event: asyncio.ensure_future(handle_event(event)),
            )

            page.set_default_timeout(10000)
            try:
                await page.goto(url)
            except Exception as e:
                print(e)
            finally:
                await page.close()
            await context.close()
            await browser.close()


asyncio.run(test())
```

By tracemalloc after each iteration memory usage increase for:
```
#1: usr/lib/python3.10/json/decoder.py:353: 304.2 MB
    obj, end = self.scan_once(s, idx)
#2: python3.10/site-packages/playwright/_impl/_connection.py:381: 99.6 MB
    result[key] = self._replace_guids_with_channels(value)
#3: python3.10/site-packages/playwright/_impl/_impl_to_api_mapping.py:51: 87.7 MB
    o[name] = self.from_maybe_impl(value, visited)
```

Everything goes fine without CDPSession.on()
But memory usage starts accumulating with it

mxschmitt:
Thanks for the great bug report! Fix will be available in the next release.

PR #1604, opened 2022-10-26
chore: fix impl to api name memory leak

mxschmitt:
fixes https://github.com/microsoft/playwright-python/issues/1602

### DIFF
--- a/tests/test_reference_count_async.py
+++ b/tests/test_reference_count_async.py
@@ -36,17 +36,18 @@ async def test_memory_objects(server: Server, browser_name: str) -> None:
 
         await page.route("**/*", lambda route, _: route.fulfill(body="OK"))
 
-        # https://github.com/microsoft/playwright-python/issues/1602
-        client = await page.context.new_cdp_session(page)
-        await client.send("Network.enable")
-
         def handle_network_response_received(event: Any) -> None:
             event["__pw__is_last_network_response_received_event"] = True
 
-        client.on(
-            "Network.responseReceived",
-            handle_network_response_received,
-        )
+        if browser_name == "chromium":
+            # https://github.com/microsoft/playwright-python/issues/1602
+            client = await page.context.new_cdp_session(page)
+            await client.send("Network.enable")
+
+            client.on(
+                "Network.responseReceived",
+                handle_network_response_received,
+            )
 
         for _ in range(100):
             response = await page.evaluate("""async () => (await fetch("/")).text()""")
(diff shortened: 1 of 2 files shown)


tests/test_reference_count_async.py
```diff
@@ -44,12 +57,17 @@ async def test_memory_objects(server: Server, browser_name: str) -> None:
     gc.collect()
 
     pw_objects: defaultdict = defaultdict(int)
+    found_cdp_network_events = False
     for o in objgraph.by_type("dict"):
         name = o.get("_type")
+        if o.get("__pw__is_last_network_response_received_event", False):
+            found_cdp_network_events = True
```
mxschmitt, line 64: maybe move assert there

pavelfeldman (approved):

nay-kang:
I had face this memory leak for a week. thank you for fixing it.
but I'm confusing with this code:
* there is seems no call to `from_maybe_impl` function with `visited` argument
* it has no benift to create a `visited` inside this function. in the previous version the shared `visisted` argument could cache obj between calls to speed up but will cause memory leak.

so I guess if we can remove the `visited` argument. or using an instance attribute belongs to `BrowserContext` then we can close `BrowserContext` to relase memory.

and there has another code could cause memory leak `playwright/_impl/_js_handle.py`
```python
@dataclass
class VisitorInfo:
    visited: Map[Any, int] = Map()
    last_id: int = 0

    def visit(self, obj: Any) -> int:
        assert obj not in self.visited
        self.last_id += 1
        self.visited[obj] = self.last_id
        return self.last_id
```
the `visited` attribute in the `VisitorInfo` is a class attribute, which will share between instance and will not release memory until reload the module.

all this code are belongs to the same commit:`ed13a53281bc213e1a4341c50d47fbc65670288f`

mxschmitt:
I see one, see [here](https://github.com/microsoft/playwright-python/blob/8ecd961bdb532dc658151f1fa2c9ecc9b062182c/playwright/_impl/_impl_to_api_mapping.py#L55).

Will create a follow-up with this potential VisitorInfo leak, thanks for pointing this out.

nay-kang:
hi @mxschmitt ,you memtioned the `from_maybe_impl` function with `visited` argument call [here](https://github.com/microsoft/playwright-python/blob/8ecd961bdb532dc658151f1fa2c9ecc9b062182c/playwright/_impl/_impl_to_api_mapping.py#L55) 
```python
o[name] = self.from_maybe_impl(value, visited)
```  
 is a recursive call to himself, so I think it could run well without `visisted` argument

PR #1611, opened 2022-10-28
Use SPDX license expression in project metadata

RazerM:
As a downstream user it is desirable to be able to programmatically determine the precise licenses used by our dependencies. The emerging convention for this is the [SPDX License List](https://spdx.org/licenses/). SPDX License Expressions are already used in the JavaScript (npm) and Rust (crates.io) ecosystems, for example.

In Python, there is [PEP 639](https://peps.python.org/pep-0639/) which would add a standard 'License-Expression' metadata field. The discussion around this PEP seems to have gone stale in 2021.

This merge request is in lieu of that PEP coming soon. Unless you as project maintainers think the proposed change in this PR has downsides (which you are entitled to!), I would ask that you accept this change so that it's possible for users to attempt to parse your license metadata as an SPDX License Expression.

If PEP 639 is accepted, this package would be ready for the change just by changing the `license` key to `license_expression` or whatever key the PEP lands on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -204,6 +204,7 @@ setup(
     description="A high-level API to automate web browsers",
     long_description=Path("README.md").read_text(encoding="utf-8"),
     long_description_content_type="text/markdown",
+    license="Apache-2.0",
     url="https://github.com/Microsoft/playwright-python",
     project_urls={
         "Release notes": "https://github.com/microsoft/playwright-python/releases",


mxschmitt:
Thank you for your Pull Request, I'd prefer waiting until the PEP is accepted. Closing in the meantime.

RazerM:
@mxschmitt To be clear, it's the `license_expression` field and constraining it to an SPDX License Expression that's part of the PEP.

The `license` field I've used is a [standard field of core metadata](https://packaging.python.org/en/latest/specifications/core-metadata/#license) which is free form; I'd ask that you please put something there. The reasoning is that the [classifier used](https://github.com/microsoft/playwright-python/blob/51e416b9cee7ea2789ddfc2e8bdc8657ef5a2954/setup.py#L228) (`License :: OSI Approved :: Apache Software License`) is ambiguous, as it does not specify whether it is Apache License 1.0, 1.1, or 2.0.

Usually packages are already using the `license` field but with something ambiguous like "Apache" and I've had several projects agree to use an unambigous expression like "Apache-2.0".

If you don't with to use the SPDX License Identifer `Apache-2.0`, could you use `Apache License 2.0` instead?

The license field of the TypeScript package's [`package.json`](https://github.com/microsoft/playwright/blob/445404abb7057e9e0a3a54c31a501027a9ed1d1e/package.json#L14) is already `Apache-2.0` as well.

mxschmitt:
Sounds good, I looked at some other popular projects and they also have them.

RazerM:
Thank you!

Issue #1534, opened 2022-08-25
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/622c73cc1e677834ebdc27e95c89ff101bdccf54 (feat: expect(locator).toHaveAttribute to assert attribute presence (#16767))
  - [x] https://github.com/microsoft/playwright/commit/a07a4a25a26e2fa6a976fb2ff174f25b24414da6 (chore: make parent scope explicit (#16819))
  - [x] https://github.com/microsoft/playwright/commit/f30ac1d67827fbcf3f8be7a851ef81b761842c52 (fix: support toHaveAttribute(name, options) (#16941))
  - [x] https://github.com/microsoft/playwright/commit/306ab34aa3bcdfbb2369e1c698a161565bacc275 (feat(assertions): support toBeEnabled({ enabled }) (#17058))
  - [x] https://github.com/microsoft/playwright/commit/f0c581060959bf680d2165f38e29743388d03b21 (feat(assertions): support toBeEditable({ editable }) (#17065))
  - [x] https://github.com/microsoft/playwright/commit/bca13bc35aead0122b0609ea18a0769864616c29 (feat(assertions): support toBeVisible({ visible }) (#17207))
  - [x] https://github.com/microsoft/playwright/commit/17b203affbc373fd5c7d8e643285e1262350cc67 (feat: added follow and redirect arguments to fetch (#17033))
  - [x] https://github.com/microsoft/playwright/commit/c91df61ca33c133046425d40b45dfc25ec331673 (chore: optionally connect instead of launching (#17174))
  - [x] https://github.com/microsoft/playwright/commit/8d639ae50e123296fb7ec737d5e1104a5ffb9ec1 (chore: revert toHaveAttribute type sig and overloads (#17406))
  - [x] unskip wk by searching for https://bugs.webkit.org/show_bug.cgi?id=225281
  - [x] https://github.com/microsoft/playwright/commit/43304e980d1164c55cca8b871fd76546a133013e (feat: show browser.close() stack in "Browser has been closed" error (#17376))
  - [x] https://github.com/microsoft/playwright/commit/df143031e7eac11460bcd658dc3bb5a8951ccd64 (chore: move protocol and trace types into the top-level packages (#17486))
[...]

PR #1627, opened 2022-11-02
chore(roll): roll Playwright to 1.28.0-alpha-nov-2-2022

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1534


tests/server.py
```diff
@@ -209,7 +209,8 @@ def handle_redirect(request: http.Request) -> None:
 
 class HTTPServer(Server):
     def listen(self, factory: ClientFactory) -> None:
-        reactor.listenTCP(self.PORT, factory)
+        reactor.listenTCP(self.PORT, factory, interface="127.0.0.1")
```
mxschmitt, line 212: I think we should not do that and try to avoid that instead.

yury-s (approved):

PR #1707, opened 2023-01-03
chore: roll to Playwright version 1.29.1-beta-1672665986000

### LINKED ISSUE #1699: PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS=1 warns to stdout and breaks transport protocol
**Context:**
- Playwright Version: 1.29.0

**Describe the bug**

playwright writes to stdout when `PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS=1` is set https://github.com/microsoft/playwright/blob/7508c574e12e31c774668b717f3c2449a82565b1/packages/playwright-core/src/server/registry/index.ts#L657
and that breaks transport protocol that expects 4 bytes length + json on stdout https://github.com/microsoft/playwright-python/blob/eb4b88af72aa6757c95f5a74f11ad6503bf16038/playwright/_impl/_transport.py#L140-L160

rwoll:
Fixes #1699.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ except ImportError:
     InWheel = None
 from wheel.bdist_wheel import bdist_wheel as BDistWheelCommand
 
-driver_version = "1.29.1"
+driver_version = "1.29.1-beta-1672665986000"
 
 
 def extractall(zip: zipfile.ZipFile, path: str) -> None:


mxschmitt:
its not a regression so we don't cherry-pick.

PR #1724, opened 2023-01-15
Allow entry point to be called from a script

StevenClontz:
Adds arguments to `main()` to support being called directly by a script.

### DIFF
--- a/playwright/__main__.py
+++ b/playwright/__main__.py
@@ -18,10 +18,12 @@ import sys
 from playwright._impl._driver import compute_driver_executable, get_driver_env
 
 
-def main() -> None:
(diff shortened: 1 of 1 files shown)


StevenClontz:
We're using playwright-python to take screenshots in our [application](https://github.com/PreTeXtBook/pretext-cli), and the intent is to allow us to do the `playwright install` step for the user from the library, rather than using than calling the entry point via a subprocess. Would this be appropriate on the playwright-python developers' end?

mxschmitt:
Sorry for the late response, I recommend to start with a feature request first, where we can discuss API design, wait for other customers who encounter a similar problem etc.

Thanks for your understanding!

PR #1726, opened 2023-01-16
fix: include-hidden arg in get_by_role_selector

lewinkaestner:
The `get_by_role()` locator currently throws an error (`Unknown attribute “include-hiddenen“`) when using the include_hidden attribute (e.g. `page.get_by_role("radio", include_hidden=True)`).

I fixed the typo which leads to this error.


mxschmitt (approved):
thanks

Issue #1745, opened 2023-01-30
[BUG] Can't install Playwright with Conda on M1 Mac


mburki32:
**Context:**
- Playwright Version: `v1.30`
- Operating System: `MacOS Ventura 13.1`
- Python version: `3.9`
- Extra: `conda 22.11.1`

**Code Snippet**
```
% conda create -n test_playwright python=3.9

% conda activate test_playwright

% conda install -c microsoft playwright
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: - 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                          

UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions
```

mxschmitt:
Looks like it found conflicts but it's not able to list them. I'm not a conda expert, so would appreciate if you could dig a bit - maybe some log files somewhere will give us more hints or enabling a verbose mode.

mxschmitt:
I tried it on my M1 mac with miniconda Intel where it was working without any issues. If you have miniconda for Apple M1 Silicon Chip, then its indeed not working.

mburki32:
> I tried it on my M1 mac with miniconda Intel where it was working without any issues. If you have miniconda for Apple M1 Silicon Chip, then its indeed not working.

I do have a Apple M1 pro Silicon Chip...

what are the workarounds for this (if any)?

mxschmitt:
You can install miniconda Intel instead as a workaround. Maybe you can also override the arch which conda is installing.

mburki32:
is direct support for M1 Silicon Chip in the works? when will it be available?

In the meantime I will try what @mxschmitt suggested. I found the following 2 resources that might help me with the workaround:
* [How to install Miniconda x86_64 Apple M1 side-by-side on Mac Book M1](https://towardsdatascience.com/how-to-install-miniconda-x86-64-apple-m1-side-by-side-on-mac-book-m1-a476936bfaf0)
[...]

PR #1821, opened 2023-03-21
add osx-arm64 variant config file

wolfv:
I didn't have a chance to test this yet -- any idea on how to test this easily?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -46,3 +46,10 @@ jobs:
         run: |
           conda config --set anaconda_upload yes
           conda build --user microsoft .
+      - name: Build and Upload macOS Apple Silicon
+        env:
+          ANACONDA_API_TOKEN: ${{ secrets.ANACONDA_API_TOKEN }}
+        if: matrix.os == 'macos-latest'
+        run: |
+          conda config --set anaconda_upload yes
+          conda build --user microsoft . --variant-config-files ./conda_build_config_osx_arm64.yaml
--- a/conda_build_config_osx_arm64.yaml
+++ b/conda_build_config_osx_arm64.yaml
@@ -1,0 +1,5 @@
+python:
+  - 3.9
+  - "3.10"
+  - "3.11"
+target_platform: osx-arm64


mxschmitt:
I think no way to test it easily, but we release a new Playwright version shortly, so we can then see if it works!

mxschmitt:
We have build workflows configured, you. could try to add it there as well: https://github.com/microsoft/playwright-python/blob/890af94d3380c6dfbe9fabeaedea6b92aa232b11/.github/workflows/ci.yml#L188-L207

Can you actually publish m1 from a amd mac?

wolfv:
Does the `pip setup.py` download any binaries from the internet? We might have to tell it to grab the appropriate ones for `osx-arm64` as otherwise it migth think that it's still on `osx-64`.

All the `osx-arm64` packages on conda-forge are built by cross-compiling so it should work fine from the "conda-side".

wolfv:
Just FYI cross compiling with conda-build works by putting the dependencies for the build platforms (osx-64) that need to run _during_ build in the `build` section of the recipe, and the ones that are linked against in the `host` section (that section will use packages from the `osx-arm64` architecture.

mxschmitt:
> Does the pip setup.py download any binaries from the internet?

Yes it does.

By default it builds wheels containing the driver for the current platform/architecture. See [here](https://github.com/microsoft/playwright-python/blob/890af94d3380c6dfbe9fabeaedea6b92aa232b11/setup.py#L126-L135). You can pass `--all` and then it builds all the wheels. I think as of today the issue is that it builds only the current platform/arch by default, so we require conda to run on this specific arch to be able to get the corresponding wheel while publishing the conda package. 

> We might have to tell it to grab the appropriate ones for osx-arm64 as otherwise it migth think that it's still on osx-64.

Yes I think thats what we need to do. And then tell conda to invoke setup.py per platform.

mxschmitt:
Closing as per above. If anyone wants to revisit this in the future, feel free to create a new PR, thanks!

mxschmitt:
Actually let me re-open, things have changed, GitHub now has M1 runners, do you mind adding this to the matrix? Then it should work I believe: https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

mxschmitt:
Landed in https://github.com/microsoft/playwright-python/issues/1745.

mxschmitt:
This should be fixed as part of v1.43.0

wolfv:
Thanks!

Issue #1777, opened 2023-02-21
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/d10fac4f6a0dfeb3577eec6cdff7ec3bdc843c01 (feat: Locator.or(locator) (#21884))
  - [x] https://github.com/microsoft/playwright/commit/d4e0ef7f1abf7ebe6d806bfee2efa46f6959aee7 (chore: explicitly send route's requestUrl for tracing (#21940))
  - [x] https://github.com/microsoft/playwright/commit/525097d465414c44a7a0b3f307ff532bac7768d5 (feat: Locator.filter(locator) (#21975))
  - [x] https://github.com/microsoft/playwright/commit/f502c72f2b688ba72b38d456523c47deedd3df55 (feat: browserContext.on('console') (#21943))
  - [x] https://github.com/microsoft/playwright/commit/3b359e27b71c9d47c8c0843d09ef9fb1a36132a0 (feat: BrowserContext.on('dialog') (#22033))
  - [x] https://github.com/microsoft/playwright/commit/6929214dbf895cd961175c17d1616271f7fc72b3 (feat(expect): expect(locator).toBeAttached() (#22067))
  - [x] https://github.com/microsoft/playwright/commit/e6148bb725db6648d033ade239f485274849ec63 (feat: Locator.not(locator) (#22066))
  - [x] https://github.com/microsoft/playwright/commit/539d9873c2f3235928106751c2301d91b3b6d628 (feat: rename Locator.filter(locator) to Locator.and (#22101))
  - [x] https://github.com/microsoft/playwright/commit/37d165950809d37e0a3fec48034fd239b577b3e8 (feat(connect): support special headers for debug/attachments (#22106))
  - [x] https://github.com/microsoft/playwright/commit/5223c1ba39114dec5b0baf9ec17800cea612f9ce (feat: add timeout to `electron.firstWindow()` method (#21863))
  - [x] https://github.com/microsoft/playwright/commit/0d9ec60dc7e52bc3f77e4f032da3cd30216b2ef7 (chore: implement jsHandle._objectCount in chromium (#22127))
  - [x] https://github.com/microsoft/playwright/commit/d7b3836752fbb996a4eb7dd6e36c3970912ac2e4 (chore: revert console and dialog events on BrowserContext (#22195))
[...]

PR #1859, opened 2023-04-12
chore(roll): roll Playwright to 1.33.0-alpha-apr-12-2023

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1777

### DIFF
--- a/playwright/_impl/_api_structures.py
+++ b/playwright/_impl/_api_structures.py
@@ -64,9 +64,10 @@ class Geolocation(TypedDict, total=False):
     accuracy: Optional[float]
 
 
-class HttpCredentials(TypedDict):
+class HttpCredentials(TypedDict, total=False):
     username: str
     password: str
+    origin: Optional[str]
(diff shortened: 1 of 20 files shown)


playwright/_impl/_api_structures.py
```diff
@@ -64,9 +64,10 @@ class Geolocation(TypedDict, total=False):
     accuracy: Optional[float]
 
 
-class HttpCredentials(TypedDict):
+class HttpCredentials(TypedDict, total=False):
```
yury-s, line 67: Why total=false, isn't it enough that origin is optional?

yury-s (approved):

playwright/_impl/_api_structures.py
mxschmitt: total means that all the properties need to be specified, by default all need to be specified, Optional doesn't help there unfortunately, we used that in the past for it too.

PR #1882, opened 2023-04-27
(🎁) Add `|` operator on `Locator`s

KotlinIsland:
umm, based python moment????

### DIFF
--- a/playwright/_impl/_locator.py
+++ b/playwright/_impl/_locator.py
@@ -105,6 +105,9 @@ class Locator:
     def __repr__(self) -> str:
         return f"<Locator frame={self._frame!r} selector={self._selector!r}>"
 
+    def __or__(self, other: "Locator") -> "Locator":
+        return self.or_(other)
+
     async def _with_element(
         self,
         task: Callable[[ElementHandle, float], Awaitable[T]],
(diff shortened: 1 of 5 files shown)


KotlinIsland:

mxschmitt:
I guess the same makes sense for and? Will discuss with the team later!

KotlinIsland:
@mxschmitt Do you mean and as in an operator for `filter`?

mxschmitt:
(Never mind, `Locator.and(...)` was something we were experimenting with and I got confused, we didn't ship it)

mxschmitt (approved):

mxschmitt:
We decided for now to keep _or instead of | since or_ also provides auto completion support and shows what it is actually doing. Thanks for you suggestion tho!

KotlinIsland:
# Workaround

```py
from playwright.sync_api import Locator

Locator.__or__ = Locator.or_
```

DetachHead:
@mxschmitt it seems silly to not accept this PR when the alternative is to append an underscore to the method name as a workaround for the fact that "or" causes a syntax error.

```py
page.locator('div') | page.locator('span')
```
looks much less gross than
```py
page.locator('div').or_(page.locator('span'))
```

python specifically supports this use case, why not take advantage of it?

Issue #1883, opened 2023-04-27
[BUG] page.expect_response not respecting timeout setting in browser_context.set_default_timeout


loneguardian:
**Context:**
- Playwright Version: 1.32.1
- Operating System: Windows 10
- Python version: 3.11.2
- Browser: Chromium, Chrome, Firefox

**Code Snippet**
```python
import asyncio
from playwright.async_api import async_playwright
from playwright.sync_api import sync_playwright

async def main_async():
    async with async_playwright() as pw:
        browser = await pw.firefox.launch()
        context = await browser.new_context()
        context.set_default_timeout(10000)
        page = await context.new_page()
        #page.set_default_timeout(10000)

        try:
            async with page.expect_response("https://somethingelse.com"):
                await page.goto("https://www.yahoo.com")
        finally:
            context.close()
            browser.close()

def main_sync():
    with sync_playwright() as pw:
        browser = pw.chromium.launch()
        context = browser.new_context()
        context.set_default_timeout(10000)
        page = context.new_page()
        #page.set_default_timeout(10000)

        try:
            with page.expect_response("https://somethingelse.com"):
                page.goto("https://www.yahoo.com")
        finally:
            context.close()
            browser.close()

if __name__ == "__main__":
    asyncio.run(main_async())
    #main_sync()
```

**Describe the bug**
- Run the code snippet
- `page.expect_response` timeout at `30000ms` (default)

**What should happen**
- According to documentation of `page.expect_response` [here](https://playwright.dev/python/docs/api/class-page#page-wait-for-response), `browser_context.set_default_timeout()` method should affect the timeout, hence the snippet should timeout at `10000ms`

**Add any other details about the problem here.**
- `page.set_default_timeout()` works correctly
- Same behaviour observed in sync version of the API
- Same behaviour observed for the following browsers: Chromium, Chrome and Firefox
- I am not sure of any other page methods that does not conform to `browser_context`'s timeout setting.

PR #1889, opened 2023-05-02
fix: inherit context timeout correctly

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1883


playwright/_impl/_helper.py
```diff
@@ -177,8 +177,8 @@ class HarLookupResult(TypedDict, total=False):
 class TimeoutSettings:
     def __init__(self, parent: Optional["TimeoutSettings"]) -> None:
         self._parent = parent
-        self._timeout = 30000.0
-        self._navigation_timeout = 30000.0
+        self._timeout: Optional[float] = None
```
yury-s, line 180: I guess this is unrelated to the actual fix

Issue #1903, opened 2023-05-07
[Question]: "The reaction to the self-closing of a window using JavaScript."


skn78:
### Your question

JavaScript closes the window using the close() call.
I'm trying to determine the moment when a window is closed. page.is_closed() doesn't work, it shows False both before and after the window is closed.
I found a way to determine this using the following code:
```
with new_page.expect_event("close") as event_info:
          pass
```
This code snippet waits for the window to close on its own, but during execution, a message is displayed on the screen:
```
Future exception was never retrieved
future: <Future finished exception=Error('Target page, context or browser has been closed')>
playwright._impl._api_types.Error: Target page, context or browser has been closed

```
What is the correct way to track window closing? How can I avoid the error messages in my code?

PR #1905, opened 2023-05-09
chore: do not continue with no_reply messages

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1903

Tricky to test that a warning does not get displayed. Before the following:

```python

    browser = playwright.chromium.launch()
    page = browser.new_page()
    page.goto('http://example.com/')
    with page.expect_popup() as popup_info:
        page.evaluate('window.open("http://example.com/")')
    popup = popup_info.value
    with popup.expect_event('close') as popup_info:
        popup.evaluate('window.close()')
    browser.close()
```

resulted in:

```
(env) ➜  playwright-python git:(main) ✗ python test.py
Future exception was never retrieved
future: <Future finished exception=Error('Target page, context or browser has been closed')>
playwright._impl._api_types.Error: Target page, context or browser has been closed
```

because the "after" `"waitForEventInfo"` call was returning an exception and in Python you need to await all the exceptions.

Ways thought about fixing it:
a) never add it to the self._callbacks
b) add a special no_reply flag to the callbacks - did this for now.

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -68,8 +68,7 @@ class Channel(AsyncIOEventEmitter):
         )
 
     def send_no_reply(self, method: str, params: Dict = None) -> None:
-        # No reply messages are used to e.g. waitForEventInfo(after) which returns exceptions on page close.
-        # To prevent 'Future exception was never retrieved' we just ignore such messages.
+        # No reply messages are used to e.g. waitForEventInfo(after).
         self._connection.wrap_api_call_sync(
             lambda: self._connection._send_message_to_server(
                 self._guid, method, {} if params is None else params, True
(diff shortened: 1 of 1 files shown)


playwright/_impl/_connection.py
```diff
@@ -357,6 +359,8 @@ def dispatch(self, msg: ParsedMessagePayload) -> None:
             callback = self._callbacks.pop(id)
             if callback.future.cancelled():
                 return
+            if callback.no_reply:
```
pavelfeldman, line 362: What was bad about calling set_result on the callback future?

mxschmitt: See the PR description, that we received an error (waitForEventInfo after), were setting it on the protocol callback but never awaiting it, hence it was showing the warning in the console.

pavelfeldman (approved):

PR #1906, opened 2023-05-09
chore: follow-ups for #1897 connection changes

mxschmitt:


playwright/_impl/_json_pipe.py
```diff
@@ -48,18 +46,15 @@ async def wait_until_stopped(self) -> None:
     async def connect(self) -> None:
         self._stopped_future: asyncio.Future = asyncio.Future()
 
-        close_error: Optional[str] = None
+        def handle_closed(close_error: str = None) -> None:
+            self.emit("close", close_error)
```
pavelfeldman, line 50: I think you are emitting multiple close events now.

playwright/_impl/_json_pipe.py
```diff
@@ -48,18 +46,15 @@ async def wait_until_stopped(self) -> None:
...
             try:
                 self.on_message(cast(ParsedMessagePayload, message))
             except Exception as e:
-                nonlocal close_error
-                close_error = str(e)
-
-        def handle_closed() -> None:
-            self.emit("close", close_error)
-            self._stopped_future.set_result(None)
+                handle_closed(str(e))
```
pavelfeldman, line 57: I wonder if you want to close the pipe instead?

playwright/_impl/_json_pipe.py
mxschmitt: Verified that we only emit it a single time. I can protect it if you want that we do it only a single time for safety.

playwright/_impl/_json_pipe.py
mxschmitt: we can just remove this code if you want, feel like its not covered by test anyways.

pavelfeldman (approved):

pavelfeldman (approved):

Issue #1843, opened 2023-04-03
[BUG] context.close() hangs when using together with signal


chenstrace:
### System info
- Playwright Version: [v1.27.0]
- Operating System: [All]
- Browser: [All, Chromium, Firefox, WebKit]
- Other info:

### Source code

```python
import signal
import time

from playwright.sync_api import sync_playwright

def my_sig_handler(signum, frame):
    global exit
    exit = True


exit = False
playwright = sync_playwright().start()
browser = playwright.chromium.launch(headless=False)
context = browser.new_context(record_har_path="/Users/chenstrace/example.har")
page = context.new_page()
page.goto("https://www.baidu.com/")
signal.signal(signal.SIGINT, my_sig_handler)

try:
    while True:
        if exit:
            break
        else:
            print("sleep 1", flush=True)
            time.sleep(1)
finally:
    print("close context", flush=True)
    context.close()
    print("close browser", flush=True)
    browser.close()
```

**Steps**
- save the code to file a.py
- run the program in terminal, using  ```python3 a.py```
- wait for 5 seconds
- press CTRL+C in terminal.

**Expected**
- python program can exit
- a har file was saved to /Users/chenstrace/example.har

**Actual**
- the program hangs on context.close()
- no har file was saved.

mxschmitt:
I can reproduce.

mxschmitt:
The issue is that if you press Command + C, the driver gets killed and then https://github.com/microsoft/playwright-python/blob/b913f4830f065883419e6840e3f5be089c47cb08/playwright/_impl/_browser_context.py#L448 will stall since we don't cancel the callbacks when the driver connection gets closed.

mxschmitt:
This will be fixed with 1.34, thanks for filing the bug!

B-2U:
hello there, I just tested the same code in [playwright==1.34.0]
but its still happened
the program hangs on context.close()

PR #1908, opened 2023-05-09
test: add test for Command + C does not terminate driver connection

mxschmitt:
Issue what this solves is that by default if a user presses Command + C it will send it to the `main.py` script (good) but also to the `driver` (bad) which will instantly close the browsers etc. This is because the driver is getting launched as the same process group, we should launch it as a separate process group imo.

See https://unix.stackexchange.com/a/149756

Since upstream we can do the following:

```ts
(async () => {
  const browser = await chromium.launch({ handleSIGINT: false, handleSIGHUP: false, handleSIGTERM: false, headless: false });
  const context = await browser.newContext();
  const page = await context.newPage();

  let exited = false;

  process.on('SIGINT', async () => {
    exited = true;
    console.log('SIGINT signal received, graceful shutdown!');
    await context.close();
    await browser.close();
  });
  console.log(`Process Command + C to exit`);
  while (!exited)
    await page.waitForTimeout(1000);

})();
```

we should be able to do the same in Python. It should be an implementation detail that we are launching a subprocess internally.

https://github.com/microsoft/playwright-python/issues/1843

### DIFF
--- a/tests/common/test_signals.py
+++ b/tests/common/test_signals.py
@@ -14,6 +14,7 @@ from playwright.sync_api import sync_playwright
 def _test_signals_async(
     browser_name: str, launch_arguments: Dict, wait_queue: "multiprocessing.Queue[str]"
 ) -> None:
+    os.setpgrp()
     sigint_received = False
 
     def my_sig_handler(signum: int, frame: Any) -> None:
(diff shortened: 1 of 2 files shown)


tests/common/test_signals.py
```diff
@@ -0,0 +1,121 @@
...
+def _create_signals_test(
+    target: Any, browser_name: str, launch_arguments: Dict
+) -> None:
+    wait_queue: "multiprocessing.Queue[str]" = multiprocessing.Queue()
+    process = multiprocessing.Process(
+        target=target, args=[browser_name, launch_arguments, wait_queue]
+    )
+    process.start()
+    assert process.pid is not None
+    wait_queue.get()
```
yury-s, line 99: nit: add this in the logs too?

yury-s (approved):

Issue #1910, opened 2023-05-10
[BUG] playwright._impl._api_types.Error: headers: expected array, got object


blv-matheuslins:
**Context:**
- Playwright Version: `v1.33.0`
- Operating System: `Linux`
- Python version: `3.10`
- Browser: `Chromium`

**Code Snippet**

```python
if __name__ == "__main__":
    import asyncio
    from playwright.async_api import async_playwright


    async def main():
        async with async_playwright() as p:
                headers = {'User-Agent': "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/88.0.4324.96 Chrome/88.0.4324.96 Safari/537.36"}
                browser = await p.chromium.connect_over_cdp("http://whatsmyuseragent.org/", headers=headers)
                page = await browser.new_page()
                await page.goto('http://whatsmyuseragent.org/')
                await browser.close()


    asyncio.run(main())
```

**Describe the bug**

I'm facing this error with a simple script only trying to create a connection with cdp.

**Traceback**
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 646, in run_until_complete
    return future.result()
  File ".config/JetBrains/PyCharmCE2023.1/scratches/playwright_connection.py", line 11, in main
    browser = await p.chromium.connect_over_cdp("http://whatsmyuseragent.org/", headers=headers)
  File "/venv/lib/python3.10/site-packages/playwright/async_api/_generated.py", line 14829, in connect_over_cdp
    await self._impl_obj.connect_over_cdp(
  File "/venv/lib/python3.10/site-packages/playwright/_impl/_browser_type.py", line 168, in connect_over_cdp
    response = await self._channel.send_return_as_dict("connectOverCDP", params)
  File "/venv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 66, in send_return_as_dict
    return await self._connection.wrap_api_call(
  File "/venv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 461, in wrap_api_call
    return await cb()
  File "/venv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 96, in inner_send
    result = next(iter(done)).result()
playwright._impl._api_types.Error: headers: expected array, got object

Process finished with exit code 1
```

[...]

PR #1912, opened 2023-05-11
fix: make it possible to send connect_over_cdp headers

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1910

### DIFF
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -44,6 +44,7 @@ from playwright._impl._helper import (
     locals_to_params,
 )
 from playwright._impl._json_pipe import JsonPipeTransport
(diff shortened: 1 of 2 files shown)


playwright/_impl/_browser_type.py
```diff
@@ -166,6 +167,8 @@ async def connect_over_cdp(
         headers: Dict[str, str] = None,
     ) -> Browser:
         params = locals_to_params(locals())
+        if params.get("headers"):
```
blv-matheuslins, line 170: Will that be applied to the sync context as well?

mxschmitt: yes

yury-s (approved):

Issue #1049, opened 2021-11-22
Use wider collection types instead of narrow in parameters


kumaraditya303:
Use collections.abc.Sequence instead of typing.List in func parameters etc.

mxschmitt:
Could you elaborate on that? What it gives us, the risks and whats required in terms of changes?

kumaraditya303:
Playwright requires List here https://github.com/microsoft/playwright-python/blob/174d6249228c74813a23eeaa2df26dee73ba86d3/playwright/async_api/_generated.py#L11040 but it should accept any sequence otherwise even tuple cannot be used there.

See also https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#wide-vs-narrow-types

mxschmitt:
We recommend lists and internally check always for `isinstance(..., list)`. This would require changes in playwright.dev, internally classes and generator so we keep it for now how it is caused by the complexity of the changes.

PR #1954, opened 2023-06-05
use `Sequence` instead of `List` in `FileChooser.set_files`

### LINKED ISSUE #1953: [BUG] `FileChooser.set_files` takes a `list[str | Path]` which is invariant
**Context:**
- Playwright Version: 1.34.0

**Code Snippet**

```python
with page.expect_file_chooser() as file_chooser:
    ...
file_chooser.value.set_files(["foo.txt", "bar.txt"])
```

**Describe the bug**
mypy error:
```
Argument 1 to "set_files" of "FileChooser" has incompatible type "list[str]"; expected "str | Path | FilePayload | list[str | Path] | list[FilePayload]"  [arg-type]
"List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
Consider using "Sequence" instead, which is covariant
```

it should be using a `Sequence` instead, as the error suggests

DetachHead:
fixes #1953


mxschmitt:
Thank you for your contribution! Looks like [lint](https://github.com/microsoft/playwright-python/actions/runs/5173277563/jobs/9348662107?pr=1954) is failing. So we'll have to adjust the generator, probably around [here](https://github.com/microsoft/playwright-python/blob/9f73b8844bb83e9c159559f873c005752ad00bcf/scripts/documentation_provider.py#L428). We thought about doing it already in https://github.com/microsoft/playwright-python/issues/1049#issuecomment-1001689086 and were unsure about how it affects our isinstance checks. So this requires definitely adding some test coverage.

PR #1962, opened 2023-06-08
Fix context manager freeze on SIGINT

x-mass:
Current PR contains tiny changes to prevent context manager from freezing in `__aexit__` if SIGINT received. For instance, we have the following code:
```python
import asyncio

async def main():
        await asyncio.sleep(100)

asyncio.run(main())
```
We run it and interrupt with Ctrl+C, while `sleep` is being awaited. Then we have execution chain as: https://github.com/microsoft/playwright-python/blob/45b1312bda8b36c5ec546c52431d01297bd06aa9/playwright/async_api/_context_manager.py#L54-L58
https://github.com/microsoft/playwright-python/blob/45b1312bda8b36c5ec546c52431d01297bd06aa9/playwright/_impl/_connection.py#L281-L284
https://github.com/microsoft/playwright-python/blob/45b1312bda8b36c5ec546c52431d01297bd06aa9/playwright/_impl/_transport.py#L104-L105
But `self._stopped_future` would never be set, because 
https://github.com/microsoft/playwright-python/blob/45b1312bda8b36c5ec546c52431d01297bd06aa9/playwright/_impl/_transport.py#L164-L165
is never reached due to `asyncio.CancelledError` raised in the above loop.

Looks like the last lines should be executed regardless what happened in the loop. That's why I moved it to `finally` block

### DIFF
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -139,38 +139,40 @@ class PipeTransport(Transport):
     async def run(self) -> None:
         assert self._proc.stdout
         assert self._proc.stdin
-        while not self._stopped:
-            try:
-                buffer = await self._proc.stdout.readexactly(4)
-                if self._stopped:
-                    break
-                length = int.from_bytes(buffer, byteorder="little", signed=False)
-                buffer = bytes(0)
-                while length:
-                    to_read = min(length, 32768)
-                    data = await self._proc.stdout.readexactly(to_read)
+        try:
+            while not self._stopped:
+                try:
+                    buffer = await self._proc.stdout.readexactly(4)
                     if self._stopped:
                         break
-                    length -= to_read
-                    if len(buffer):
-                        buffer = buffer + data
-                    else:
-                        buffer = data
-                if self._stopped:
+                    length = int.from_bytes(buffer, byteorder="little", signed=False)
+                    buffer = bytes(0)
+                    while length:
+                        to_read = min(length, 32768)
+                        data = await self._proc.stdout.readexactly(to_read)
+                        if self._stopped:
+                            break
+                        length -= to_read
+                        if len(buffer):
+                            buffer = buffer + data
+                        else:
+                            buffer = data
+                    if self._stopped:
+                        break
+
+                    obj = self.deserialize_message(buffer)
+                    self.on_message(obj)
+                except asyncio.IncompleteReadError:
+                    if not self._stopped:
+                        self.on_error_future.set_exception(
+                            Exception("Connection closed while reading from the driver")
+                        )
                     break
+                await asyncio.sleep(0)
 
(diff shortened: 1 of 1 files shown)


x-mass:
> @microsoft-github-policy-service agree

mxschmitt:
I tried your reproduction and for me when I CMD+C it, it does not hang, I get the following stack trace. Which OS and Python version are you using?

```
^CTraceback (most recent call last):
  File "/Users/maxschmitt/.pyenv/versions/3.11.2/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/maxschmitt/.pyenv/versions/3.11.2/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/maxschmitt/Developer/playwright-python/testa.py", line 6, in main
    await asyncio.sleep(100)
  File "/Users/maxschmitt/.pyenv/versions/3.11.2/lib/python3.11/asyncio/tasks.py", line 639, in sleep
    return await future
           ^^^^^^^^^^^^
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File "/Users/maxschmitt/Developer/playwright-python/testa.py", line 8, in <module>
    asyncio.run(main())
  File "/Users/maxschmitt/.pyenv/versions/3.11.2/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/maxschmitt/.pyenv/versions/3.11.2/lib/python3.11/asyncio/runners.py", line 123, in run
    raise KeyboardInterrupt()
KeyboardInterrupt

(env) ➜  playwright-python git:(roll/1.35.0-beta-1686247644000") ✗ 
```

x-mass:
> Which OS and Python version are you using?

I've tried Debian GNU/Linux 11 (bullseye) with both Python 3.9.17 and 3.10.11. The primary problem persists.

However, when I try Python 3.11.2, I got the same error as yours. To get rid of such error, modify the test script to:
```python
import asyncio

async def main():
    try:
            await asyncio.sleep(100)
    except asyncio.exceptions.CancelledError as e:
        pass

asyncio.run(main())
```
And it would work just fine even without current review due to handling the Ctrl+C signal as `asyncio.IncompleteReadError`, which is already caught here:
https://github.com/microsoft/playwright-python/blob/45b1312bda8b36c5ec546c52431d01297bd06aa9/playwright/_impl/_transport.py#L156-L161
Maybe modify it to be `except (asyncio.IncompleteReadError, asyncio.exceptions.CancelledError):`, so different versions could handle it the same way?

PR #2039, opened 2023-08-09
devops: do not publish conda packages with v prefix

mxschmitt:
Conda version identifier should not have a v prefix, we did this by accident.


aslushnikov (approved):

owenlamont:
Nice! We'll have to remove the v prefix from the pytest-playwright meta.yaml dependency to for its next release.

I'm not sure what the implications of the older v prefixed versions will be when comparing greater / lesser versions If some sets a min or max dependency version.

PR #2110, opened 2023-10-12
test: unflake test_page_route_should_work_with_redirects_for_subresources

mxschmitt:


tests/async/test_interception.py
```diff
@@ -328,13 +328,12 @@ async def test_page_route_should_not_work_with_redirects(page, server):
...
-    await page.route(
-        "**/*",
-        lambda route: (
-            asyncio.create_task(route.continue_()),
-            intercepted.append(route.request),
-        ),
-    )
+
+    async def handle_route(route) -> None:
+        await route.continue_()
```
yury-s, line 333: In Node.js we don't await I think, not sure how it helps with flakiness.

yury-s (approved):

tests/async/test_interception.py
mxschmitt: You are right, not landing for now.

Issue #2096, opened 2023-10-03
[Feature] Support for Python 3.12


cyberw:
Needs an update of `greenlet` to 3.0.0

xsoft7:
i have a greenlet 3.0.0 installed and cannot install playwright 1.38.0.

using latest pycharm 2023.2.2


the error is:

`Collecting playwright
  Obtaining dependency information for playwright from https://files.pythonhosted.org/packages/e1/b0/c59fcdda1a05cdfce282ba5c410d57e696aec614bba9a4f5c43bee7731fc/playwright-1.38.0-py3-none-macosx_11_0_arm64.whl.metadata
  Using cached playwright-1.38.0-py3-none-macosx_11_0_arm64.whl.metadata (3.5 kB)
Collecting greenlet==2.0.2 (from playwright)
  Using cached greenlet-2.0.2.tar.gz (164 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Collecting pyee==9.0.4 (from playwright)
  Using cached pyee-9.0.4-py2.py3-none-any.whl (14 kB)
Requirement already satisfied: typing-extensions in /Users/dannyk/PycharmProjects/pw2/venv/lib/python3.12/site-packages (from pyee==9.0.4->playwright) (4.8.0)
Using cached playwright-1.38.0-py3-none-macosx_11_0_arm64.whl (31.4 MB)
Building wheels for collected packages: greenlet
  Building wheel for greenlet (pyproject.toml): started
  Building wheel for greenlet (pyproject.toml): finished with status 'error'
Failed to build greenlet

  error: subprocess-exited-with-error
  
  × Building wheel for greenlet (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [177 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.macosx-10.9-universal2-cpython-312
      creating build/lib.macosx-10.9-universal2-cpython-312/greenlet
      copying src/greenlet/__init__.py -> build/lib.macosx-10.9-universal2-cpython-312/greenlet
      creating build/lib.macosx-10.9-universal2-cpython-312/greenlet/platform
[...]

PR #2116, opened 2023-10-12
chore: support Python 3.12

mxschmitt:
~~Blocked by https://github.com/jfhbrook/pyee/pull/135~~

---

Bisecting linting workflow failure

- Bad: 64.0.0
- Good: 63.4.3

makes it work:

```
SETUPTOOLS_ENABLE_FEATURES=legacy-editable pip install -e .
```

We had to adjust the `scripts/generate_async_api.py` imports a bit because we cannot import it with "scripts.*" since scripts is not a module.

---

Fixes https://github.com/microsoft/playwright-python/issues/2096


tests/test_generation_scripts.py
```diff
@@ -1,39 +0,0 @@
-# Copyright (c) Microsoft Corporation.
```
mxschmitt, line 1: Removed since we used this purely for code coverage tracking (https://github.com/microsoft/playwright-python/pull/99) and code coverage tracking got removed since a while.

yury-s (approved):

PR #2134, opened 2023-10-24
Update example to use context manager

ranmocy:
It makes sure resources are released. And less lines of code.

These usages are tested under:

https://github.com/microsoft/playwright-python/blob/59369fe126f49c10597d5c9099840bdc8ccdcf15/tests/sync/test_context_manager.py

https://github.com/microsoft/playwright-python/blob/59369fe126f49c10597d5c9099840bdc8ccdcf15/tests/async/test_context_manager.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,10 @@ from playwright.sync_api import sync_playwright
 
 with sync_playwright() as p:
     for browser_type in [p.chromium, p.firefox, p.webkit]:
-        browser = browser_type.launch()
-        page = browser.new_page()
-        page.goto('http://playwright.dev')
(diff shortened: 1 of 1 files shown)


ranmocy:

mxschmitt:
Thank you for your contribution. We learned over the past few years while context managers are great to use, that the community often miss-understands them, so we prefer keeping it as it is.

ranmocy:
@mxschmitt Thanks for replying. Is there any related thread or docs talking about the common mistakes? I'm very curious to learn about them.

mxschmitt:
e.g. think about that you use a context manager to get a page, you use it after the expect and wonder, why is it closed? We saw this is a popular pattern which confused users.

Issue #2154, opened 2023-11-10
[BUG] timeout not properly restored in pause()


pelme:
### System info
- Playwright Version: v1.39.0
- Operating System: All
- Browser: All

### Source code

```python
def test_invalid_link_should_timeout(page):
    page.goto("https://playwright.dev/")

    print("before pause")
    page.pause()  # <-- causes timeouts to not work
    print("after pause")

    page.get_by_role("link", name="Invalid link - should timeout!").click()
```

**Steps**
- Run the test with `pytest -s`. 

**Expected**

The test should fail after the default timeout of 30 seconds with an output like this:
```
E       playwright._impl._api_types.TimeoutError: Timeout 30000ms exceeded.
E       =========================== logs ===========================
E       waiting for get_by_role("link", name="Invalid link - should timeout!")
E       ============================================================

.venv/lib/python3.12/site-packages/playwright/_impl/_connection.py:99: TimeoutError
=============================================== short test summary info ================================================
FAILED test_stuck_after_pause.py::test_invalid_link_should_timeout[chromium] - playwright._impl._api_types.TimeoutError: Timeout 30000ms exceeded.
================================================== 1 failed in 31.09s ==================================================
```

**Actual**

The test gets stuck on `page.get_by_role(...)` forever and never times out. Commenting out `page.pause()` makes the test behave as expected.

mxschmitt:
I was able to reproduce, thank you for filing this bug.

PR #2157, opened 2023-11-13
fix: page.pause() timeout reset

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2154

### DIFF
--- a/playwright/_impl/_browser_context.py
+++ b/playwright/_impl/_browser_context.py
@@ -248,7 +248,8 @@ class BrowserContext(ChannelOwner):
     def _set_default_navigation_timeout_impl(self, timeout: Optional[float]) -> None:
         self._timeout_settings.set_default_navigation_timeout(timeout)
         self._channel.send_no_reply(
-            "setDefaultNavigationTimeoutNoReply", dict(timeout=timeout)
+            "setDefaultNavigationTimeoutNoReply",
+            {} if timeout is None else {"timeout": timeout},
         )
 
     def set_default_timeout(self, timeout: float) -> None:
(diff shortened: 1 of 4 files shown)


playwright/_impl/_browser_context.py
```diff
@@ -248,15 +248,18 @@ def set_default_navigation_timeout(self, timeout: float) -> None:
     def _set_default_navigation_timeout_impl(self, timeout: Optional[float]) -> None:
         self._timeout_settings.set_default_navigation_timeout(timeout)
         self._channel.send_no_reply(
-            "setDefaultNavigationTimeoutNoReply", dict(timeout=timeout)
+            "setDefaultNavigationTimeoutNoReply",
+            {} if timeout is None else {"timeout": timeout},
```
mxschmitt, line 252: The issue was that on protocol level we don't allow timeout to be null.

tests/async/test_page.py
```diff
@@ -1294,3 +1294,33 @@ def binding(source, o):
...
+
+async def test_page_pause_should_reset_default_timeouts(
+    page: Page, headless: bool, server: Server
+) -> None:
+    if not headless:
+        pytest.skip()
+
+    await page.goto(server.EMPTY_PAGE)
+    page.pause()
+    with pytest.raises(Error, match="Timeout 30000ms exceeded."):
```
yury-s, line 1307: Do we really want to slow down all tests by 30s?

playwright/_impl/_browser_context.py
yury-s: Should we consider serializing None as undefined or too many things break?

yury-s (approved):

playwright/_impl/_browser_context.py
mxschmitt: too many things would break I believe!

tests/async/test_page.py
mxschmitt: I think no other option instead of hacking around the impl for testing.

Issue #2078, opened 2023-09-16
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/9fe037fb63a87844e94bc4111d1a0717b22f4623 (chore: populate matcherResult in web assertions (#27133))
  - [x] https://github.com/microsoft/playwright/commit/bb4268d165fa3df300d40569fc90ab919b5ff0ba (fix(locators): do not escape regular expressions with u or v flag (#27188))
  - [x] https://github.com/microsoft/playwright/commit/287d720b8e541868e1e6fb3e0fb432f8a2bcd26b (chore: remove dedicated console channel (#27234))
  - [x] https://github.com/microsoft/playwright/commit/ffd20f43f8ee1a7a016cd9b29c372e25ec685a62 (chore: dispose stale handles to prevent oom, 1000 of a kind max (#27315))
  - [x] https://github.com/microsoft/playwright/commit/81694b7401c458964431c0b202bca99d5cfcb059 (test: unflake some tests (#27354))
  - [x] https://github.com/microsoft/playwright/commit/ac48a47d33455c83c2d62f9959b05fe69820456e (feat: toHaveAttribute without value (#27418))
  - [x] https://github.com/microsoft/playwright/commit/045e8aa3681de0a5dd732e9e19632adf6dd1f5df (chore: move `devices` from `Playwright` to `LocalUtils` (#27437))
  - [x] https://github.com/microsoft/playwright/commit/dcc8dcca739ac0032c69194bb3a133fc66e3da36 (Fix various typos (Fixes #27396) (#27391))
  - [x] https://github.com/microsoft/playwright/commit/ae5610f5c1e18240257e5d092d3ee91c41a5e328 (chore: include full messages into `pw:channel` logs (#27488))
  - [x] https://github.com/microsoft/playwright/commit/11a4b3f7f55b0681a69140c1f255c34746adaa89 (chore: remove parsed stack trace (#27496))
  - [x] https://github.com/microsoft/playwright/commit/fd6bf8aa2c7a34e97c8b97d98953411ca74f88c2 (chrome: improve error messages on vscode side (#27521))

---

Roll 1:

  - [x] https://github.com/microsoft/playwright/commit/f212fd1a8318345147c7357ced36db21acd477a0 (chore: unify target closed errors (#27540))
[...]

PR #2167, opened 2023-11-16
chore(roll): roll Playwright to 1.40.0-alpha-oct-18-2023

mxschmitt:
https://github.com/microsoft/playwright-python/issues/2078

### DIFF
--- a/playwright/_impl/_waiter.py
+++ b/playwright/_impl/_waiter.py
@@ -16,15 +16,15 @@ import asyncio
 import math
 import uuid
 from asyncio.tasks import Task
-from typing import Any, Callable, List, Tuple
+from typing import Any, Callable, List, Tuple, Union
 
 from pyee import EventEmitter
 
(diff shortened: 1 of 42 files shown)


playwright/_impl/_waiter.py
```diff
@@ -16,15 +16,15 @@
...
 
 from pyee import EventEmitter
 
-from playwright._impl._api_types import Error, TimeoutError
 from playwright._impl._connection import ChannelOwner
+from playwright._impl._errors import Error, TimeoutError
 
 
-class WaitHelper:
+class Waiter:
```
yury-s, line 27: I like the old name better, it is less ambiguous as it has less to do with restaurants business.

yury-s (approved):

playwright/_impl/_waiter.py
mxschmitt: sure, but keep the ports aligned! :/

PR #2170, opened 2023-11-21
chore(roll): roll Playwright to 1.40.0

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2078

### DIFF
--- a/playwright/_impl/_frame.py
+++ b/playwright/_impl/_frame.py
@@ -22,6 +22,7 @@ from pyee import EventEmitter
 from playwright._impl._api_structures import AriaRole, FilePayload, Position
 from playwright._impl._connection import (
     ChannelOwner,
+    filter_none,
     from_channel,
     from_nullable_channel,
 )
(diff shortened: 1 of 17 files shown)


playwright/_impl/_frame.py
```diff
@@ -689,17 +690,21 @@ async def set_input_files(
...
-            )
-        else:
-            await self._channel.send(
-                "setInputFilePaths",
-                locals_to_params({**params, **converted, "files": None}),
-            )
+        await self._channel.send(
+            "setInputFiles",
+            {
+                **filter_none(
```
mxschmitt, line 697: I might follow up and do it like in .NET that we remove None by default in send().

playwright/_impl/_set_input_files_helpers.py
```diff
@@ -34,81 +34,61 @@
...
-                }
-            )
-
-    return file_payloads
+    return InputFilesList(
+        payloads=[
+            {
+                "name": item["name"],
+                "mimeType": item["mimeType"],
+                "buffer": base64.b64encode(item["buffer"]).decode(),
```
yury-s, line 90: why is it encode than decode, is it just the python way to do base64 encoding?

yury-s (approved):

playwright/_impl/_set_input_files_helpers.py
mxschmitt: b64encode returns bytes which we need to decode to a str.

Issue #2174, opened 2023-11-27
[OUTDATED] Outdated and/or Unmaintained Tests


Vinyzu:
### System info
- Playwright Version: [v1.40]
- Operating System: All
- Browser: All
### Source code

There some tests that are outdated.
As an example see the un-awaited `(await) page.pause` in [test_page.py](https://github.com/microsoft/playwright-python/blob/main/tests/async/test_page.py#L1321).
Ive generally seen many instances where a linter should solve some issues, as some basic pep8 guidelines are not met.
This is not urgent, but proper (maintained) tests are definitely a must-have for a testing library.

mxschmitt:
Investigation notes: pyright linter does currently not lint tests/async directory.

Action item: We should fix that and let pyright lint this directory as well.

Thanks for filing!

Vinyzu:
Btw a quick hint for another thing i found whilst seting up my own tests:
[test_page.py](https://github.com/microsoft/playwright-python/blob/main/tests/async/test_page.py#L147)

I feel like the `return messages` is a debug or wrong statement.

PR #2178, opened 2023-11-29
chore: use Sequence for input List like types

mxschmitt:
Closes https://github.com/microsoft/playwright-python/pull/1954
Fixes https://github.com/microsoft/playwright-python/issues/1953
Relates https://github.com/microsoft/playwright-python/issues/2174

### DIFF
--- a/scripts/documentation_provider.py
+++ b/scripts/documentation_provider.py
@@ -172,7 +172,7 @@ class DocumentationProvider:
                 if not doc_value:
                     self.errors.add(f"Parameter not documented: {fqname}({name}=)")
                 else:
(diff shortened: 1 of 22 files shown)


scripts/documentation_provider.py
```diff
@@ -421,7 +424,7 @@ def inner_serialize_doc_type(self, type: Any, direction: str) -> str:
         if "templates" in type:
             base = type_name
             if type_name == "Array":
-                base = "List"
+                base = "Sequence"
```
dgozman, line 427: Let's make sure this is only applied to input parameters, not output values.

dgozman (approved):

PR #2180, opened 2023-11-30
test: unskip tests / unignore linting in tests

mxschmitt:
https://github.com/microsoft/playwright-python/issues/2174

### DIFF
--- a/playwright/_impl/_api_structures.py
+++ b/playwright/_impl/_api_structures.py
@@ -39,6 +39,7 @@ class Cookie(TypedDict, total=False):
     sameSite: Literal["Lax", "None", "Strict"]
 
 
+# TODO: We are waiting for PEP705 so SetCookieParam can be readonly and matches Cookie.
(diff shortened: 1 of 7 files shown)


playwright/_impl/_frame.py
```diff
@@ -111,15 +111,15 @@ def __repr__(self) -> str:
     def _on_load_state(
         self, add: DocumentLoadState = None, remove: DocumentLoadState = None
     ) -> None:
+        if not self._parent_frame and add == "load" and self._page:
```
mxschmitt, line 114: This fixes [this](https://github.com/microsoft/playwright-python/blob/248f3ec434df48f3347aeecf9c0b0ae64b4e1714/tests/async/test_page.py#L167-L174) test because we used to dispatch the `load` event after we dispatched the `loadstate` update to the eventemitter.

dgozman (approved):

Issue #2193, opened 2023-12-15
[BUG] Page.route URL glob matching is broken


hchargois:
### System info
- Playwright Version: [v1.40.0]
- Operating System: [All]
- Browser: [All]

### Source code

Basically using the example in the documentation here https://playwright.dev/python/docs/api/class-page#page-route

```python
from playwright.sync_api import sync_playwright

with sync_playwright() as p:
    browser = p.chromium.launch()
    page = browser.new_page()

    # this doesn't work as expected
    page.route("**/*.{png,jpg,jpeg}", lambda route: route.abort())

    page.goto("https://playwright.dev")
    page.screenshot(path="example.png")
    browser.close()
```

**Expected**

All image requests should be aborted

**Actual**

Image requests are not aborted

### Explanation

The `_helper.py:URLMatcher` class uses Python standard library's `fnmatch` to translate the first argument of `page.route()` to a regexp:

```
            regex = fnmatch.translate(match)
            self._regex_obj = re.compile(regex)
```

That is incorrect. `fnmatch` syntax is not the same as the glob syntax. The `match` is used to set up interception patterns, while the `_regex_obj` is used to match the correct handler once the request is intercepted. Which means that if you call `page.route()`...

* with (correct) glob syntax (e.g. `**/*.{png,jpg,jpeg}` as in the example), the interception pattern is correctly set up, but the regex will be wrong and won't match, and the handler not called
* with `fnmatch` syntax (e.g. `*.png`), the interception pattern will not be set up correctly even though the regex would match

### Solution(s)

I see only 2 solutions: either `page.route()` uses glob syntax, or it uses `fnmatch` syntax. That's a choice to make. I guess using glob syntax is the best, as it is what's already documented, and what's used in Playwright implementations in other languages.

If it uses glob syntax, then `fnmatch.translate` must be swapped for a correct glob matcher (i.e. a Python version of this https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/glob.ts).

If it uses `fnmatch` syntax, then the interception pattern must correctly match (a superset of) the `fnmatch` pattern. Trivially, this could just be the `{"glob": "**/*"}` catchall, as is already done when using a callback as the route matcher (see `RouteHandler.prepare_interception_patterns()`).  In that case, the documentation of `page.route()` referenced above should be updated.

PR #2230, opened 2024-01-09
chore: migrate to own glob parser

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2193


tests/async/test_page_request_fallback.py
```diff
@@ -164,10 +164,9 @@ async def handler_with_header_mods(route: Route) -> None:
     await page.route("**/*", handler_with_header_mods)
 
     await page.goto(server.EMPTY_PAGE)
-    async with page.expect_request("/sleep.zzz") as request_info:
```
mxschmitt, line 167: this and following aligns the test with upstream.

yury-s (approved):
I'd rather go the other way and use third-party library for matching globs. But since you maintain python port it's your call.

PR #2290, opened 2024-02-11
test: unflake test_locators_should_support_locator_and

mxschmitt:
**Motivation:** This will make the Docker tests green.

**Investigation notes:**

The issue was that we changed the testID for the selectors in the `test_should_use_data_test_id_in_strict_errors` test while it was used in the `test_locators_should_support_locator_and` test.

This will reset them inside the test for now. Other languages will reset it in the test-runner, but since in Playwright for Python we don't use our `pytest-playwright` plugin for our internal tests, I'll do the encapsulated fix for now.


yury-s (commented):
lgtm

PR #2261, opened 2024-01-25
test: add test for storage_state with lone surrogates

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2819

Issue #2246, opened 2024-01-17
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/d023829dd543c0ced528a60a0365327c39c6e419 (fix(locator parser): allow escaped quotes in the digest function (#29012))
  - [x] https://github.com/microsoft/playwright/commit/f3fac6f4e95f0818a47c88c055854576d98103e0 (feat: `page.handleLocator` that closes any interstitial pages/dialogs (#29029))
  - [x] https://github.com/microsoft/playwright/commit/d89ed88679cbbd27c92f576b8e08b9b5050ac7d7 (docs: do not recommend PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD anymore (#29121))
  - [x] https://github.com/microsoft/playwright/commit/3e2d5b2800e3143280c0573507b1ba9d5c438a00 (docs: add an example of how to use "+" key as part of a shortcut (#29221))
  - [x] https://github.com/microsoft/playwright/commit/c85308d717b757f2e9a4616b3131e725e1a0c8fb (chore: remove deprecated debug controller events (#29203))
  - [x] https://github.com/microsoft/playwright/commit/13550b7329c428063b95663c979c241c27b32656 (fix: add new line before sourceURL (#29199))
  - [x] https://github.com/microsoft/playwright/commit/aff6cf3c83d5567ec2c662eebdc246090df39982 (chore: extract ws server util (#29247))
  - [x] https://github.com/microsoft/playwright/commit/47f8ba2a040c8383a472b4c27bf392b899fcfde6 (feat(electron): expose ElectronApplication console events (#29322))
  - [x] https://github.com/microsoft/playwright/commit/20699c36ba82c5d65993c70fc4ed12141059513b (chore: simplify types handling in toHaveScreenshot (#29374))
  - [x] https://github.com/microsoft/playwright/commit/721d84f17a0e3b906fc1c5dd91d6932073342072 (fix(har): do not throw on -1 status code (#29371))
  - [x] https://github.com/microsoft/playwright/commit/61955e55b1884b778e256687a85f7cd44c3fcdc4 (fix(handleLocator): address API review feedback (#29412))
  - [x] https://github.com/microsoft/playwright/commit/f0244b8a7640ef1e0abf5588dc9780cca8bcd715 (chore: rename handleLocator to addLocatorHandler (#29443))
[...]

PR #2313, opened 2024-02-20
chore(roll): roll Playwright to 1.42.0-alpha-2024-02-19

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2246
Fixes https://github.com/microsoft/playwright-python/pull/2261

### DIFF
--- a/playwright/_impl/_page.py
+++ b/playwright/_impl/_page.py
@@ -20,10 +20,10 @@ import sys
 from pathlib import Path
 from types import SimpleNamespace
 from typing import (
-    Coroutine,
     TYPE_CHECKING,
     Any,
     Callable,
+    Coroutine,
     Dict,
     List,
     Literal,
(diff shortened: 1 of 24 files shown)


playwright/_impl/_page.py
```diff
@@ -1029,6 +1046,8 @@ async def pdf(
         preferCSSPageSize: bool = None,
         margin: PdfMargins = None,
         path: Union[str, Path] = None,
+        outline: bool = None,
```
danphenderson, line 1049: Shouldn't these be `outline: Optional[bool] = None` etc.

dgozman (approved):

playwright/_impl/_page.py
danphenderson: Please let me know if I am missing something.

mxschmitt: We use `strictParameterNoneValue = false` hence changing a single parameter won't make a big difference there.

mxschmitt: https://github.com/microsoft/playwright-python/blob/976f04fb4d56d1497fd85e0c5199a9daeed4a138/pyproject.toml#L44

PR #2318, opened 2024-02-26
Read stream for downloads

BdEgh:
All languages except Python seem to have such a thing.
Why not support this feature here as well?


BdEgh:

BdEgh:
Just fixed mypy errors, nothing more
Should be ok now

BdEgh:
I can't really find what linter wants to be documented :(

A comment here isn't enough?
https://github.com/microsoft/playwright-python/pull/2318/files#diff-d6f13195208ce0c069a6fab675c2cbfb71019d9eefc0a5c2a64bfb596b690238R6966-R6969

Issue #2136, opened 2023-10-25
[Feature] Document page.route 


EvanTheB:
I am having some trouble using page.route within pytest. I think the issue stems from the way the route functions are run. I believe the routing is run in a background thread, and the routes are eventually run inside a coroutine event loop. 

Relevant docs: 

https://playwright.dev/python/docs/network

https://playwright.dev/python/docs/api/class-route

https://playwright.dev/python/docs/api/class-page#page-route

I think the documentation needs to call out the expected behavior of the route functions clearly:

What happens when a route function throws an exception? 

What happens if the route function returns without calling one of the route methods (eg route.fulfill)?

I have observed that in pytest asserting or calling pytest.fail from a route function does not fail the test reliably. I am also observing some other non deterministic behaviour, but I do not have a simple reproduction yet. I think clarity on the above points may help.

EvanTheB:
As an example, here are some tests:

```
import time
from playwright.sync_api import Page, Route


def test_routing_ok(page: Page) -> None:

    def handle_err(route: Route):
        print("this should not happen", route.request.url)
        route.continue_()

    def handle_ok(route: Route):
        print("this is ok", route.request.url)
        route.continue_()

    page.route("http://10.42.0.10/**", handle_err)
    page.route("http://10.42.0.10/**", handle_ok)
    page.goto("http://10.42.0.10/cgi-bin/luci/admin/network/network")
    page.wait_for_load_state("networkidle")


def test_routing_no_route(page: Page) -> None:
    """This example times out.
    """
    def handle_err(route: Route):
        print("this should not happen", route.request.url)
        route.continue_()

    def handle_ok(route: Route):
        print("this is ok", route.request.url)

    page.route("http://10.42.0.10/**", handle_err)
    page.route("http://10.42.0.10/**", handle_ok)
    page.goto("http://10.42.0.10/cgi-bin/luci/admin/network/network")
    page.wait_for_load_state("networkidle")

def test_routing_raise_before(page: Page) -> None:
    """This example times out.
    How is the exception from the handle_ok function treated?
    We also see a message in the logs:
    ERROR    asyncio:base_events.py:1758 Task exception was never retrieved
    """
    def handle_err(route: Route):
        print("this should not happen", route.request.url)
        route.continue_()

    def handle_ok(route: Route):
        print("this is ok", route.request.url)
        assert False
        route.continue_()

    page.route("http://10.42.0.10/**", handle_err)
[...]

PR #2321, opened 2024-02-26
fix: report route handler errors to the user after they completed

mxschmitt:
Issue: Exceptions which will get thrown in route handlers after route.{fulfill,continue,abort} got called,
won't get surfaced to the user in the sync and async version of Playwright for Python.

Scope: Event handlers which use asyncio.create_task
  - {BrowserContext,Page}.on("route")
  - {Page}.on({"route", "locatorHandlerTriggered"})

There were multiple issues in the implementation:

1. `playwright/_impl/_helper.py` (sync only): we were only waiting until a handler's {continue,fulfill,abort} got called, and did not wait until the actual callback completed. Fix: Wait until the handler completed.
2. `playwright/_impl/_connection.py:423` (sync only): We call event listeners manually, this means that `pyee`'s `"error"` event is not working there. Fix: attach a done callback manually, like pyee is doing inside their library.
3. `playwright/_impl/_connection.py:56` (async): attach an `"error"` event handler for the async error reporting
4. `playwright/_impl/_connection.py:90` if we report an error to the user, we should cancel the underlying future otherwise a warning gets emitted, that there was a future exception never received.

This should resolve https://github.com/microsoft/playwright-python/issues/2136.

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -78,14 +78,13 @@ class Channel(AsyncIOEventEmitter):
     ) -> Any:
         if params is None:
             params = {}
-        callback = self._connection._send_message_to_server(
-            self._object, method, _filter_none(params)
-        )
         if self._connection._error:
             error = self._connection._error
             self._connection._error = None
-            callback.future.cancel()
             raise error
+        callback = self._connection._send_message_to_server(
+            self._object, method, _filter_none(params)
+        )
         done, _ = await asyncio.wait(
             {
                 self._connection._transport.on_error_future,
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -53,6 +53,7 @@ class Channel(AsyncIOEventEmitter):
         self._connection = connection
         self._guid = object._guid
         self._object = object
+        self.on("error", lambda exc: self._connection._on_event_listener_error(exc))
 
     async def send(self, method: str, params: Dict = None) -> Any:
         return await self._connection.wrap_api_call(
@@ -83,6 +84,7 @@ class Channel(AsyncIOEventEmitter):
         if self._connection._error:
             error = self._connection._error
             self._connection._error = None
+            callback.future.cancel()
             raise error
         done, _ = await asyncio.wait(
             {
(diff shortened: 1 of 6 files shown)


tests/async/test_browsercontext_request_intercept.py
```diff
@@ -179,3 +180,14 @@ async def test_should_give_access_to_the_intercepted_response_body(
...
+
+
+async def test_should_show_exception_after_fulfill(page: Page, server: Server) -> None:
+    async def _handle(route: Route) -> None:
+        await route.continue_()
+        raise Exception("Exception text!?")
+
+    await page.route("*/**", _handle)
+    await page.goto(server.EMPTY_PAGE)
+    with pytest.raises(Exception, match="Exception text!?"):
```
dgozman, line 192:
```suggestion
    # Any next API call should throw because handler did throw during previous goto()
    with pytest.raises(Exception, match="Exception text!?"):
```

playwright/_impl/_connection.py
```diff
@@ -83,6 +84,7 @@ async def inner_send(
         if self._connection._error:
             error = self._connection._error
             self._connection._error = None
+            callback.future.cancel()
```
dgozman, line 87: Perhaps we should first check the error, and if there is no error, only then send the message?

playwright/_impl/_connection.py
```diff
@@ -416,10 +418,22 @@ def dispatch(self, msg: ParsedMessagePayload) -> None:
...
             if self._is_sync:
                 for listener in object._channel.listeners(method):
+                    # Event handlers like route/locatorHandlerTriggered require us to perform async work.
+                    # In order to report their potential errors to the user, we need to catch it and store it in the connection
+                    def _done_callback(future: asyncio.Future) -> None:
+                        exc = future.exception()
+                        if exc:
+                            self._on_event_listener_error(exc)
+
+                    def _listener_with_error_handler_attached(params: Any) -> None:
```
dgozman, line 428: Let's make sure we have test coverage for this.

dgozman (approved):

smolkin895:
Hi!
I'm sorry, but I can't do a pytest assert even before calling the fullfiil, fallback, continue_ function. An exception occurs inside hadler, but pytest does not crash, and the endpoint whose route I am monitoring does not receive a response. How to do assert in a network in a convenient way? This is one of the most important parts of UI testing.

PR #2322, opened 2024-02-26
fix: datetime parsing with non-UTC timezones

mxschmitt:
Motivation: This makes datetime with non-UTC timezone objects

### DIFF
--- a/playwright/_impl/_js_handle.py
+++ b/playwright/_impl/_js_handle.py
@@ -129,7 +129,7 @@ def serialize_value(
         if math.isnan(value):
             return dict(v="NaN")
     if isinstance(value, datetime):
-        return dict(d=value.isoformat() + "Z")
+        return dict(d=value.isoformat())
     if isinstance(value, bool):
         return {"b": value}
     if isinstance(value, (int, float)):
--- a/playwright/_impl/_js_handle.py
+++ b/playwright/_impl/_js_handle.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 import collections.abc
+import datetime
 import math
-from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


tests/async/test_evaluate.py
```diff
@@ -216,17 +216,30 @@ async def test_evaluate_evaluate_date(page: Page) -> None:
...
+        "date": datetime.fromisoformat("2020-05-27T01:31:38.506").replace(
+            tzinfo=timezone.utc
+        )
+    }
 
 
 async def test_evaluate_roundtrip_date(page: Page) -> None:
     date = datetime.fromisoformat("2020-05-27T01:31:38.506")
     result = await page.evaluate("date => date", date)
+    assert result.timestamp() == date.timestamp()
```
yury-s, line 229: Why do we compare only timestamps, is `result != date`?

playwright/_impl/_js_handle.py
```diff
@@ -205,7 +210,10 @@ def parse_value(value: Any, refs: Optional[Dict[int, Any]] = None) -> Any:
             return a
 
         if "d" in value:
-            return datetime.fromisoformat(value["d"][:-1])
+            # Node.js Date objects are always in UTC.
+            return datetime.datetime.strptime(
```
yury-s, line 214: It's unfortunate that we send an iso string over the wire, parsing millis since epoch (with `datetime.fromtimestamp`) would be more straightforward. But that's a big change.

yury-s (approved):

tests/async/test_evaluate.py
mxschmitt: Because `datetime.fromisoformat("2020-05-27T01:31:38.506")` has `tzinfo=None` while result has `tzinfo=UTC` so the strict assert would fail.

PR #2341, opened 2024-03-06
chore(context): when running without with context

aamirbhat:
context api has start method but no stop method this causes issue while runnig multiple files fixes #29827


mxschmitt:
I see we have something similar already, would that work for you?

```python
    playwright = sync_playwright().start()
    playwright.stop()
    playwright.stop()
```

mxschmitt:
Closing as per above, since we have Playwright [`playwright.stop`](https://playwright.dev/python/docs/api/class-playwright#playwright-stop) for that already.

PR #2225, opened 2024-01-05
chore(roll): roll to Playwright 1.41.0-beta-1705101589000

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2176

PR #2345, opened 2024-03-07
Fix browser_context.route for service worker requests

LironHaroni:
Fix issue: https://github.com/microsoft/playwright-python/issues/2346

`browser_context.route` is broken for requests initiated by service worker:
https://github.com/microsoft/playwright-python/pull/2225/files#diff-4deacaefd5231b207c0afda793da70cb3694e10165da8fccd2ac9f1a03a36c2dR270


LironHaroni:
@microsoft-github-policy-service agree company="Savvy Security"

playwright/_impl/_network.py
```diff
@@ -262,6 +266,8 @@ def _target_closed_future(self) -> asyncio.Future:
         return page._closed_or_crashed_future
 
     def _safe_page(self) -> "Optional[Page]":
+        if self.from_service_worker:
+            return None
         return cast("Frame", from_channel(self._initializer["frame"]))._page
```
mxschmitt, line 271:
```suggestion
        frame = from_nullable_channel(self._initializer["frame"])
        if not frame:
            return None
        return cast("Frame", frame)._page
```

maybe something like this instead? This would align it with the upstream port: 

https://github.com/microsoft/playwright/blob/9ca895dea087a36219202624a8550921bb924bf0/packages/playwright-core/src/client/network.ts#L219

LironHaroni: `self._initializer["frame"]` this will throw `"KeyError" `
I can do something like this:
```
try:
         frame = from_nullable_channel(self._initializer["frame"])
except KeyError:
         return None
return cast("Frame", frame)._page
```

mxschmitt: Let's do:

```python
   frame = from_nullable_channel(self._initializer.get("frame"))
   if not frame:
       return None
   return cast("Frame", frame)._page
```

mxschmitt (approved):

PR #2435, opened 2024-05-10
prepare support for Nuitka

KRRT7:
this PR will prepare playwright to make it easier to be compiled into standalone / onefile programs for ease of distribution
the following changes have been made:

1) added the python version of 
 ```js
const registryDirectory = exports.registryDirectory
```

in `playwright._impl._driver`, Nuitka will use this to resolve the path to the browsers .

2) check if `_compiled__` is in the globals dict, if it is, set `PLAYWRIGHT_BROWSERS_PATH` to 0 just like for pyinstaller

### DIFF
--- a/playwright/_impl/_driver.py
+++ b/playwright/_impl/_driver.py
@@ -22,31 +22,6 @@ import playwright
 from playwright._repo_version import version
 import pathlib
 
-def get_registry_directory() -> str:
-    env_defined = os.environ.get('PLAYWRIGHT_BROWSERS_PATH')
-    path_home = Path.home()
-    if env_defined == '0':
-        result = pathlib.Path(__file__).parent.parent.parent.joinpath('.local-browsers')
-    elif env_defined:
-        result = pathlib.Path(env_defined)
-    else:
-        cache_directory: Union[str, Path] = ''
-        if os.name == 'posix':
-            cache_directory = os.environ.get('XDG_CACHE_HOME', path_home / '.cache')
-        elif os.name == 'darwin':
-            cache_directory = path_home / 'Library' / 'Caches'
-        elif os.name == 'nt':
-            cache_directory = os.environ.get('LOCALAPPDATA', path_home / 'AppData' / 'Local')
-        else:
-            raise RuntimeError('Unsupported platform: ' + os.name)
-        result = pathlib.Path(cache_directory, 'ms-playwright')
-
-    if not result.is_absolute():
-        init_cwd = os.environ.get('INIT_CWD') or os.getcwd()
-        result = pathlib.Path(init_cwd).resolve().joinpath(result)
-
-    return str(result)
-
 
 def compute_driver_executable() -> Tuple[str, str]:
     driver_path = Path(inspect.getfile(playwright)).parent / "driver"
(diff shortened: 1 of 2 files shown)


KRRT7:
> @KRRT7 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

mxschmitt:
happy to accept the changes in `playwright/_impl/_transport.py`. What is the rationale of adding the `get_registry_directory` method? Its not used.

mxschmitt:
We have the following workflow for Pyinstaller, that should also work for Nuitka?

1. `PLAYWRIGHT_BROWSERS_PATH=0 playwright install chromium` This will install the browsers locally, inside the site-modules directory.
2. `pyinstaller -F main.py` This will bundle the site-modules directory.

If Nuitka isn't bundling all the files inside the site-modules directory, we recommend instead do do something like this:

1. `PLAYWRIGHT_BROWSERS_PATH=./browsers playwright install chromium` This will install the browsers locally inside `./browsers`.
2. And set `PLAYWRIGHT_BROWSERS_PATH=./browsers` inside your program, so Playwright knows about it.

I'm not a huge fan of `get_registry_directory` since in my case, it contains 50 browsers, for a lot of Playwright versions - we'd need to match first while the solution above allows you to install only what you want.

KRRT7:
we don't like affecting the user experience, so we generally just look wherever the program looks and go from there, this is in case the user runs from either an venv, or global packages, users don't normally set the `PLAYWRIGHT_BROWSERS_PATH=0` flag unless they're using pyinstaller or set `PLAYWRIGHT_BROWSERS_PATH` unless they want the browsers in different location, we're also looking into the directory ourselves so that we can look thru the folder and see what browsers are available, and then we prompt the user on which one they want to use as to avoid including unnecesary bloat, i.e including all browsers when they're only using a particular one, however i'll be happy with just the changes in `playwright/_impl/_transport.py` :)

KRRT7:
another question, i see playwright downloads ffmpeg when you do a bare `playwright install`, i would like to know when ffmpeg is used or not so that we also know when to bundle it? thanks.

mxschmitt:
> another question, i see playwright downloads ffmpeg when you do a bare `playwright install`, i would like to know when ffmpeg is used or not so that we also know when to bundle it? thanks.

ffmpeg is needed for chromium based browsers in order to record video.

> however i'll be happy with just the changes in playwright/_impl/_transport.py :

let's do that, then the Nuitka experience is on par with Pyinstaller.

mxschmitt (approved):

PR #2462, opened 2024-06-11
fix(driver): consider PLAYWRIGHT_NODEJS_PATH from host env win32

MikhailSeliakov:
Adds the ability to use the env variable on win32 PLAYWRIGHT_NODEJS_PATH to use local nodejs


mxschmitt (approved):

mxschmitt:
Thanks looks like I missed that!

PR #2474, opened 2024-07-01
chore(roll): roll to Playwright v1.45.0-beta-1719505820000

mxschmitt:
`fromtimestamp` and `timestamp()` both return float's - so choosing a float seems more reasonable here.

### DIFF
--- a/playwright/_impl/_clock.py
+++ b/playwright/_impl/_clock.py
@@ -32,7 +32,7 @@ class Clock:
 
     async def fast_forward(
         self,
-        ticks: Union[float, str],
+        ticks: Union[int, str],
     ) -> None:
         await self._browser_context._channel.send(
             "clockFastForward", parse_ticks(ticks)
@@ -51,7 +51,7 @@ class Clock:
 
     async def run_for(
         self,
-        ticks: Union[float, str],
+        ticks: Union[int, str],
     ) -> None:
         await self._browser_context._channel.send("clockRunFor", parse_ticks(ticks))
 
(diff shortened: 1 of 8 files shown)


playwright/_impl/_clock.py
```diff
@@ -25,22 +25,22 @@ def __init__(self, browser_context: "BrowserContext") -> None:
         self._loop = browser_context._loop
         self._dispatcher_fiber = browser_context._dispatcher_fiber
 
-    async def install(self, time: Union[int, str, datetime.datetime] = None) -> None:
+    async def install(self, time: Union[float, str, datetime.datetime] = None) -> None:
```
pavelfeldman, line 28: Why is it float? Figuring out if this will confuse Python folks that use seconds for time.

playwright/_impl/_clock.py
```diff
@@ -57,25 +57,27 @@ async def run_for(
...
         await self._browser_context._channel.send(
             "clockSetSystemTime", parse_time(time)
         )
 
 
-def parse_time(time: Union[int, str, datetime.datetime]) -> Dict[str, Union[int, str]]:
-    if isinstance(time, int):
-        return {"timeNumber": time}
+def parse_time(
+    time: Union[float, int, str, datetime.datetime]
```
pavelfeldman, line 74: drop int

playwright/sync_api/_generated.py
```diff
@@ -6818,15 +6820,13 @@ def fast_forward(self, ticks: typing.Union[int, str]) -> None:
         Parameters
         ----------
         ticks : Union[int, str]
-            Time may be the number of milliseconds to advance the clock by or a human-readable string. Valid string formats are
```
pavelfeldman, line 6821: why did it go?

scripts/documentation_provider.py
```diff
@@ -483,6 +483,8 @@ def inner_serialize_doc_type(self, type: Any, direction: str) -> str:
             return "bool"
         if type_name.lower() == "string":
             return "str"
+        if type_name.lower() == "long":
```
pavelfeldman, line 486: That's not great, I'd rather fix upstream

pavelfeldman (approved):

pavelfeldman (approved):

PR #2499, opened 2024-07-29
chore(roll): roll Playwright to v1.46

mxschmitt:


tests/async/test_page_evaluate.py
```diff
@@ -208,8 +208,52 @@ async def test_evaluate_throw_if_underlying_element_was_disposed(page: Page) ->
...
+        """e => {
+            return { message: e.message, name: e.name, stack: e.stack };
+        }""",
+        error_for_roundtrip,
+    )
+    assert error["message"] == "error message"
+    assert error["name"] == "foobar"
+    assert "test stack" in error["stack"]
+
+    error = await page.evaluate(
```
mxschmitt, line 248: maybe use class/type name as a Error name?

tests/async/test_browsercontext_client_certificates.py
```diff
@@ -0,0 +1,127 @@
+# Copyright (c) Microsoft Corporation.
```
mxschmitt, line 1: Don't forget these tests to the tests/sync directory.

playwright/_impl/_network.py
```diff
@@ -83,6 +84,34 @@ def serialize_headers(headers: Dict[str, str]) -> HeadersArray:
...
+    clientCertificates: Optional[List[ClientCertificate]],
+) -> Optional[List[Dict[str, str]]]:
+    if not clientCertificates:
+        return None
+    out = []
+    for clientCertificate in clientCertificates:
+        out_record = {
+            "origin": clientCertificate["origin"],
+        }
+        if passphrase := clientCertificate.get("passphrase"):
```
mxschmitt, line 97: Note: walrus operator is supported since Python 3.8+ (last version we support) and helps us here esp. with optional types. since inside the if its e.g. not Optional anymore.

tests/async/test_page_evaluate.py
dgozman: Yeah, why not. We should do the same as we do in `BindingCall.reject`. We can change both to include the name.

dgozman (approved):

Skn0tt (approved):

PR #2507, opened 2024-08-01
add service worker property

LironHaroni:
This addition allows us to identify requests from the service worker


LironHaroni:
@mxschmitt

playwright/_impl/_network.py
```diff
@@ -233,6 +233,10 @@ def frame(self) -> "Frame":
 
     def is_navigation_request(self) -> bool:
         return self._initializer["isNavigationRequest"]
+    
+    @property
+    def service_worker(self) -> bool:
```
mxschmitt, line 238: We usually don't expose Chromium only features in the language ports, hence we [didn't add it so far](https://github.com/microsoft/playwright/blob/main/docs/src/api/class-request.md#method-requestserviceworker). So far `service_name` is only exposed in Node.js and exposes a `Worker` object rather than a boolean. What is your use-case for it? This would help us to prioritise it.

Looking at:

> Requests originated in a Service Worker do not have a [method: Request.frame] available.

It looks like if you check for `request.frame is None` it might work as well.

LironHaroni: Thanks for replying! 🙏  My use-case is when I use route to fulfill requests with a desired response - For that, I need to know what is the source of that request - if it is from the service worker I need it to be responded with a specific response.

LironHaroni: what do you think? 😃 🙏 @mxschmitt

PR #2541, opened 2024-08-30
refactor(errors): extend TimeoutError and TargetClosedError with built-in exceptions

kaixinol:


mxschmitt:
Last time we did it we reverted it: https://github.com/microsoft/playwright-python/pull/2312

I would still wait until the issue reaches more momentum or it fixes a concrete user-facing bug.

Issue #2497, opened 2024-07-29
[Ports]: Backport client side changes for 1.47


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/f45cf659215e97f921add54f6de44cbc171ab303 (chore: add maxRetries to APIRequestContext.delete (#31893))
  - [x] https://github.com/microsoft/playwright/commit/58b0c76f2068417279fdb6acc34fe8db38c19c16 (chore: use soft event emitter (#31868))
  - [x] https://github.com/microsoft/playwright/commit/55187207e43462ea5b89de4675cca98bb893423d (chore: various roll fixes for .NET (#31914))
  - [x] https://github.com/microsoft/playwright/commit/613ccb8d5b8aca5a51c73dbbdce8afcadf6a8add (chore(client-certificates): rewrite error for unsupported PFX errors (#32008))
  - [x] https://github.com/microsoft/playwright/commit/193013c9ee40185aca8d6ee0cbddd8abcc4ed7d9 (docs(har): default update mode is minimal (#32016))
  - [x] https://github.com/microsoft/playwright/commit/3c87f217df293119964fe42e532e638dd9e17073 (feat(events): allow waiting for removeAllListeners (#31941))
  - [x] https://github.com/microsoft/playwright/commit/43e852334b451c34336d2dafe104831a836a0866 (docs: route.fallback() vs. route.continue() (#32035))
  - [x] https://github.com/microsoft/playwright/commit/2ae196f7080d79df352400b36cec8d0344fb7570 (fix(docs): API types do not extend EventEmitter (#32124))
  - [x] https://github.com/microsoft/playwright/commit/308381eeae85cb7b260758f5b359dcf20b9105bf (feat: allow `URLSearchParams` and `string` as params in `APIRequestContext` (#32120))
  - [x] https://github.com/microsoft/playwright/commit/b7ed4d7b9e433f1be414c6e580a0ddc39de71618 (docs: deprecate: Request.serviceWorker() (#32136))
  - [x] https://github.com/microsoft/playwright/commit/058883430740044783125df8ae2e3a3fbe9f7718 (feat: allow `URLSearchParams` and `string` as params in `APIRequestContext` (follow-up) (#32143))
  - [x] https://github.com/microsoft/playwright/commit/f8eef3897c0f522e79d321cc6398660c55ed534c (chore: move urlMatch to isomorphic (#32142))
[...]

PR #2546, opened 2024-09-09
chore(roll): roll Playwright to v1.47.0

mxschmitt:
Corresponding upstream PR wich was required to make doclint pass: https://github.com/microsoft/playwright/pull/32516

Fixes https://github.com/microsoft/playwright-python/issues/2497

### DIFF
--- a/playwright/_impl/_fetch.py
+++ b/playwright/_impl/_fetch.py
@@ -18,6 +18,7 @@ import pathlib
 import typing
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union, cast
+from urllib.parse import parse_qs
 
 import playwright._impl._network as network
 from playwright._impl._api_structures import (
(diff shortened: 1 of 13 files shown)


playwright/_impl/_fetch.py
```diff
@@ -429,6 +430,23 @@ async def storage_state(
...
+        return None
+    if isinstance(params, dict):
+        return object_to_array(params)
+    if params.startswith("?"):
+        params = params[1:]
+    parsed = parse_qs(params)
+    if not parsed:
+        return None
+    out = []
+    for name, values in parsed.items():
```
dgozman, line 444: I wonder why don't we pass a string over the protocol instead? That would be aligned with our approach to do as much on the server, to avoid aligning the details on the client side. Let's discuss with the team.

dgozman (approved):

PR #2588, opened 2024-10-09
chore: update linters 2024

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2582

Recent VSCode versions were using more recent versions of Pyright which was showing errors. This updates our linter setup.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v5.0.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
@@ -15,16 +15,16 @@ repos:
       - id: check-executables-have-shebangs
       - id: check-merge-conflict
   - repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 24.8.0
     hooks:
       - id: black
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.8.0
+    rev: v1.11.2
     hooks:
       - id: mypy
-        additional_dependencies: [types-pyOpenSSL==23.2.0.2, types-requests==2.31.0.10]
+        additional_dependencies: [types-pyOpenSSL==24.1.0.20240722, types-requests==2.32.0.20240914]
   - repo: https://github.com/pycqa/flake8
-    rev: 6.1.0
+    rev: 7.1.1
     hooks:
       - id: flake8
   - repo: https://github.com/pycqa/isort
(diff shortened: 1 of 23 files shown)


.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v5.0.0
```
dgozman, line 5: What is this one?

local-requirements.txt
```diff
@@ -7,7 +7,7 @@ mypy==1.11.2
 objgraph==3.6.1
 Pillow==10.4.0
 pixelmatch==0.3.0
-pre-commit==3.4.0
+pre-commit==3.5.0
```
dgozman, line 10: What is this one? Why is it different from the previous file?

local-requirements.txt
```diff
@@ -7,7 +7,7 @@ mypy==1.11.2
...
 pyOpenSSL==24.2.1
```
dgozman, line 11: This one does not seem to match the previous file either.

pyproject.toml
```diff
@@ -35,10 +35,12 @@ profile = "black"
...
 include = ["playwright", "tests", "scripts"]
+exclude = ["**/node_modules", "**/__pycache__", "**/.*", "./build"]
 pythonVersion = "3.8"
 reportMissingImports = false
 reportTypedDictNotRequiredAccess = false
 reportCallInDefaultInitializer = true
 reportOptionalSubscript = false
 reportUnboundVariable = false
 strictParameterNoneValue = false
+reportIncompatibleVariableOverride = false
```
dgozman, line 46: What is this one? Do we actually want to enable it?

setup.cfg
```diff
@@ -3,3 +3,5 @@ ignore =
     E501
     W503
     E302
+    # Conflicts with black
```
dgozman, line 6: More details would be beneficial.

dgozman (approved):
IIUC, this has no behavior change for the user, has it?

pyproject.toml
mxschmitt: We override `self._loop` from being `Optional` to being not `Optional`. Mypy doesn't allow it anymore after they enabled this setting. Keeping it disabled seems fine for now imo.

.pre-commit-config.yaml
mxschmitt: Hooks which we consume from the `pre-commit` project - they are not shipped automatically so we specify them like that.

local-requirements.txt
mxschmitt: This is `pre-commit` - like ESLint - and above are some rules we consume.

local-requirements.txt
mxschmitt: In both cases the latest version. Seems like its like `definitely-typed` - so their version can differ.

mxschmitt:
> IIUC, this has no behavior change for the user, has it?

no user facing change yes.

PR #2608, opened 2024-10-22
Updated the local-requirements.txt with the latest versions.

vignesh1507:
Updated the libraries/modules with the latest version.


mxschmitt:
We have dependabot in place which takes care of updates. Please don't update them in bulk. We also usually only update them when we have a reason for it.

PR #2609, opened 2024-10-22
launch mode specification in update_versions.py

vignesh1507:
Launch Mode: Specified the headless parameter in the `browser_type.launch()` call to avoid ambiguity.


vignesh1507:

mxschmitt:
The browser runs headless by default - what are we fixing here?

mxschmitt:
Closing as per above.

PR #2610, opened 2024-10-22
Fixed few changes in setup.py

vignesh1507:
1. Dependency Versioning:

> The install_requires specifies fixed versions (`greenlet==3.1.1, pyee==12.0.0`). This could lead to installation problems if a user has a different environment or if those versions are not compatible with other libraries. You might want to specify version ranges (e.g., `greenlet>=3.1.1,<4.0.0`) to allow for flexibility.

2. Linux Wheel Handling:

> The `manylinux1_x86_64.whl` may not be the best choice if you’re targeting a broader range of Linux distributions. Consider using `manylinux2014` or `manylinux2010` for better compatibility.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,6 @@ def download_driver(zip_name: str) -> None:
         url = url + "next/"
     url = url + zip_file
     print(f"Fetching {url}")
-    # Don't replace this with urllib - Python won't have certificates to do SSL on all platforms.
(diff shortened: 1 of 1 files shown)


vignesh1507:

mxschmitt:
Please start with an issue before creating a PR which changes a lot of different internal key components. I'll close it for now since we don't plan to land it in the current form and didn't experience any issues with manylinux etc. which justifies modifying it.

Issue #2616, opened 2024-10-25
[Bug]: to_be_visible assertion says "Locator expected to be visible" if visible=False


MilovanovM:
### Version

latest

### Steps to reproduce

Steps to reproduce:
1. use the assertion `to_be_visible(visible=False)`

### Expected behavior

get `AssertionError: Locator expected to be hidden`

### Actual behavior

get `AssertionError: Locator expected to be visible`

### Environment

- Operating System: [Ubuntu 22.04]
- CPU: [arm64]
- Browser: [All, Chromium, Firefox, WebKit]
- Python Version: [3.12]
- Other info:

mxschmitt:
Looks like we need a condition here: https://github.com/microsoft/playwright-python/blob/3352d85e4493c4b95de8304d6c16df787058cadd/playwright/_impl/_assertions.py#L672

MilovanovM:
Just checked other implementations (Node, Java, donNet), no issue there

PR #2619, opened 2024-10-28
fix(assertions): error messages from negated matchers

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2616

drive-by: align more sync tests with async.

### DIFF
--- a/playwright/_impl/_assertions.py
+++ b/playwright/_impl/_assertions.py
@@ -511,15 +511,14 @@ class LocatorAssertions(AssertionsBase):
         timeout: float = None,
     ) -> None:
         __tracebackhide__ = True
+        if attached is None:
+            attached = True
+        attached_string = "attached" if attached else "detached"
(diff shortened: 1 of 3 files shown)


playwright/_impl/_assertions.py
```diff
@@ -511,15 +511,14 @@ async def to_be_attached(
...
+        if attached is None:
+            attached = True
+        attached_string = "attached" if attached else "detached"
         await self._expect_impl(
-            (
-                "to.be.attached"
-                if (attached is None or attached is True)
-                else "to.be.detached"
-            ),
+            ("to.be.attached" if attached else "to.be.detached"),
```
Skn0tt, line 518: not: should we reuse `attached_string` here?

```suggestion
            f"to.be.{attached_string}"
```

Skn0tt (approved):
looks great! added tiny nit, take it or leave it.

PR #2635, opened 2024-11-06
test: update pytest-asyncio to 0.24.0

mxschmitt:
They recommend setting it as per [here](https://github.com/pytest-dev/pytest-asyncio/blob/adcae5bea72a9e16b8420d06e23b3915c6f82333/pytest_asyncio/plugin.py#L203-L205). Full changelog see [here](https://pytest-asyncio.readthedocs.io/en/latest/reference/changelog.html).

### DIFF
--- a/tests/async/conftest.py
+++ b/tests/async/conftest.py
@@ -16,6 +16,7 @@ import asyncio
 from typing import Any, AsyncGenerator, Awaitable, Callable, Dict, Generator, List
 
 import pytest
+from pytest_asyncio import is_async_test
 
 from playwright.async_api import (
     Browser,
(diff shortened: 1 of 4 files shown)


tests/async/conftest.py
```diff
@@ -38,8 +39,10 @@ def utils() -> Generator[Utils, None, None]:
 
 # Will mark all the tests as async
 def pytest_collection_modifyitems(items: List[pytest.Item]) -> None:
-    for item in items:
-        item.add_marker(pytest.mark.asyncio)
+    pytest_asyncio_tests = (item for item in items if is_async_test(item))
```
mxschmitt, line 42: This is token from [here](https://pytest-asyncio.readthedocs.io/en/latest/how-to-guides/run_session_tests_in_same_loop.html).

dgozman (approved):

PR #2642, opened 2024-11-12
chore: convert setup.py to build

mxschmitt:


setup.py
```diff
@@ -60,132 +108,78 @@ def download_driver(zip_name: str) -> None:
 
 
 class PlaywrightBDistWheelCommand(BDistWheelCommand):
-    user_options = BDistWheelCommand.user_options + [
```
mxschmitt, line 63: Env var seems simpler, since option passing via `build` is not fully working anyways: https://github.com/pypa/setuptools/issues/2491

setup.py
```diff
@@ -19,18 +19,67 @@
...
+        "zip_name": "win32_x64",
+    },
+]
+
+if len(sys.argv) == 2 and sys.argv[1] == "--list-wheels":
+    for bundle in base_wheel_bundles:
+        print(bundle["wheel"])
+    exit(0)
+
+from setuptools import setup  # noqa: E402
```
mxschmitt, line 76: The order here is non-common because we have the dependencies available only in non-build environments since I removed the deps from `local-requirements.txt`.

playwright/sync_api/_generated.py
```diff
@@ -683,7 +683,7 @@ def fulfill(
         json: typing.Optional[typing.Any] = None,
         path: typing.Optional[typing.Union[str, pathlib.Path]] = None,
         content_type: typing.Optional[str] = None,
-        response: typing.Optional["APIResponse"] = None
```
mxschmitt, line 686: I'll extract them into a separate PR tomorrow.

playwright/async_api/_generated.py
```diff
@@ -4989,7 +4989,7 @@ def get_by_text(
         self,
         text: typing.Union[str, typing.Pattern[str]],
         *,
-        exact: typing.Optional[bool] = None
+        exact: typing.Optional[bool] = None,
```
yury-s, line 4992: can you revert this trailing coma change as it is unrelated, can be committed in a separate PR?

setup.py
```diff
@@ -60,132 +109,78 @@ def download_driver(zip_name: str) -> None:
...
-                "machine": "amd64",
-                "platform": "win32",
-                "zip_name": "win32_x64",
-            },
-        ]
-        self._download_and_extract_local_driver(base_wheel_bundles)
-
-        wheels = base_wheel_bundles
-        if not self.all:
-            # Limit to 1, since for MacOS e.g. we have multiple wheels for the same platform and architecture and Conda expects 1.
```
yury-s, line 127: is this comment still relevant? if so restore?

playwright/async_api/_generated.py
mxschmitt: Yeah see https://github.com/microsoft/playwright-python/pull/2642#discussion_r1837347478.

setup.py
mxschmitt: Not anymore since we always output a single wheel now instead of multiple like before.

yury-s (approved):

playwright/sync_api/_generated.py
mxschmitt: Extracted into https://github.com/microsoft/playwright-python/pull/2646.

mxschmitt: Its hard to split the changes, since `black` [starts reading `[project.requires-python]` ](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#:~:text=By%20default%2C%20Black%20will%20infer%20target%20versions%20from%20the%20project%20metadata%20in%20pyproject.toml%2C%20specifically%20the%20%5Bproject.requires%2Dpython%5D%20field.%20If%20this%20does%20not%20yield%20conclusive%20results%2C%20Black%20will%20use%20per%2Dfile%20auto%2Ddetection.) from the `pyproject.toml`. Lets land together.

Issue #2606, opened 2024-10-21
[Ports]: Backport client side changes for 1.49


playwrightmachine:
Please backport client side changes:
- [x] https://github.com/microsoft/playwright/pull/33211
- [x] https://github.com/microsoft/playwright/pull/33218
- [x] https://github.com/microsoft/playwright/pull/33208
- [x] https://github.com/microsoft/playwright/pull/33269
- [x] https://github.com/microsoft/playwright/pull/33300
- [x] https://github.com/microsoft/playwright/pull/33309
- [x] https://github.com/microsoft/playwright/pull/33338
- [x] https://github.com/microsoft/playwright/pull/33334
- [x] https://github.com/microsoft/playwright/pull/33387
- [x] https://github.com/microsoft/playwright/pull/33407
- [x] https://github.com/microsoft/playwright/pull/33081
- [x] https://github.com/microsoft/playwright/pull/33450
- [x] https://github.com/microsoft/playwright/pull/33457
- [x] https://github.com/microsoft/playwright/pull/33458
- [x] https://github.com/microsoft/playwright/pull/33409
- [x] https://github.com/microsoft/playwright/pull/33487
- [x] https://github.com/microsoft/playwright/pull/33482
- [x] https://github.com/microsoft/playwright/pull/33512
- [x] https://github.com/microsoft/playwright/pull/33522
- [x] https://github.com/microsoft/playwright/pull/33544
- [x] https://github.com/microsoft/playwright/pull/33567

Issue #1949, opened 2023-06-01
[Feature] ability group outputs in the trace viewer


scott-au:
the [show steps/fixtures in the trace](https://youtu.be/JeFD6rqDbBo?t=94) is certainly a nice feature released in 1.34 - grouping all the associated actions makes a trace file easier to go through.

is there any chance we could also have some sort of mechanism that allows us to group actions together in a similar fashion for users not using playwright test.

DetachHead:
i looked into this and would like to have a go at implementing it. it seems straightforward, looks like trace events just need the `parentId` field in the metadata.

i will probably make a `step` context manager that works the same as the `test.step` method in `@playwright/test`

EldoradoEU:
Any update ? I also looking for similar feature

DetachHead:
unfortunately it turned out to be more difficult than i thought, so i went with a very hacky solution that's tied to my own project which i can't really share.

@mxschmitt would you happen to have some insight on how i could go about implementing this? do i need to expose `TestInfo._addStep` to python? if so, how?

danielmhair:
Yes, grouping outputs is important to what I'm doing but I'm in Python, not typescript. I would love this feature!

scott-au:
> unfortunately it turned out to be more difficult than i thought, so i went with a very hacky solution that's tied to my own project which i can't really share.
> 
> @mxschmitt would you happen to have some insight on how i could go about implementing this? do i need to expose `TestInfo._addStep` to python? if so, how?

@mxschmitt any guidance here ? sounds like @DetachHead is keen to submit a PR

Snooz82:
this feature should be possible when https://github.com/microsoft/playwright/pull/33081 is merged into Playwright NodeJS.

it introduces a grouping possibility for `context.tracing`

Snooz82:
FYI: this feature has just been merged to core!🎊

Hi Dmitry (@dgozman),

is that already automatically added to Playwright-Python?

if so, i think this issue can also be closed.

Thanks
René

DetachHead:
they have a bot that automatically creates issues for porting changes from the main playwright repo. since your PR is mentioned in #2606 i guess this issue can now be closed

thanks for your work on this! i can finally delete a bunch of hacky patches from my project

PR #2660, opened 2024-11-21
chore: roll to v1.49.0

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2606
Fixes https://github.com/microsoft/playwright-python/issues/1949

### DIFF
--- a/playwright/_impl/_helper.py
+++ b/playwright/_impl/_helper.py
@@ -142,27 +142,26 @@ class FrameNavigatedEvent(TypedDict):
 Env = Dict[str, Union[str, float, bool]]
 
 
-class URLMatcher:
-    def __init__(self, base_url: Union[str, None], match: URLMatch) -> None:
-        self._callback: Optional[Callable[[str], bool]] = None
-        self._regex_obj: Optional[Pattern[str]] = None
-        if isinstance(match, str):
-            if base_url and not match.startswith("*"):
-                match = urljoin(base_url, match)
-            regex = glob_to_regex(match)
-            self._regex_obj = re.compile(regex)
-        elif isinstance(match, Pattern):
-            self._regex_obj = match
-        else:
-            self._callback = match
-        self.match = match
-
-    def matches(self, url: str) -> bool:
-        if self._callback:
-            return self._callback(url)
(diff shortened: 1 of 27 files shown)


playwright/_impl/_helper.py
```diff
@@ -142,27 +142,26 @@ class FrameNavigatedEvent(TypedDict):
...
+            and re.match(r"^wss?://", url_string)
+        ):
+            base_url = re.sub(r"^http", "ws", base_url)
+        if base_url:
+            match = urljoin(base_url, match)
+    if isinstance(match, str):
+        match = glob_to_regex(match)
+    if isinstance(match, Pattern):
+        return bool(match.search(url_string))
+    return match(url_string)
```
mxschmitt, line 164: For some reason Python does not have this code:

https://github.com/microsoft/playwright/blob/f43b86fea43aeb99cfea4a0773c577dd9801b35d/packages/playwright-core/src/utils/isomorphic/urlMatch.ts#L110-L118

The "past path" seems an optimisation while the 'parseURL path matching' seems not implemented at all?

dgozman: It seems like js has some dead code. `match` cannot be a string there because we do `match = globToRegex(match);` before.

playwright/_impl/_page.py
```diff
@@ -1274,27 +1261,15 @@ def expect_response(
...
-            else URLMatcher(
-                self._browser_context._options.get("baseURL"), urlOrPredicate
-            )
-        )
-        predicate = urlOrPredicate if callable(urlOrPredicate) else None
-
-        def my_predicate(response: Response) -> bool:
-            if matcher:
-                return matcher.matches(response.url)
-            if predicate:
```
mxschmitt, line 1289: ditto

dgozman (approved):

playwright/_impl/_page.py
mxschmitt: Addressed in https://github.com/microsoft/playwright-python/pull/2660/commits/0a44a70ae6105dd7a1c24f2174879589e59c126b.

Issue #2666, opened 2024-11-25
[Feature]: relax greenlet dependency version requirements in pyproject.toml


mistercrunch:
### 🚀 Feature Request

First! Thanks for an awesome project/library. We use playwright in [apache/superset](https://github.com/apache/superset/) and clearly it has simplified running end-to-end tests on multiple browsers and headless-browser workloads like taking screenshots / thumbs. Grateful for much sanity in the headless browser space.

I was wondering it it would make sense for you all to relax the [currently] pinned dependencies for the library. Currently playwright forces a certain version of greenlet (3.1.1) https://github.com/microsoft/playwright-python/blob/main/pyproject.toml#L16, which conflicts with other reqs in our repository, and prevents us from upgrading greenlet to a newer version while using playwright with other libs that need other versions of greenlet. 

For the CLI functionality it's not an issue, but somehow we also import the lib and use it in our modules too, so given python's lack of support for diamond dependencies it's creating a conflict.

Anyhow, the standard practice is to relax deps to ranges in your package metadata, say `greenlet>=3.1.1`, and to use `pip-compile` or `uv` to pin dependencies, usually in a `requirements.txt`, that can be used in CI to get deterministic outcomes, and can be bumped periodically. Though I understand there may be good reasons for you all to pin directly in pyproject.toml (?)

Anyhow, happy to help too. I almost opened a PR but I'm not familiar with CI scripts and other automation you may have.

### Example

`pip install playwright`

### Motivation

People will be able to use playwright in combination with other libraries that have library range restrictions for greenlet that fall outside of 3.1.1

mxschmitt:
Happy to relax it. Do you mind creating a PR? Looks like it was filed before in https://github.com/microsoft/playwright-python/issues/2190.

This requires changing the meta.yaml and pyproject.toml.

mistercrunch:
https://github.com/microsoft/playwright-python/pull/2669

PR #2669, opened 2024-11-26
feat: relax dependencies in pyproject.toml and add a requirements.txt

mistercrunch:
Related to this: https://github.com/microsoft/playwright-python/issues/2666

- made a call on using `uv` as it seems it's what the cool kids are using nowadays
- looked at `.github/workflows/` and modified so that it would install the pinned files

NOTE:
- I noticed `local-requirements.txt` file and added `uv` to it, though I'd recommend moving to `optional-dependencies` in pyproject.toml, maybe as unpinned there, and generate a requirements-dev.txt or similar from there. Though not super important
- I might have missed other workflows/actions internal or external to the repo, but generally a good practice to install the pinned deps in all CI/automation


mistercrunch:
@microsoft-github-policy-service agree [company="Preset Inc."]

requirements.txt
```diff
@@ -0,0 +1,8 @@
+# This file was autogenerated by uv via the following command:
+#    uv pip compile pyproject.toml -o requirements.txt
+greenlet==3.1.1
```
mxschmitt, line 3: What I'm worried about here is that when e.g. greenlet 3.1.2 comes out, which includes e.g. a breaking change or a bug for some reason, our CI would still run with 3.1.1 and not install the bad version.

Why is the requirements.txt file needed at all in this case? To provide reproducible builds?

pyproject.toml
```diff
@@ -13,8 +13,8 @@ license = {text = "Apache-2.0"}
 dynamic = ["version"]
 requires-python = ">=3.9"
 dependencies = [
-    "greenlet==3.1.1",
-    "pyee==12.1.1",
+    "pyee>=12.1.1",
```
mxschmitt, line 16: Make sure to relax the version in the `meta.yaml` as well for our conda customers.

mistercrunch: mmmmh yeah it's not great for this to not be DRY. I'm not familiar enough with conda to know the best practices on that side of the fence. Here's a related GPT thread discussing the different options/pros/cons -> https://chatgpt.com/share/67476ce2-0288-8010-bc05-e20ff72d15c0

mistercrunch: I don't like the fact conda doesn't seem to be compatible with dependabot. Is it a requirement to support conda? Seems it's fading https://theregister.com/2024/08/08/anaconda_puts_the_squeeze_on/

A1exKH (changes requested):
@mistercrunch could sync you branch with main, please? 
This branch is out-of-date with the base branch.

mistercrunch:
@A1exKH I just rebased the PR

pyproject.toml
mxschmitt: Yes, I'd prefer to not drop it for now. Updating it manually on conda-side is good enough for us.

local-requirements.txt
```diff
@@ -20,3 +20,4 @@ service_identity==24.2.0
 twisted==24.11.0
 types-pyOpenSSL==24.1.0.20240722
 types-requests==2.32.0.20241016
+uv==0.5.4
```
mxschmitt, line 23: I'd prefer to not introduce a new package for this. What do we loose if we don't use it? afaik its only used to update the command in the `requirements.txt`?

requirements.txt
mxschmitt: Yeah lets do something like: `greenlet>=3.1.1,<4.0.0`. For pyee something similar, so its only 12.x for now and we can increase if there is a new version at some point. Just to make sure 13.x does not break us. Typing extensions is backed by the Python team, I think usually you don't depend on a max version there..

mxschmitt (commented):
Sorry for being slow! I think once the feedback is applied we can merge it. Please also adjust meta.yml so it has the same ranges as pyproject.toml. Thanks!

local-requirements.txt
mistercrunch: I needed something to compile the requirements, it's either this `uv` or `pip-tools` or maybe `poetry` (?). Unless `conda` has options to do this but I think `conda` is slowing down, and `uv` is the future of package management in python, replacing/improving-on/consolidating `pip`, `virtualenv`, `pyenv`, `conda`, ...

mistercrunch:
Would suggest disabling that GH feature forcing to rebase, isn't scalable, forces rebasing all PRs on every merge... But happy to rebase

local-requirements.txt
mistercrunch: Here https://chatgpt.com/share/6764e629-4344-8010-8465-c1efe5e1813e

mxschmitt: I see - so both dependencies: `pyee` and `greenlet` have no dependencies on itself - this is where I got confused. uv helps us to pin the transitive dependencies (where there aren't any yet). In the future there might be some. Is this correct?

mistercrunch: that's right, generally good to have pin deps for CI otherwise a new lib comes out some day and your CI starts failing.

A1exKH (approved):
LGTM.

stollero:
Hi there @mxschmitt,

can this PR be applied ? We would like to update pyee but playwright currently blocks it 😬 Thanks a lot

pciporin25:
Would also appreciate this being merged as it's blocking a `pyee` upgrade; thank you in advance!

mxschmitt:
Fixed in https://github.com/microsoft/playwright-python/pull/2698

mistercrunch:
Hey @mxschmitt, this is not a big deal to me, but by re-implementing a similar solution and closing this PR I don't get proper attribution for my contribution on GitHub. What matters to me is having a solution in the main branch, but would recommend supporting contributors to completion as a better open source practice.

Issue #2686, opened 2024-12-12
[Bug]: page.route_web_socket doesn't intercept the ws connection


nishirken:
### Version

1.49.1

### Steps to reproduce

The test code uses a custom chromium path. TEST_BROWSER_PATH.

1. Clone https://github.com/nishirken/websocket-test
2. Install uv `pip install uv`
3. Install deps `uv sync && npm i`
4. Launch the client `python -m http.server 8000`
5. Launch the server `uv run server.py`
6. Launch the python test `uv run test_websocket.py`
7. Launch the ts test `npx ts-node test_websocket.ts`

### Expected behavior

The python and ts versions should behave the same.
The python version prints a message when the WebSocket route is established and sends a fake message to the client. (The ts version does it as expected)

### Actual behavior

No logs and no messages in the python version.

### Environment

```Text
- Operating System: Linux 6.11 NixOS 24.05 (Uakari) 24.05 (Uakari)
- CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
- Browser: Chromium
- Python Version: [3.11.10]
- Other info:
```

mxschmitt:
Thank you for the great bug report! Looks like this is about the URL matching - I created a PR with a fix in https://github.com/microsoft/playwright-python/pull/2687. As a workaround could you try this?

`await page.route_web_socket("ws://localhost:8080/", route_handler)`

nishirken:
I've installed the version from github. It works with a string URL.
But a regex url doesn't work.
I've updated the example with a get param in the url.

mxschmitt:
`r"^ws://.*"` is just a `str` in Python - if you want to use a regex, you need to use:

        await page.route_web_socket(re.compile(r"^ws://.*"), route_handler)

PR #2687, opened 2024-12-12
fix(webSocketRoute): allow no trailing slash in route matching

mxschmitt:
https://github.com/microsoft/playwright-python/issues/2686

### DIFF
--- a/playwright/_impl/_helper.py
+++ b/playwright/_impl/_helper.py
@@ -34,7 +34,7 @@ from typing import (
     Union,
     cast,
 )
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
 from playwright._impl._api_structures import NameValue
 from playwright._impl._errors import (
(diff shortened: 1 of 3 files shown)


playwright/_impl/_helper.py
```diff
@@ -157,6 +157,10 @@ def url_matches(
             base_url = re.sub(r"^http", "ws", base_url)
         if base_url:
             match = urljoin(base_url, match)
+        parsed = urlparse(match)
+        if parsed.path == "":
+            parsed = parsed._replace(path="/")
```
mxschmitt, line 162: `_replace` seems to be the recommendation as per [here](https://docs.python.org/3/library/urllib.parse.html).

yury-s (approved):
Why does it work in Node.js and other ports?

mxschmitt:
> Why does it work in Node.js and other ports?

In Node.js we do the same here: https://github.com/microsoft/playwright/blob/16a1552e7490139f295086ab52cc0d7b3901eaac/packages/playwright-core/src/utils/isomorphic/urlMatch.ts#L128

The `URL` constructor does normalise the URL and add a trailing slash.

Issue #2693, opened 2024-12-20
[Feature]: Dependency Updates


Houbein:
### 🚀 Feature Request

Update Python Dependencies. 
Currently Playwright Python requires as a dependency pyee 12.0.0. This is now 4 months old and pyee 12.1.1 is the newest available version, please update this. 
Even better, please be less strict. Playwright currently demand pyee 12.0.0 down to the minor version, this is really not good practise and this version lock should be restricted to major versions only. 

### Motivation

Locking down even the minor version is really not good practise for bug and security reasons.

mxschmitt:
Thank you for filing this issue. We are aware on the issue and trying to solve it as part of https://github.com/microsoft/playwright-python/issues/2666.

Issue #2190, opened 2023-12-11
[BUG] playwright 1.40.0 is strictly depend on pyee==11.0.1


vzhirnov:
### System info

- Playwright Version: 1.40.0
- Operating System: macos monterey  12.2.1
(Apple M1)

- Other info:

pyee==11.0.1

**Steps**
Try to update pyee from version 11.0.1 to 11.1.0

**Expected**

update pyee from version 11.0.1 to 11.1.0 is completed successfully.

**Actual**

We cannot install new pyee version, because
```
The conflict is caused by:
    The user requested pyee==11.1.0
    playwright 1.40.0 depends on pyee==11.0.1
```

Is that the correct behaviour? Do we need use pyee with only 11.0.1 version and not above?

vzhirnov:
The issue is updated. I'm sorry, I initially specified the wrong version of the operating system.

RazerM:
This is a general issue I'm facing too where greenlet is also pinned. This seems needlessly restrictive.

The commits that bump pyee and greenlet have no code changes, so there doesn't seem to be a compatibility reason (and even then, ranges like `'greenlet>=3.0.0,<4,!=3.0.1'` would be better)

christopher-watanabe-snkeos:
On a related note, I recently noticed failing tests due to [AsyncIOEventEmitter deprecation in pyee v12.0.0](https://github.com/jfhbrook/pyee/releases/tag/v12.0.0). I was using a the much older v1.13.1 playwright and was not pinning the dependencies.

I updated and pinned the versions as follows: 
playwright == 1.47.0
pyee == 12.0.0
greenlet == 3.0.3

And my tests ran as expected.

mborsetti:
Any chance you can stop pinning so we don't have to live in dependency (and installation) hell?

Also see #2564 for more pinning woes

PR #2698, opened 2024-12-27
chore: relax dependency versions

mxschmitt:
Motivation: Over the last few years it came up multiple times that its more common in the Python ecosystem to not pin to a specific version of a dependency. This PR aims to relax this, since this was requested in https://github.com/microsoft/playwright-python/issues/2693, https://github.com/microsoft/playwright-python/issues/2190, https://github.com/microsoft/playwright-python/issues/2666 to change that.

- This will keep a lower/upper range so that future versions of e.g. `pyee` won't break us as long as they are using semver correctly.
- `uv` is used to `compile` the dependencies and create a lock file. So our builds are reproducible.
- Dependabot will keep bumping the prod deps in `requirements.txt` so that our CI will run the tests on them.
- This PR supersedes https://github.com/microsoft/playwright-python/issues/2666


A1exKH (approved):
LGTM.

dgozman (approved):

PR #2719, opened 2025-01-20
Run build GitHub Action on an ARM processor

cclauss:
```diff
+        - os: ubuntu-24.04-arm
+          python-version: '3.13'
+          browser: chromium
```
[Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)
```diff
-      if: matrix.os == 'ubuntu-latest'
+      if: runner.os == 'Linux'
```
[Detecting the operating system](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#detecting-the-operating-system)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,9 @@ jobs:
         - os: ubuntu-latest
           python-version: '3.13'
           browser: chromium
+        - os: ubuntu-24.04-arm
+          python-version: '3.13'
+          browser: chromium
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


mxschmitt:
I apologise for the late review / reply. We have some internal infrastructure which already tested linux-arm64. As part of https://github.com/microsoft/playwright-python/pull/2837 we will move this to this repository so we don't need the private infrastructure anymore. I'll close it for now since we'll continue in https://github.com/microsoft/playwright-python/pull/2837 with it. Thank you for bringing this up!

Cool trick with `runner.os` - I didn't know about that!

Issue #1272, opened 2022-04-25
Can we please get soft assertions added to the python tools


heathiviesyncro:
It would be nice to get the soft assertions that are available in node to be available in python.

https://playwright.dev/docs/test-assertions#soft-assertions

Thank you

aspenboy:
I am using `pytest-check` for soft assertions. Adds another dependency to the project, but it works great.

jfp1992:
> pytest-check

I tried this one but couldn't get it working with expect(). Do you have an example of this?

Also +1 for soft assertions to be implemented in python as well as typescript.

aspenboy:
> I tried this one but couldn't get it working with expect(). Do you have an example of this?

It's a separate assertion mechanism, so you don't use it within `expect()`. Pretty nice examples are included on the github page: https://github.com/okken/pytest-check

Also - you can mix assertions in your tests, so it's fine to use `assert`, `expect()` and `check()`.

mbroton:
I agree, it would be really useful to have soft assertions in Python version of Playwright. Right now I'm using workarounds in some of my projects, similar to what was already mentioned above.

I spent some time exploring the code and came up with an approach that seems to work. Here’s an example of how it looks:
```python
def test_foo(page: Page) -> None:
    # ...
    with soft_assertions() as expect:
        expect(page.locator("div")).to_have_text("foo")
        expect(page.locator("div")).to_have_text("bar")
```
`expect` in this example works just like a normal `expect` (works with pages, locators and api responses).

At first, I considered an `expect.soft(...)` approach like in Node, but I couldn’t find a solution that didn’t require a lot of changes.

Would love any feedback on this!

Skn0tt:
Let's also try out implementing it via a fixture in our pytest integration.

PR #2750, opened 2025-02-18
Implement soft assertions

mbroton:
Related to #1272

This PR adds support for soft assertions, allowing multiple assertion failures to be collected before raising them together. I implemented this using a context manager pattern since it feels more natural and Pythonic:

```python
with expect.soft() as soft_expect:
    await soft_expect(page).to_have_title("wrong title")
    await soft_expect(page.locator("#div1")).to_have_text("wrong")
    response = await page.request.get("...")
    await soft_expect(response).to_be_ok()
```

Example error report:

```
>       with expect.soft() as soft_expect:
E       AssertionError: Soft assertion failures
E       1. Page title expected to be 'wrong title'
E       Actual value: Page Title
E       Call log:
E         - PageAssertions.to_have_title with timeout 5000ms
E         -   - waiting for locator(":root")
E         -     9 × locator resolved to <html>…</html>
E         -       - unexpected value "Page Title"
E
E       2. Locator expected to have text 'wrong'
E       Actual value: Text1
E       Call log:
E         - LocatorAssertions.to_have_text with timeout 5000ms
E         -   - waiting for locator("#div1")
E         -     9 × locator resolved to <div id="div1">Text1</div>
E         -       - unexpected value "Text1"
E
E       3. Response status expected to be within [200..299] range, was '404'
E       Call log:
E       → GET http://localhost:2137/non-existent
E         -   user-agent: (...)
E         -   accept: */*
E         -   accept-encoding: gzip,deflate,br
E         - ← 404 Not Found
E         -   content-type: text/plain
E         -   transfer-encoding: chunked
```

## Implementation Details

- The implementation maintains backward compatibility with all existing assertion methods
- Both sync and async APIs are fully supported with identical behavior
- Error messages preserve the order of failures and include clear numbering (1., 2., etc.)
- Custom error messages are properly handled and included in the final error report
- The implementation handles nested errors correctly (when other exceptions occur within the soft assertion block)
- Timeout settings are respected both at the global and individual assertion level

Initially, I used `contextlib.contextmanager` for `expect.soft()` as it was a simpler implementation. However, this approach included some unrelevant information in the traceback when assertions failed. To address this, I created a custom `SoftAssertionContextManager` class that gives us more control over the error handling and traceback presentation.

This is just one way to add soft assertions in Python Playwright. I think the context manager pattern works well with Python, but I’m totally open to other ideas if they fit the project better.

### DIFF
--- a/playwright/_impl/_assertions.py
+++ b/playwright/_impl/_assertions.py
@@ -1111,14 +1111,13 @@ class SoftAssertionContextManager(Generic[E]):
         __tracebackhide__ = True
 
         if self._context.has_failures():
-            if exc_type is not None:
+            if exc_type is not None and exc_val is not None:
                 failure_message = (
                     f"{str(exc_val)}"
                     f"\n\nThe above exception occurred within soft assertion block."
                     f"\n\nSoft assertion failures:\n{self._context.get_failure_messages()}"
                 )
-                if exc_val is not None:
-                    exc_val.args = (failure_message,) + exc_val.args[1:]
+                exc_val.args = (failure_message,) + exc_val.args[1:]
                 return
 
             raise AssertionError(
--- a/playwright/_impl/_assertions.py
+++ b/playwright/_impl/_assertions.py
@@ -13,7 +13,19 @@
 # limitations under the License.
 
 import collections.abc
-from typing import Any, List, Optional, Pattern, Sequence, Union
+from types import TracebackType
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Generic,
+    List,
+    Optional,
+    Pattern,
+    Sequence,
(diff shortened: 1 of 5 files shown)


mbroton:

mbroton:
I reviewed the failing checks and also tried to reproduce some of them locally without success. I couldn't find any indication that these failures are connected to my changes, so it seems they might be flaky or not related? Please let me know if you think further investigation is needed on my end.

mbroton:
Just wanted to follow up on this PR - happy to address any feedback or make adjustments if needed.

Skn0tt:
Implemented in https://github.com/microsoft/playwright-python/pull/3065

PR #2759, opened 2025-03-02
Ensure send() method uses drain() to handle writes correctly

allrob23:
This PR updates the `send()` method to ensure it properly queues data when a write operation fails and uses `drain()` to handle writes efficiently.

### Changes  
- Added `await self._output.drain()` after writing data to ensure the write operation is properly flushed.  
- This prevents potential issues where data might not be sent immediately, aligning with the expected `asyncio` usage pattern.

I opened this PR because the [asyncio documentation](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamWriter.write) recommends using `drain()` when writing to a stream. This ensures that writes are properly handled without blocking or data loss, making it a best practice for managing asynchronous I/O.

### DIFF
--- a/playwright/_impl/_transport.py
+++ b/playwright/_impl/_transport.py
@@ -176,3 +176,4 @@ class PipeTransport(Transport):
         self._output.write(
             len(data).to_bytes(4, byteorder="little", signed=False) + data
         )
+        await self._output.drain()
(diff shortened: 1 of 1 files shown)


mxschmitt:
Looks like all the bots are failing.

allrob23:
I updated the change

mxschmitt:
I don't think thats how async methods supposed to be called:

  /Users/runner/work/playwright-python/playwright-python/playwright/_impl/_transport.py:179: RuntimeWarning: coroutine 'StreamWriter.drain' was never awaited
    self._output.drain()

I'll close it for now since it doesn't seem to fix a real word issue users run into. Feel free to revisit if you have a specific repro which is broken.

Issue #2717, opened 2025-01-17
[Ports]: Backport client side changes for 1.51


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/pull/34379
  - [x] https://github.com/microsoft/playwright/pull/34458
  - [x] https://github.com/microsoft/playwright/pull/34413
  - [x] https://github.com/microsoft/playwright/pull/34471
  - [x] https://github.com/microsoft/playwright/pull/34503
  - [x] https://github.com/microsoft/playwright/pull/34510
  - [x] https://github.com/microsoft/playwright/pull/34537
  - [x] https://github.com/microsoft/playwright/pull/34550
  - [x] https://github.com/microsoft/playwright/pull/34560
  - [x] https://github.com/microsoft/playwright/pull/34560
  - [x] https://github.com/microsoft/playwright/pull/34560
  - [x] https://github.com/microsoft/playwright/pull/34600
  - [x] https://github.com/microsoft/playwright/pull/34494
  - [x] https://github.com/microsoft/playwright/pull/34591
  - [x] https://github.com/microsoft/playwright/pull/34614
  - [x] https://github.com/microsoft/playwright/pull/34630
  - [x] https://github.com/microsoft/playwright/pull/34649
  - [ ] https://github.com/microsoft/playwright/pull/34654
  - [x] https://github.com/microsoft/playwright/pull/34650
  - [x] https://github.com/microsoft/playwright/pull/34611
  - [x] https://github.com/microsoft/playwright/pull/34680
  - [x] https://github.com/microsoft/playwright/pull/34681
  - [x] https://github.com/microsoft/playwright/pull/34682
  - [x] https://github.com/microsoft/playwright/pull/34705
  - [x] https://github.com/microsoft/playwright/pull/34683
  - [x] https://github.com/microsoft/playwright/pull/34690
  - [x] https://github.com/microsoft/playwright/pull/34689
  - [x] https://github.com/microsoft/playwright/pull/34718
  - [ ] https://github.com/microsoft/playwright/pull/34710
  - [x] https://github.com/microsoft/playwright/pull/34346
  - [ ] https://github.com/microsoft/playwright/pull/34730
  - [x] https://github.com/microsoft/playwright/pull/34739
  - [x] https://github.com/microsoft/playwright/pull/34759
  - [x] https://github.com/microsoft/playwright/pull/34763
  - [x] https://github.com/microsoft/playwright/pull/34766
[...]

PR #2767, opened 2025-03-05
chore(roll): roll Playwright to 1.51.0-beta-1741166263000

mxschmitt:
Closes https://github.com/microsoft/playwright-python/issues/2717

### DIFF
--- a/playwright/_impl/_api_structures.py
+++ b/playwright/_impl/_api_structures.py
@@ -72,9 +72,10 @@ class LocalStorageEntry(TypedDict):
     value: str
 
 
-class OriginState(TypedDict):
+class OriginState(TypedDict, total=False):
     origin: str
     localStorage: List[LocalStorageEntry]
+    indexedDB: Optional[List[Any]]
 
 
 class PdfMargins(TypedDict, total=False):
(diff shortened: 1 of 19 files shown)


playwright/_impl/_api_structures.py
```diff
@@ -72,9 +72,10 @@ class LocalStorageEntry(TypedDict):
     value: str
 
 
-class OriginState(TypedDict):
+class OriginState(TypedDict, total=False):
```
dgozman, line 75: Undo this one now :)

playwright/_impl/_browser_context.py
```diff
@@ -599,8 +599,12 @@ async def _inner_close() -> None:
         await self._channel.send("close", {"reason": reason})
         await self._closed_future
 
-    async def storage_state(self, path: Union[str, Path] = None) -> StorageState:
-        result = await self._channel.send_return_as_dict("storageState")
+    async def storage_state(
+        self, path: Union[str, Path] = None, indexedDB: bool = None
```
dgozman, line 603: why not `indexed_db`?

dgozman (approved):

playwright/_impl/_browser_context.py
mxschmitt: Our generator layer has some magical logic which maps `indexed_db` (user consuming API) to `indexedDB` (driver). The reason for this is that we like to use the `protocol.yaml` compatible property names in the implementation so we can simply pass `locals()` to the driver.

PR #2772, opened 2025-03-10
test: migrate flaky plugin to rerunfailures

mxschmitt:
This PR changes our assumption from marking individual tests as flaky to marking the whole test run as flaky and retrying all the tests if needed. It replaces the flaky plugin with rerunfailures plugin (maintained by the pytest org) since its not compatible [with each-other](https://pypi.org/project/pytest-rerunfailures/#compatibility). This aligns it how we do it in Node.js.

It enables 3 retries only for CI runs.

### DIFF
--- a/.github/workflows/test_docker.yml
+++ b/.github/workflows/test_docker.yml
@@ -49,5 +49,5 @@ jobs:
         docker exec "${CONTAINER_ID}" pip install -r requirements.txt
         docker exec "${CONTAINER_ID}" pip install -e .
         docker exec "${CONTAINER_ID}" python -m build --wheel
-        docker exec "${CONTAINER_ID}" xvfb-run pytest -vv tests/sync/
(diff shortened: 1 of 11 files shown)


.github/workflows/test_docker.yml
```diff
@@ -49,5 +49,5 @@ jobs:
         docker exec "${CONTAINER_ID}" pip install -r requirements.txt
         docker exec "${CONTAINER_ID}" pip install -e .
         docker exec "${CONTAINER_ID}" python -m build --wheel
-        docker exec "${CONTAINER_ID}" xvfb-run pytest -vv tests/sync/
-        docker exec "${CONTAINER_ID}" xvfb-run pytest -vv tests/async/
+        docker exec "${CONTAINER_ID}" xvfb-run pytest tests/sync/
```
mxschmitt, line 52: drive-by: we set `-vv` already in our `setup.cfg`.

PR #2800, opened 2025-03-25
chore: publish win32-arm64 wheel

mxschmitt:
What is it solving? Before this change Playwright for Python was not able to run on `win32-arm64` since we don't publish a wheel on that platform. The install via PIP was failing. In Node.js its already working since Windows emulates all the arm64 instructions. This patch adds a new wheel using a dedicated `win32-arm64` driver (contains also win32-arm64 Node.js) in order to make Playwright work there.

Relates https://github.com/microsoft/playwright/pull/35185

This requires a new roll to make the Docker tests pass.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,12 @@ base_wheel_bundles = [
         "platform": "win32",
         "zip_name": "win32_x64",
     },
+    {
+        "wheel": "win_arm64.whl",
+        "machine": "arm64",
+        "platform": "win32",
+        "zip_name": "win32_arm64",
+    },
 ]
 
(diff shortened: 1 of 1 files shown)


dgozman (approved):

es0m-msft:
confirmed successful creation and subsequent installation using 
python setup.py bdist_wheel 
pip install dist\playwright...whl
playwright install chromium
and then running the sample:

with sync_playwright() as p:
    for browser_type in [p.chromium]:
        page = browser.new_page()
        page.goto('http://playwright.dev')
        browser.close()

Issue #2744, opened 2025-02-15
[Feature]: Option to disable `inspect.stack()` calls for performance optimization


jl-martins:
### Summary

playwright-python currently makes frequent calls to [inspect.stack](https://docs.python.org/3/library/inspect.html#inspect.stack), which significantly impacts performance in certain use cases, such as web scraping with [scrapy-playwright](https://github.com/scrapy-plugins/scrapy-playwright). This feature request proposes adding an option to disable stack inspections to improve execution speed when debugging information is not required.

### Use case

Although Playwright is primarily designed for end-to-end testing, some projects leverage it for web scraping, where performance is a key concern. In my case, I am using [scrapy-playwright](https://github.com/scrapy-plugins/scrapy-playwright), and profiling results from [cProfile](https://docs.python.org/3/library/profile.html) indicate that `inspect.stack` calls contribute to roughly 25% of the total execution time, as illustrated in the following icicle and call graph:

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/57309c9d-e7c3-4859-b444-3104c232e15b" />

![Image](https://github.com/user-attachments/assets/b44736b4-eff8-4c92-a6cf-c169acb585f9)

As far as I can tell, these calls are primarily used for debugging and are not critical for normal execution.

### Proposed solution

Introduce an environment variable (e.g.: PW_INSPECT_STACK) that allows users to disable stack inspections when debugging is not required. The default value would be 1 to preserve the current behavior.

### Motivation

* **Performance Boost**: Reducing unnecessary function calls can significantly speed up Playwright for scraping-heavy applications.
* **Flexibility**: Users who require stack traces for debugging can leave the option enabled.
* **Backward Compatibility**: The default behavior remains unchanged, ensuring no disruption for existing users.

luisferreira93:
+1

mxschmitt:
Do you have in comparison how much time the actual page.goto/click/fill etc's took? Lets say 1s in average with network resources around different origins and sometimes even iframes etc. This **should** be way larger and demonstrate that its not worth investing into that end (inspector.stack).

neoncube2:
I think I might be affected by this, too.

In my case, I load a page, check to make sure that a certain amount of buttons are visible, then execute a few thousand `page.evaluate()` calls in a tight loop.

**Profiling results**

![Image](https://github.com/user-attachments/assets/854979e7-aae8-4138-8585-7cc7bc3005b1)

**Use case**

[...]

PR #2835, opened 2025-04-29
_connect.py: Don't request context when inspecting stack

neoncube2:
This PR partially fixes https://github.com/microsoft/playwright-python/issues/2744

This PR makes two changes:
1. Call `inspect.stack(0)` instead of `inspect.stack()`
2. Only call `inspect.stack()` when `getattr(task, "__pw_stack__")` doesn't find a stack in the task

`inspect.stack()` by default requests context for the first frame of the callstack, which is a [very expensive operation](https://stackoverflow.com/a/17407257/132540). We don't use the returned context information, so we can safely use `inspect.stack(context=0)`.

In my testing, this improved the speed of `evaluate()` by about 2.5x:

**Test:**
```py
async def test_speed_test(page: Page) -> None:
    start_time = time.time()

    for i in range(0, 1000):
        await page.evaluate('() => 1 + 1')

    print(time.time() - start_time)
```

Before this PR: 10s
After this PR: 4s

The second change is that we originally were calling `getattr(task, "__pw_stack__", inspect.stack())`, which retrieved the entire callstack and then threw it away if `getattr(task, "__pw_stack__")` succeeded. I wasn't able to figure out how to get `getattr(task, "__pw_stack__")` to return a value, so I'm not sure how much this change sped things up, but it seems like a sensible change to make.

I've added async and sync tests.

I have a more detailed fix, but I figured I'd first try to get this PR landed, since it involves small changes for a pretty large speedup :)

### DIFF
--- a/local-requirements.txt
+++ b/local-requirements.txt
@@ -1,7 +1,7 @@
 autobahn==23.1.2
 black==25.1.0
 build==1.2.2.post1
-flake8==7.1.2
+flake8==7.2.0
 mypy==1.15.0
 objgraph==3.6.2
 Pillow==11.1.0
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -533,7 +533,9 @@ class Connection(EventEmitter):
         if self._api_zone.get():
             return cb()
         task = asyncio.current_task(self._loop)
-        st: List[inspect.FrameInfo] = getattr(task, "__pw_stack__", inspect.stack())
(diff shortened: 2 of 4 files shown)


neoncube2:

neoncube2:
I have a feeling that this might noticeably improve the runtime of the test suite in `tests/`, too! :D

neoncube2:
Ready to have the workflows rerun :)

mxschmitt (approved):
Looks reasonable, thank you for the contribution!

PR #2836, opened 2025-04-30
Even faster stack traces for _connection.py

### LINKED ISSUE #2744: [Feature]: Option to disable `inspect.stack()` calls for performance optimization

neoncube2:
This PR optimizes `_connection.py` even further. This further fixes https://github.com/microsoft/playwright-python/issues/2744.

In the current code, we:
1. Retrieve the entire stack, by calling `getattr(task, "__pw_stack__", inspect.stack(0))`
2. Parse the entire stack
3. Throw away all of the parsed stack except for the first relevant frame, unless we're tracing (e.g. unless `self._tracing_count > 0`)

This PR makes it so that we only retrieve and parse as much of the stack as necessary.

Test code:

```python
async def test_speed_test(page: Page) -> None:
    start_time = time.time()

    for i in range(0, 1000):
        await page.evaluate('() => 1 + 1')

    print(time.time() - start_time)
```

Before this PR: 4s
With this PR: 1.75s

Speed improvement: 2.3x

I'd like add tests for when `self._is_internal_type = True`, but I'm not sure the best way to go about constructing an internal type and then triggering an exception in it. (Any suggestions?)

Similarly, I'd like to add tests for when `getattr(task, "__pw_stack__")` returns a value, but again, I'm not sure when this occurs. If anyone has suggestions, please let me know! :)


neoncube2:
Ready to rerun tests

neoncube2:
Probably best to wait until https://github.com/microsoft/playwright-python/pull/2840 is merged, and then I can run checks locally.

playwright/_impl/_connection.py
```diff
@@ -508,17 +510,44 @@ def _replace_guids_with_channels(self, payload: Any) -> Any:
             return result
         return payload
 
+    def needs_full_stack_trace(self) -> bool:
+        return self._tracing_count > 0
+
+    def get_frame_info(self) -> Iterator[FrameInfo]:
```
mxschmitt, line 516: Just thinking out loud - why does it need to be an Iterator in order to get the faster stacktraces? I assume that we call it only when we actually need it in line 546 and 566? 

Also should we maybe use this similar function in playwright/_impl/_sync_base.pyplaywright/_impl/_sync_base.py:108` ? Then sync users - what the majority is - would benefit as well?

neoncube2: As to your question: In the original code, when we were calling `inspect.stack(0)`, it would immediately walk the entire stack (essentially, it would call `frame.f_back` in a loop until there were no more frames). This is pretty expensive. By using an iterator, we delay walking the stack until we need it, and then we can only walk part of the stack (because  `frame.f_back` is only called when we need another frame)

Basically: Calling `frame.f_back` is more expensive than one might think.

Thanks for the heads up about `playwright/_impl/_sync_base.pyplaywright/_impl/_sync_base.py:108`! :)  No wonder I couldn't find any information about  `__pw_stack__`... I thought it was being set by `asyncio`, but it's actually being set by this package! XD

Let me take a look.

neoncube2:
@mxschmitt I'm still working on this, but I've split off a few simple and easy to review optimizations into https://github.com/microsoft/playwright-python/pull/2844

neoncube2:
@mxschmitt While continuing to work on this, I've discovered something interesting.

1. In `_sync_base.py`, we set `__pw_stack__` and `__pw_stack_trace__`
2. In `_connection.py`, we check to see if the current task has `__pw_stack__` or `__pw_stack_trace__` set, and if not, we generate a new stack trace to use.

Interetingly, for all of the tests in the `tests\` folder, `__pw_stack__` and `__pw_stack_trace__` are `None` whenever they're retrieved in `_connection.py`. This makes me wonder if we should stop setting `__pw_stack__` and `__pw_stack_trace__` in `_sync_base.py` (because these values are never being used) or if we just need to better tests for this.

Any thoughts?

mxschmitt:
I wonder when it was introduced if it was any different back then: https://github.com/microsoft/playwright-python/commit/c05cad257a026d450644aa3baae6f9d81bd1b3ad#diff-291261f4050ff453e2c661edf96c77af65a779b75815708d924809ee49d5dc64 - would be great to find out if it was used back in the days and not anymore or if our tests don't have the coverage.

neoncube2:
@mxschmitt Similarly, it'd be nice to know if sending `apiName` and `internal` as part of the metadata (https://github.com/microsoft/playwright-python/blob/e87e340fae92cb52f98dacdd80bb55c04e9bc4e3/playwright/_impl/_connection.py#L351) is necessary or was just added because it was nice to have :)

Eliminating that should speed things up _drastically_, I think.

PR #2844, opened 2025-05-08
More stack-related performance improvements

neoncube2:
Building off of https://github.com/microsoft/playwright-python/pull/2835, this PR replaces the remaining instances of `stack()` with `stack(0)`, so that we avoid getting context information for the first frame of the stack, which is expensive.

I also discovered that we were calling `traceback.format_list(traceback.extract_stack()[-10:])`, which does the following:
1. Extract the entire stack
2. Format the entire stack into strings
3. Throw away everything except for the last 10 formatted frames of the stack

To fix this, I've made `extract_stack()` use the `limit=10` parameter.

Finally, I optimized `_path_utils.py` to only retrieve two frames from the stack, instead of retrieving the entire stack and then discarding everything but the second frame :)

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -387,9 +387,7 @@ class Connection(EventEmitter):
                 parsed_error = parse_error(
                     error["error"], format_call_log(msg.get("log"))  # type: ignore
                 )
-                parsed_error._stack = "".join(
-                    traceback.format_list(callback.stack_trace)
-                )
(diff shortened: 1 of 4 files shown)


mxschmitt (approved):

neoncube2:
@mxschmitt Sorry, I accidentally closed and reopened the pull request '-_-

neoncube2:
@mxschmitt Woohoo, thanks! ^_^

PR #2951, opened 2025-08-15
fix: screenshot type inferred from path file extension

xianghuzhao:
The screenshot methods for both element and page do not really infer file type from path file extension if `type` is not provided. I tried some name like 'xxx.jpeg' without `type` in arguments, the generated file format was actually png.
This pull request is trying to fix this.

### DIFF
--- a/playwright/_impl/_element_handle.py
+++ b/playwright/_impl/_element_handle.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import base64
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):
Looks great! Any chance for a test? Ideally in `tests/async/test_screenshot.py`.

mxschmitt (approved):

Issue #3005, opened 2025-11-18
[Bug]: ignore_default_args is broken


Devpmin:
### Version

1.48.0

### Steps to reproduce

when launching browser, (checked on firefox, chrome, chromium) with ignore_default_args=True or False, getting error 
playwright._impl._errors.Error: BrowserType.launch: ignore_default_args: expected array, got boolean

### Expected behavior

it was supposed to disable all default args set by playwright on the browser launch, but its not accepting booleans. only list or arrays are accepted

### Actual behavior

-> browser = await browser_type.launch(**launch_args)
(Pdb) launch_args
{'headless': False, 'channel': 'chrome', 'proxy': {'server': '<servver-ip'}, 'args': BoxList([]), 'chromium_sandbox': False, 'firefox_user_prefs': Box({}), 'slow_mo': 0.0, 'timeout': 30000.0, 'ignore_default_args': False}
(Pdb) n
playwright._impl._errors.Error: BrowserType.launch: ignore_default_args: expected array, got boolean

### Additional context

in the documentation of playwright for python, its said that bool or list is accepted as argument of ignore_default_args. but its returning an error on providing bool. also there is no other option to avoid all args given by default.
https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-option-ignore-default-args

### Environment

```Text
- Operating System: [Ubuntu 22.04]
- CPU: [arm64]
- Browser: [Firefox, Chromium, Firefox]
- Python Version: [3.8,3.9,3.10]
- Other info:
```

PR #3020, opened 2025-12-27
Fix handling of ignore_default_args when passed as a boolean.

lp07:
Problem:
Passing ignore_default_args=False was sent over the wire as a boolean, but the server expects an array for ignoreDefaultArgs, resulting in:

"expected array, got boolean".

Fix:
- ignore_default_args=True → send ignoreAllDefaultArgs=True
- ignore_default_args=False → do not send ignoreDefaultArgs at all
- ignore_default_args=list → unchanged

Tests:
Added unit tests covering bool True/False and list cases across
chromium, firefox, and webkit.

Related issue: #3005

### DIFF
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -356,6 +356,8 @@ def normalize_launch_params(params: Dict) -> None:
         if params["ignoreDefaultArgs"] is True:
             params["ignoreAllDefaultArgs"] = True
             del params["ignoreDefaultArgs"]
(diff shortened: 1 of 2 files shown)


Skn0tt (approved):

Skn0tt (changes requested):
The tests seem to not be in the right place. I'll take a look later.

Skn0tt:
merged via https://github.com/microsoft/playwright-python/pull/3058

Issue #2784, opened 2025-03-11
[Bug]: APIRequestContext multipart parameter does not allow user to send multiple files in a single field


adriangardiner-taina:
### Version

1.50.0

### Steps to reproduce

Setup a multipart data object with an array of files being sent in a single field using the ApiRequestContext object

```python
multipart_data = {
    files= [ 
        { "name"="Document1.txt", "mimeType":"text/plain", buffer: b"Text file 1 content" },
        { "name"="Document2.txt", "mimeType":"text/plain", buffer: b"Text file 2 content" }
    ]
}
``` 

Attempt to send this to an API endpoint that expects to find a list of files against the field named "files".

Example steps : TBC

### Expected behavior

I expect to receive a list of files in my API endpoint in a single form field.

### Actual behavior

0 files are sent to the API Endpoint.

### Additional context

Issue has been identified in [playwright/_impl/_fetch.py#L382](https://github.com/microsoft/playwright-python/blob/25fb05dd18012e2ed42d3c8470d954581c1674ad/playwright/_impl/_fetch.py#L382)

Suggest an extra helper method and foreach loop added to identify and attach arrays of file payloads

### Environment

```Text
- Operating System: [Windows 10]
- CPU: [Intel i7]
- Browser: [All]
- Python Version: [3.11.6]
- Other info:
```

Skn0tt:
Merging into https://github.com/microsoft/playwright-python/issues/2834.

PR #3029, opened 2026-01-30
Add support for string arrays in multipart fields

Zcodeoooo:
Version
1.57

Steps to reproduce
An extension to the issue
https://github.com/microsoft/playwright-python/issues/2784

Adds support for string arrays in `multipart` fields, allowing usage like:

```python
multipart = {
    "codes[]": ["C34376202", "C481866"],
    "files"= [ 
        { "name"="Document1.txt", "mimeType":"text/plain", buffer: b"Text file 1 content" },
        { "name"="Document2.txt", "mimeType":"text/plain", buffer: b"Text file 2 content" }
    ]
}


Previously, only single values (str/file) were supported due to type and logic limitations.
Changes made
✅ Updated multipart type hint to include List[Union[str, FilePayload]]
✅ Enhanced handling of list values in form field processing
✅ Improved is_file_array_payload to correctly identify file arrays
✅ Added support for multiple same-named fields (e.g. codes[]) via list expansion
Why this is useful
Enables sending arrays of strings (e.g. batch codes, IDs) in multipart requests — a common use case when interacting with web forms or APIs expecting repeated keys.
This aligns with real-world HTTP behavior where field[] syntax is widely used for array submission.

### DIFF
--- a/playwright/_impl/_fetch.py
+++ b/playwright/_impl/_fetch.py
@@ -40,6 +40,7 @@ from playwright._impl._helper import (
     async_readfile,
     async_writefile,
     is_file_payload,
+    is_file_array_payload,
     locals_to_params,
     object_to_array,
     to_impl,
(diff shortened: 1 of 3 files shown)


Zcodeoooo:

Skn0tt:
Let's instead have a FormData class.

PR #3033, opened 2026-02-11
fix: show not_ prefix in call log when negative assertions fail

scitech:
When a negative assertion like `not_to_be_visible` fails, the call log prints the name of the **positive** assertion, which contradicts the error message and is kind of confusing if you're reading quickly. For example:
```
AssertionError: Locator expected not to be visible
Actual value: visible 
Call log:
  - Expect "to_be_visible" with timeout 3000ms
  - waiting for locator("#some-id").first
    7 × locator resolved to <svg id="some-id">…</svg>
      - unexpected value "visible"
```

With this PR, it will say `not_to_be_visible`, like so:
```
AssertionError: Locator expected not to be visible
Actual value: visible 
Call log:
  - Expect "not_to_be_visible" with timeout 3000ms
  - waiting for locator("#some-id").first
    7 × locator resolved to <svg id="some-id">…</svg>
      - unexpected value "visible"
```

### DIFF
--- a/tests/async/test_assertions.py
+++ b/tests/async/test_assertions.py
@@ -1023,6 +1023,15 @@ async def test_should_be_able_to_set_custom_timeout(page: Page) -> None:
     assert 'Expect "to_be_visible" with timeout 111ms' in str(exc_info.value)
 
 
+async def test_negative_assertion_should_show_not_prefix_in_call_log(
+    page: Page,
+) -> None:
+    await page.set_content("<button>hello</button>")
+    with pytest.raises(AssertionError) as exc_info:
+        await expect(page.locator("button")).not_to_be_visible(timeout=111)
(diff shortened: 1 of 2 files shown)


scitech:

Skn0tt (approved):
Code looks good, but i'm unsure why CI did not run 🤔

tests/async/test_assertions.py
```diff
@@ -1023,6 +1023,15 @@ async def test_should_be_able_to_set_custom_timeout(page: Page) -> None:
     assert 'Expect "to_be_visible" with timeout 111ms' in str(exc_info.value)
 
 
+async def test_negative_assertion_should_show_not_prefix_in_call_log(
```
Skn0tt:
```suggestion
async def test_negative_assertion_shows_not_prefix_in_call_log(
```

let's see if a commit kicks off CI!

Issue #2977, opened 2025-09-16
[Bug]: Object leak in Tracing


TomasBernotenas:
### Version

1.55.0

### Steps to reproduce

Example steps (replace with your own):
1. Clone my repo at (https://github.com/TomasBernotenas/playwright_tracing_leak)
2. pip install -r requirements.txt
3. python memleak_test.py
4. you should see memory and object count increasing after each loop

if tracing is turned off then the memory stays stable and there are no additional objects created after each loop

### Expected behavior

Object count and memory should return to initial after tracing is stopped

### Actual behavior

When tracing is stopped callback objects keep accumulating

### Environment

```Text
- Operating System: [Windows 11 10.0.22631]
- CPU: [(8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz]
- Browser: [All, Chromium, Firefox, WebKit]
- Python Version: [3.11.9]
```

Alphaboy20023:
I ran a test with and without tracing across 30 iterations.
Both scenarios show a small memory increase in the first few iterations (stabilizing after ~39 MB).
No continuous growth is observed after stabilization.
This suggests normal caching behavior rather than a memory leak.

Here’s a comparison chart for reference: 

<img width="1600" height="1000" alt="Image" src="https://github.com/user-attachments/assets/e01c539d-3770-4940-94ab-8c00d6d5c304" />

`import psutil
import os
from playwright.sync_api import sync_playwright

def memory_usage():
    process = psutil.Process(os.getpid())
    return process.memory_info().rss / (1024 * 1024)  # MB

def run_test(use_tracing=False, iterations=30):
    with sync_playwright() as p:
        for i in range(1, iterations + 1):
            browser = p.chromium.launch()
            context = browser.new_context()

            if use_tracing:
                context.tracing.start(screenshots=True, snapshots=True, sources=True)

            page = context.new_page()
            page.goto("https://example.com")

            if use_tracing:
                context.tracing.stop(path=f"trace_{i}.zip")

            browser.close()

            print(f"[after iteration {i}] Memory usage: {memory_usage():.2f} MB")

print("=== Running without tracing ===")
run_test(use_tracing=False)

print("\n=== Running with tracing ===")
run_test(use_tracing=True)
`

TomasBernotenas:
[...]

Issue #1847, opened 2023-04-04
[BUG] Hanging when browser runs out of memory


nathanielobrown:
**Context:**
- Playwright Version: 1.31.1
- Operating System: Linux+docker
- Python version: 3.10.6
- Browser: Chromium

**Code Snippet**

My code hangs when trying to save a trace to a file every so often

```python
from pathlib import Path
import tempfile
from playwright.sync_api import sync_playwright

with sync_playwright() as p:
    browser = p.chromium.launch()
    with browser.new_context() as context:
        page = browser.new_page()
        context.tracing.start(screenshots=True, snapshots=True)
        try:
            page.goto("https://some_url.com")
            # Do some stuff that fails
        finally:
            file = tempfile.NamedTemporaryFile(suffix=".zip")
            path = Path(file.name)
            file.close()
            # This hangs forever, using 100% CPU
            context.tracing.stop(path=path)
            # Would return path to some parent function...
    browser.close()
```

**Describe the bug**
It seems like something goes wrong with the trace saving and then Playwright gets caught in an infinite loop on [these two lines](https://github.com/microsoft/playwright-python/blob/main/playwright/_impl/_sync_base.py#L101-L102):
```python
        while not task.done():
            self._dispatcher_fiber.switch()
```
The main Python process is pegged at 100% CPU executing these two lines. When I use `lsof`, no process even has the zipfile I'm trying to save the trace to open. There may have been an exception that inside the `try` block that Playwright is unable to handle saving traces for. I will add more logging to figure this out and report back, but I feel like Playwright should not get hung here regardless.

![hung_trace_capture](https://user-images.githubusercontent.com/5266576/229849060-0c80b24d-f716-439b-a282-50ab8a2fb7ab.png)

mxschmitt:
I would recommend for now to upgrade to 1.32.1+ since we changed tracing a lot and how it works under the hood to process it, when getting stopped.

nathanielobrown:
I upgraded Playwright to 1.32.1 and can confirm that this issue still happens on 1.32.1.

From my logs, one of the exceptions that occurred before I tried to stop tracing is `<class 'playwright._impl._api_types.Error'>: Navigation failed because page was closed!`. I'm not sure that this is the _only_ exception that causes the hang, and I also know that this exception does not cause a hang in every case.

My understanding is this error indicates a page crash. Is it valid to try to call `tracing.stop()` after a page crash?

mxschmitt:
[...]

PR #3053, opened 2026-04-29
fix: network task handles leak

Gby56:
Hello 😄 

Given several reports of memory leaks, and my personal experience of long-running contexts getting bloated after some time, leading to OOM problems, I wanted to investigate a bit with some help from Codex (and my personal investigation)

I added a failing test case to start

Findings

  - High confidence: playwright/_impl/_network.py:896 leaks a close-watcher task on normal completion. It creates on_finished_task, waits for either response-finished or page-close, but
  never cancels the page-close task when the response wins. I reproduced this with a local micro-check: normal finish leaves one pending close-watch task attached to the page close
  future.
  - High confidence: repeated browser_type.connect() can retain closed child connections. playwright/_impl/_browser_type.py:244 appends each child Connection to parent
  _child_ws_connections at playwright/_impl/_browser_type.py:291, but nothing removes it. playwright/_impl/_connection.py:327 also does not clear _objects, so a retained closed child can
  keep its object graph alive.
  - Medium confidence: disposed channel owners do not cancel pending async event tasks. playwright/_impl/_connection.py:177 removes protocol objects from maps, but does not cancel pyee
  pending futures or clear listeners. This is risky for long-running route/request handlers, because route events are scheduled as tasks in playwright/_impl/_page.py:287 and active route
  invocations retain routes in playwright/_impl/_helper.py:416. This matches the shape of GitHub issue #1779 (https://github.com/microsoft/playwright-python/issues/1779).
  - Expected-but-important: APIResponse bodies from request.fetch() / route.fetch() stay in memory until explicit dispose or context close. The implementation only releases them via pla
  ywright/_impl/_fetch.py:516, and the official docs state the body remains in memory until dispose/context close: https://playwright.dev/python/docs/api/class-apiresponse. In a long-li
  ved context, missing await response.dispose() after route.fetch() is enough to look like a leak.
  - Lower confidence: tracing cleanup still has exception-path risks. The known tracing object leak #2977 (https://github.com/microsoft/playwright-python/issues/2977) has a regression t
  est in tests/test_reference_count_async.py:78, but playwright/_impl/_tracing.py:68 increments global tracing state before tracing_started() completes, and playwright/_impl/_tracing.py
  :86 is not exception-safe around local zip/export cleanup. That lines up with trace-stop hang reports like #1847 (https://github.com/microsoft/playwright-python/issues/1847).

### DIFF
--- a/tests/async/test_page_network_response.py
+++ b/tests/async/test_page_network_response.py
@@ -71,6 +71,34 @@ async def test_should_reject_response_finished_if_context_closes(
     error = exc_info.value
     assert "closed" in error.message
 
+async def test_response_finished_should_not_leave_page_close_waiter(
+    page: Page, server: Server
+) -> None:
+    def response_finished_waiters() -> list[asyncio.Task]:
+        return [
+            task
+            for task in asyncio.all_tasks()
+            if not task.done()
+            and getattr(task.get_coro(), "__qualname__", "")
+            == "Response.finished.<locals>.on_finished"
+        ]
+
+    baseline = set(response_finished_waiters())
+    leaked_waiters = []
+    try:
+        response = await page.goto(server.EMPTY_PAGE)
+        assert response
+        await response.finished()
+        await asyncio.sleep(0)
+
+        leaked_waiters = [
+            task for task in response_finished_waiters() if task not in baseline
+        ]
+        assert not leaked_waiters, "Response.finished() leaked page-close waiter tasks"
+    finally:
+        await page.close()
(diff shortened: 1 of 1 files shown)


Gby56:

Skn0tt:
We're not accepting failing test cases without a fix.

Gby56:
@Skn0tt could i get at least a confirmation to whether the issue seems valid on your end, before i provide a fix ? I'm willing to help but I don't know all the internals here, I have just observed the problem regularly.

PR #3055, opened 2026-04-30
fix: support async predicates in page.expect_request/expect_response

### LINKED ISSUE #1545: [Question]: how can I access async response's body in the predicate?
### Your question

I can simply do this with sync api but I have no idea how to with async api;
I've tried code like:
```
def check(response):
  body_text = response.text()
  # check if it is the response I need

async def foo():
  page = await browser.new_page()
  async with page.expect_response(url_or_predicate=check) as res:
    #other job
```
but only got error said coroutine returned by response.text() needs await;
if I change check to async function I got another error said check needs await.
what is the correct way accessing response's body with async api?

Skn0tt:
Closes #1545.

Async predicates passed to `page.expect_request` / `page.expect_response` were called but never awaited, so the returned coroutine evaluated truthy and the first matching event resolved the wait (with a `RuntimeWarning: coroutine '...' was never awaited`).

- `Waiter` now detects coroutine results from predicates and awaits them in a tracked task; exceptions from predicates reject the waiter instead of being silently swallowed by pyee.
- The async API's predicate parameter type is widened to `Callable[[T], Union[bool, Awaitable[bool]]]`. The sync API keeps `Callable[[T], bool]` — codegen was extended with a `SYNC_API` flag that strips the `Awaitable` arm from both signatures and docstrings for the sync generator.
- Two tests added in `tests/async/test_page.py` covering the success path and exception propagation.

### DIFF
--- a/playwright/_impl/_waiter.py
+++ b/playwright/_impl/_waiter.py
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 import asyncio
+import inspect
 import math
 import uuid
 from asyncio.tasks import Task
-from typing import Any, Callable, List, Tuple, Union
+from typing import Any, Callable, List, Optional, Tuple, Union
 
 from pyee import EventEmitter
 
@@ -71,9 +72,11 @@ class Waiter:
         error: Union[Error, Callable[..., Error]],
         predicate: Callable = None,
     ) -> None:
+        def on_match() -> None:
+            self._reject(error() if callable(error) else error)
+
         def listener(event_data: Any = None) -> None:
-            if not predicate or predicate(event_data):
-                self._reject(error() if callable(error) else error)
+            self._evaluate_predicate(predicate, event_data, on_match)
 
         emitter.on(event, listener)
         self._registered_listeners.append((emitter, event, listener))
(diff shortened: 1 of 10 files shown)


playwright/_impl/_waiter.py
```diff
@@ -117,12 +120,43 @@ def wait_for_event(
...
+        event_data: Any,
+        on_match: Callable[[], None],
+    ) -> None:
+        if predicate is None:
+            on_match()
+            return
+        try:
+            result = predicate(event_data)
+        except Exception as e:
+            self._reject(e)
```
dgozman, line 142: Is this line a separate bugfix? I don't see where we would reject after a sync predicate throws before this PR. Perhaps add a test for this as well?

dgozman (approved):

playwright/_impl/_waiter.py
Skn0tt: Good catch — yes, this is a coupled bugfix. Previously sync predicates that threw were silently swallowed by pyee's listener machinery, leaving the `expect_*` call to hang until timeout. Pushed a test (`test_expect_response_should_reject_when_sync_predicate_throws`) covering it.

Skn0tt: Cross-checked against JS `waiter.ts`'s `waitForEvent` ([packages/playwright-core/src/client/waiter.ts:113-124](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/waiter.ts#L113-L124)) — JS already `await`s predicates and rejects on either sync or async predicate exceptions (`try/catch` wraps `await predicate(eventArg)`). This PR brings Python to parity.

PR #3057, opened 2026-04-30
fix: propagate contextvars to event handlers

### LINKED ISSUE #1816: [Feature] Propagate contextvars to event handlers
When using the Playwright Python library, event callbacks registered with `page.on()` do not have the same contextvars as the parent code. This behavior can lead to issues when users expect shared context between the main code and event handlers.

It would be great if Playwright could propagate contextvars to event handlers by default to facilitate shared context.

## Reproduction Code Sample:

```python
import contextvars

shared_var = contextvars.ContextVar("shared_var")

def on_request(request):
    try:
        print("Shared value in callback:", shared_var.get())
    except LookupError:
        print("Failed to get shared value in callback")

def main():
    shared_var.set("some value")

    with sync_playwright() as p:
        browser = p.chromium.launch()
        page = browser.new_page()

        page.on("request", on_request)

        # Perform some actions and trigger the event
        browser.close()

if __name__ == "__main__":
    main()
```

## Context
This came up for me because I used context managers to annotate logs but my logging of network requests in Playwright were not getting annotated.

Skn0tt:
Fixes #1816

Capture the caller's contextvars context when an event handler is registered, and re-establish it when the handler runs. Without this, contextvars set in user code (e.g. request IDs in logging frameworks) were not visible inside event handlers because events are dispatched from a different greenlet (sync mode) or asyncio task (async mode) than the one that registered the handler.

### Implementation per mode

- **Async-mode coroutine handler** — spawn an inner `Task` inside the captured context so the Task adopts it (Tasks copy the active context at construction).
- **Async-mode sync handler** — run via `Context.run`.
- **Sync mode** — temporarily set the `EventGreenlet`'s `gr_context` to the captured context for the duration of the handler. `Context.run` is not used here because handlers like `route.fulfill` internally call `greenlet.switch`, and `Context.run` does not compose with greenlet switches (it would unwind on switch and corrupt context state on switch-back).

### Tests

Added regression tests for both sync and async modes that set a `ContextVar`, register a request handler, navigate, and assert the handler observed the value. The async test also asserts the value survives an `await asyncio.sleep(0)` boundary inside the handler.

### Verification

- New tests pass.
- Full sync suite: 671/671 pass.
- Full async suite: 1521/1521 pass.

### DIFF
--- a/playwright/_impl/_impl_to_api_mapping.py
+++ b/playwright/_impl/_impl_to_api_mapping.py
@@ -12,8 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
+import contextvars
 import inspect
 from typing import Any, Callable, Dict, List, Optional, Sequence, Union
+
+import greenlet
 
 from playwright._impl._errors import Error
 from playwright._impl._map import Map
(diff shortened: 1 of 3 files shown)


dgozman (commented):
Looks involved, let's wait for more upvotes!

PR #3065, opened 2026-05-04
feat(assertions): add expect.soft() for collecting multiple failures

### LINKED ISSUE #1272: Can we please get soft assertions added to the python tools

Skn0tt:
## Summary
- Adds JS-style `expect.soft(locator).to_have_text(...)` that records failures into a per-scope collector instead of raising.
- The collector scope is established by the pytest plugins (microsoft/playwright-pytest#312) via an autouse fixture that re-raises collected failures (or a `BaseExceptionGroup`) at end of test.
- Outside an active scope, `expect.soft(...).<assertion>(...)` raises `RuntimeError` so the missing plugin requirement is obvious instead of silently behaving as a hard assertion.

Fixes https://github.com/microsoft/playwright-python/issues/1272

Companion PR (pytest plugin integration): microsoft/playwright-pytest#312

### DIFF
--- a/playwright/async_api/__init__.py
+++ b/playwright/async_api/__init__.py
@@ -130,18 +130,63 @@ class Expect:
     def __call__(
         self, actual: Union[Page, Locator, APIResponse], message: Optional[str] = None
     ) -> Union[PageAssertions, LocatorAssertions, APIResponseAssertions]:
+        return self._dispatch(actual, message, is_soft=False)
+
+    @overload
+    def soft(self, actual: Page, message: Optional[str] = None) -> PageAssertions: ...
+
+    @overload
+    def soft(
+        self, actual: Locator, message: Optional[str] = None
+    ) -> LocatorAssertions: ...
+
+    @overload
+    def soft(
+        self, actual: APIResponse, message: Optional[str] = None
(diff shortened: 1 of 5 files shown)


playwright/async_api/__init__.py
```diff
@@ -129,19 +129,64 @@ def __call__(
...
+        self, actual: APIResponse, message: Optional[str] = None
+    ) -> APIResponseAssertions: ...
+
+    def soft(
+        self, actual: Union[Page, Locator, APIResponse], message: Optional[str] = None
+    ) -> Union[PageAssertions, LocatorAssertions, APIResponseAssertions]:
+        """
+        Creates a [soft assertion](https://playwright.dev/python/docs/test-assertions#soft-assertions).
+        Failing soft assertions do not abort test execution, but mark the test
+        as failed. Multiple failures from the same test are surfaced together
```
dgozman, line 156: Let's mention this requires the pytest-playwright plugin.

dgozman (approved):

PR #3084, opened 2026-05-22
Fix handler wrapping for keyword-only parameters

### LINKED ISSUE #3067: [Bug]: `add_locator_handler` with a handler that takes optional arguments with a type other than `Locator` does not cause a type error and crashes at runtime
### Version

1.59.0

### Steps to reproduce

```py

with sync_playwright() as p:
    page = p.chromium.launch(headless=False).new_page()

    input_field = page.locator("input")
    page.add_locator_handler(page.locator("div"), input_field.blur)
    page.set_content("<input></input><div>asdf</div>")
    page.click("div")
```

### Expected behavior

either:

- a type checker error
  - not sure if this is possible, would need to somehow ban certain subtypes of `Callable[[], Any]`, may be achievable with overloads
  - perhaps `add_locator_handler` can be updated to enforce that the `Locator` argument is mandatory, and users who don't need to use the argument can simply use a `_`, for example:
    ```py
    page.add_locator_handler(locator, lambda _: locator.click())
    ```
    this is a common pattern for event listeners, however it would be a breaking change
- at runtime, `SyncBase._wrap_handler` should more accurately detect whether the function takes a positional argument of type `Locator`. currently it only checks the length of `parameters`, which is inaccurate: https://github.com/microsoft/playwright-python/blob/9df9fac6137bfc4948f5c6774bdbb8669886bb4a/playwright/_impl/_impl_to_api_mapping.py#L122-L125

### Actual behavior

runtime error:
```
TypeError: Locator.blur() takes 1 positional argument but 2 were given
```

this is because `Locator.blur` takes an optional `timeout` keyword argument, which playwright incorrectly assumes is a positional argument that accepts a `Locator`.

there's no type error because a function where all its arguments are optional is a subtype of `Callable[[], Any]`

### Additional context

ruff's [`unnecessary-lambda`](https://docs.astral.sh/ruff/rules/unnecessary-lambda/) rule flags code like this:

```py
page.add_locator_handler(page.locator("div"), lambda: input_field.blur())
```

when the user follows the suggestion to inline it, they'll encounter this issue.

### Environment

```Text
- Python Version: 3.14
```

rohan-patnaik:
Fixes #3067

## Summary
This fixes handler wrapping so keyword-only parameters are not counted as positional handler arguments.

## Why
Before this change, a handler like `locator.blur` was treated as if it accepted the triggering locator because it has a keyword-only `timeout` option. Playwright then passed the locator positionally and crashed with `TypeError`.

## Test plan
- `python3 -m pytest tests/common/test_impl_to_api_mapping.py`
- `python3 -m black --check playwright/_impl/_impl_to_api_mapping.py tests/common/test_impl_to_api_mapping.py`
- `python3 -m mypy playwright/_impl/_impl_to_api_mapping.py`


rohan-patnaik:

biswajeetdev (commented):
The fix correctly identifies the problem: `inspect.signature(handler).parameters` includes `KEYWORD_ONLY` and `VAR_KEYWORD` entries in its count, so a handler like `def blur(self, *, timeout=None)` was getting counted as having 2 parameters and receiving a spurious positional arg it could not accept.

The new logic — counting only `POSITIONAL_ONLY` and `POSITIONAL_OR_KEYWORD` params, and falling back to `len(args)` when `*args` is present — is exactly right.

Two small suggestions:

**1. Add a test for the `*args` / `has_varargs` branch**

The current test covers keyword-only params, but the `has_varargs` path is not exercised:

```python
def test_wrap_handler_passes_all_args_for_varargs_handler() -> None:
    calls = []

    def handler(*args):
        calls.append(args)

    ImplToApiMapping().wrap_handler(handler)("a", "b", "c")
    assert len(calls[0]) == 3
```

**2. `POSITIONAL_ONLY` params in practice**

Python's `POSITIONAL_ONLY` kind (params before `/`) is rare in pure-Python code but common in C-extension signatures and some `inspect`-wrapped callables. The fix correctly includes them — just noting it is a good edge case to keep in mind if future test fixtures include C-extension handlers.

rohan-patnaik:
Addressed the review suggestion by adding coverage for the `*args` / `has_varargs` path in `wrap_handler`.

rohan-patnaik:
@Skn0tt This PR solves the #3067 , I addressed the review feedback above by adding coverage for the `*args` path and verified the focused tests/type checks. Would you be able to take a look when you have a chance?

Skn0tt (commented):
The change itself looks good, could you turn the test into an e2e test in `tests/async/test_page_add_locator_handler.py` and `tests/sync/test_page_add_locator_handler.py`?

rohan-patnaik:
Addressed, thanks. I moved the handler coverage from the lower-level tests/common/test_impl_to_api_mapping.py test into e2e coverage in both:

-tests/async/test_page_add_locator_handler.py
-tests/sync/test_page_add_locator_handler.py

The new tests cover keyword-only handlers and the *args path through page.add_locator_handler.
Verified locally as well.

Skn0tt (approved):

PR #3085, opened 2026-05-22
feat(api): add screencast size option

### LINKED ISSUE #3083: [Feature]: Add size option to page.screencast.start()
### Feature request

Please consider exposing the `size` option on Python's `page.screencast.start()` API, matching the Node.js Playwright API.

In Node.js, `page.screencast.start()` supports controlling the output frame/video size:

```ts
await page.screencast.start({
  onFrame: ({ data }) => sendToVisionModel(data),
  size: { width: 800, height: 600 },
});
```

This is useful for real-time frame capture, thumbnails, live previews, AI vision pipelines, and controlling video receipt resolution independently from the page viewport.

### Current Python behavior

In Python, the async API currently exposes:

```py
await page.screencast.start(
    on_frame=...,
    path="video.webm",
    quality=100,
)
```

There is no `size` parameter.

I verified this in both Playwright Python 1.59.0 and 1.60.0 by inspecting the generated async API signature. In 1.60.0, the generated signature is still:

```py
async def start(
    self,
    *,
    on_frame: Optional[Callable[[ScreencastFrame], Any]] = None,
    path: Optional[Union[pathlib.Path, str]] = None,
    quality: Optional[int] = None,
) -> AsyncContextManager:
```

### Why viewport is not enough

Changing the viewport affects page layout, but it does not give Python users a way to control screencast output resolution.

For example, with a 1920x1080 viewport, the page layout is rendered at 1920x1080, but `page.screencast.start(path="video.webm")` still appears to produce a video/frame stream scaled down to the default 800x800 bounding behavior, typically around 800x450 for a 16:9 viewport.

For many automation/agent workflows, this is the wrong tradeoff: we may want to keep the page viewport at 1920x1080 for desktop layout fidelity, while requesting lower-resolution frames for a vision model, or higher-resolution video receipts for human review.

### Expected API

Something like this would provide parity with Node.js:

```py
await page.screencast.start(
    on_frame=lambda frame: send_to_vision_model(frame["data"]),
    size={"width": 800, "height": 600},
    quality=80,
)
```

and/or:

```py
await page.screencast.start(
    path="video.webm",
    size={"width": 1920, "height": 1080},
    quality=100,
)
```

### Use case

I am using Playwright Python for browser automation with agentic video receipts and page-level evidence collection. `page.screencast` is a much better fit than context-level video recording because it is page-scoped and works well with multi-tab/shared-context workflows. However, without `size`, Python users cannot control the output resolution the way Node.js users can.

Thanks for considering this API parity improvement.

rohan-patnaik:
Fixes #3083

## Summary
This adds the missing `size` option to `page.screencast.start()` for both async and sync Python APIs.

## Why
The screencast protocol already accepts a size, but Python users could only set `quality`, `path`, and `on_frame`. That meant users could not control screencast frame/video resolution without changing the page viewport.

## Test plan
- `python3 - <<'PY'` signature import check for async and sync `Screencast.start`
- `git diff --check`

### DIFF
--- a/playwright/_impl/_screencast.py
+++ b/playwright/_impl/_screencast.py
@@ -16,7 +16,7 @@ import base64
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Union
 
-from playwright._impl._api_structures import ScreencastFrame
+from playwright._impl._api_structures import ScreencastFrame, ViewportSize
 from playwright._impl._artifact import Artifact
 from playwright._impl._connection import from_nullable_channel
 from playwright._impl._disposable import DisposableStub
@@ -70,6 +70,7 @@ class Screencast:
         onFrame: ScreencastFrameCallback = None,
         path: Union[str, Path] = None,
         quality: int = None,
+        size: ViewportSize = None,
     ) -> DisposableStub:
         if self._started:
             raise Error("Screencast is already started")
@@ -80,6 +81,7 @@ class Screencast:
             None,
             {
                 "quality": quality,
+                "size": size,
                 "sendFrames": bool(onFrame),
                 "record": bool(path),
             },
--- a/playwright/async_api/_generated.py
+++ b/playwright/async_api/_generated.py
@@ -7692,6 +7692,7 @@ class Screencast(AsyncBase):
         ] = None,
         path: typing.Optional[typing.Union[pathlib.Path, str]] = None,
         quality: typing.Optional[int] = None,
+        size: typing.Optional[ViewportSize] = None,
     ) -> "AsyncContextManager":
         """Screencast.start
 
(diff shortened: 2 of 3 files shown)


Skn0tt:
Don't change the `_generated.py` files by hand.

rohan-patnaik:
@Skn0tt thanks, you’re right. I hand-edited `_generated.py` because I saw `size` was already accepted by the driver/protocol path and assumed the Python wrapper was just missing the same option.

I should have checked the generation source first. After doing that, I found `Screencast.start.size` is present in `api.json`, but marked `langs.only: ["js"]`, so Python codegen intentionally filters its docs out.

If you’re open to reopening this PR, I can fix it properly by keeping the `_impl` change, regenerating the wrappers, and adding the expected API mismatch. If you’d prefer this to be fixed in upstream API metadata first, I can move it there instead.

rohan-patnaik:
Just found that microsoft/playwright#41183 fixed the upstream source-of-truth metadata for this by removing the JS-only marker from `Screencast.start.size`.
It makes sense to leave this PR closed then.

PR #3091, opened 2026-06-08
chore: drop Conda distribution

Skn0tt:
There's very little downloads and a healthy conda-forge distribution.

PR #3092, opened 2026-06-08
devops: Build driver from source

Skn0tt:
## Build the Playwright driver from source instead of downloading it from the CDN

`setup.py` no longer downloads the prebuilt driver bundles from
`cdn.playwright.dev`. Instead it clones `microsoft/playwright` at tag
`v{driver_version}` and builds the six platform bundles from source via the
upstream `utils/build/build-playwright-driver.sh`, then embeds them into the
platform wheels exactly as before. Same released version, same artifacts — just
self-built, so the wheels no longer depend on the CDN being reachable.

### What changed
- **`setup.py`**: `download_driver()` → `ensure_driver_bundle()`, which shells
  out to a new portable `scripts/build_driver.sh <version>` (and early-returns if
  the target bundle already exists).
- **`scripts/build_driver.sh`** (new): portable bash builder that clones the
  pinned tag and produces all six bundles on a single host. Written in bash so it
  can be shared with the other language ports.
- **`.github/workflows/ci.yml`**: a single `build-driver` job builds the bundles
  once on Linux and shares them via the `driver-bundles` artifact; the `infra`,
  `build`, and `test-stable` jobs download + verify the bundles instead of each
  rebuilding the toolchain. Cross-platform builds all happen on Linux (upstream's
  script fetches a per-target Node), so Windows/macOS only need the prebuilt zips.
- **`.github/workflows/{publish_docker,test_docker}.yml`,
  `.azure-pipelines/publish.yml`**: provision Node so the wheel build can compile
  the driver.
- **Docs** (`CONTRIBUTING.md`, `ROLLING.md`, `CLAUDE.md`,
  `.claude/skills/playwright-roll/SKILL.md`): document the build-from-source flow.

### Notes
- Rebased on top of current `main` (driver `1.60.0`); contains no conda changes
  (rebased after #3091 dropped the conda distribution).
- Opening as a **draft** to see CI.


scripts/build_driver.sh
```diff
@@ -0,0 +1,150 @@
...
+}
+
+checked_out_sha() {
+  if [[ -d "$SOURCE_DIR/.git" ]]; then
+    git -C "$SOURCE_DIR" rev-parse HEAD 2>/dev/null || true
+  fi
+}
+
+clone_source() {
+  # Reuse an existing checkout only if it already points at exactly the pinned
```
dgozman: Not really sure why wipe the repo, instead of fetch/checkout the desired commit?

scripts/build_driver.sh
```diff
@@ -0,0 +1,150 @@
...
+    echo "Checked out commit '$(checked_out_sha)' but '$EXPECTED_SHA' was requested." >&2
+    exit 1
+  fi
+}
+
+build_source() {
+  echo "Installing Playwright dependencies (npm ci)"
+  (cd "$SOURCE_DIR" && npm ci)
+  echo "Building Playwright (npm run build)"
+  (cd "$SOURCE_DIR" && npm run build)
```
dgozman, line 124: I'd recommend `npm run clean` as well for a reused checkout.

setup.py
```diff
@@ -98,24 +102,20 @@ def extractall(zip: zipfile.ZipFile, path: str) -> None:
...
-        or "-next" in driver_version
-    ):
-        url = url + "next/"
-    url = url + zip_file
-    temp_destination_path = destination_path + ".tmp"
-    print(f"Fetching {url}")
-    # Don't replace this with urllib - Python won't have certificates to do SSL on all platforms.
-    subprocess.check_call(["curl", url, "-o", temp_destination_path])
-    os.rename(temp_destination_path, destination_path)
+    # Build all platform bundles from source (microsoft/playwright @ the commit
```
dgozman: It seems like I saw this comment ~15 times in this PR 😄

setup.py
```diff
@@ -152,10 +152,13 @@ def _build_wheel(
         assert self.dist_dir
         base_wheel_location: str = glob.glob(os.path.join(self.dist_dir, "*.whl"))[0]
         without_platform = base_wheel_location[:-7]
-        download_driver(wheel_bundle["zip_name"])
-        zip_file = f"driver/playwright-{driver_version}-{wheel_bundle['zip_name']}.zip"
+        ensure_driver_bundle(wheel_bundle["zip_name"])
```
dgozman, line 153: Are we going to add all drivers for all platforms into the single wheel now, since we build all of them at once? Add a filter somewhere?

dgozman (approved):

setup.py
Skn0tt: only one is extracted, see below.

PR #3096, opened 2026-06-09
chore: generate api.json for codegen instead of reading it from the driver

Skn0tt:
Python is the only Playwright port that consumes api.json. Until now the codegen read it out of the driver bundle via the hidden `print-api-json` driver CLI. This generates it directly in build_driver.sh (`API_JSON_MODE=1 node generateApiJson.js`) and stages it next to the per-platform bundles as `driver/playwright-<sha>-api.json`, so it survives upstream dropping api.json from the driver bundle.

- build_driver.sh: generate and stage api.json; require it on the fast path.
- documentation_provider.py: read the staged api.json file directly (drops the subprocess call to `python -m playwright print-api-json`).
- update_api.sh: preflight build_driver.sh so the codegen input is staged.
- ci.yml: carry the whole driver/ folder (bundles + api.json) in the shared artifact, excluding the playwright-src checkout.
- Docs: CLAUDE.md, ROLLING.md, playwright-roll SKILL.md.

Follow-up to #3092 (build driver from source).


Skn0tt:
doesn't really give us anything. if we want to improve things, let's mirror what .NET does and parse markdown docs directly.

PR #3100, opened 2026-06-09
fix: tolerate non-utf8 request post data

### LINKED ISSUE #3098: [Bug]: Request.post_data raises UnicodeDecodeError for non-UTF8 request bodies
### Version
main branch as of 2026-06-09

### Steps to reproduce
Send a request with binary bytes that are not valid UTF-8, then read `request.post_data`:

```py
async with page.expect_request("**/upload") as request_info:
    await page.evaluate("""
        () => fetch('/upload', {
            method: 'POST',
            body: new Uint8Array([255, 254])
        })
    """)
request = await request_info.value
print(request.post_data)
```

### Expected behavior
`request.post_data` should not crash. It can return a best-effort UTF-8 string, while `request.post_data_buffer` remains the exact binary-safe API.

This would match JavaScript behavior, where `Buffer.toString("utf-8")` replaces invalid bytes instead of throwing.

### Actual behavior
Python decodes request body bytes with strict UTF-8 decoding, so `request.post_data` can raise:

```text
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff ...
```

### Potential fix direction
Decode text post data with replacement, for example `decode("utf-8", errors="replace")`, and add a regression test that verifies:

- `request.post_data` does not raise for invalid UTF-8 bytes
- `request.post_data_buffer` still returns the exact original bytes

rohan-patnaik:
## Summary
- Make `Request.post_data` tolerate invalid UTF-8 bytes by decoding with replacement.
- Keep `Request.post_data_buffer` as the exact binary-safe API.
- Add a regression test for a non-UTF8 POST body.

Closes #3098

## Tests
- `python -m py_compile playwright\_impl\_network.py tests\async\test_network.py`
- source-level repro confirming `post_data` returns replacement text and `post_data_buffer` preserves `b"\xff\xfeA"`

Full browser tests were not run locally because this checkout does not contain the generated Playwright driver payload.

### DIFF
--- a/playwright/_impl/_network.py
+++ b/playwright/_impl/_network.py
@@ -209,10 +209,12 @@ class Request(ChannelOwner):
     def post_data(self) -> Optional[str]:
         data = self._fallback_overrides.post_data_buffer
         if data:
-            return data.decode()
+            return data.decode("utf-8", errors="replace")
         base64_post_data = self._initializer.get("postData")
         if base64_post_data is not None:
-            return base64.b64decode(base64_post_data).decode()
+            return base64.b64decode(base64_post_data).decode(
+                "utf-8", errors="replace"
+            )
         return None
 
     @property
--- a/tests/async/test_network.py
+++ b/tests/async/test_network.py
@@ -485,6 +485,23 @@ async def test_should_work_with_binary_post_data(page: Page, server: Server) -> 
         assert buffer[i] == i
 
 
+async def test_should_not_throw_for_non_utf8_post_data(
+    page: Page, server: Server
+) -> None:
+    await page.goto(server.EMPTY_PAGE)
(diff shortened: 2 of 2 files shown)


Skn0tt:
This is a tradeoff between parity with Playwright Node.js, and parity with pythonicness. The Pythonic thing is to crash on non-utf8 decoding, and our docstrings don't specify the behaviour here.

I'm torn on the tradeoff, so let's err on the side of caution and keep tings as they are. We can collect more feedback on the issue and if there's more interest we can revisit this.

rohan-patnaik:
Thanks for taking a look. Was treating this mostly as Node.js parity, but I understand the concern around keeping Python’s stricter decoding behaviors.
I guess we can take a look at this later if other users run into the same issue or there’s agreement to change the behavior.

PR #3102, opened 2026-06-10
chore(roll): v1.61.0

### LINKED ISSUE #3083: [Feature]: Add size option to page.screencast.start()

Skn0tt:
This rolls the driver to `1cc5a90cfa3eaa430b1a991963100f95126caa47` (v1.61) and ports the new upstream APIs:

- `APIResponse.security_details` / `.server_addr`
- `Credentials` class (WebAuthn) + `BrowserContext.credentials`
- `WebStorage` class + `Page.local_storage` / `.session_storage`
- `Screencast.start(size=)`, `.show_actions(cursor=)`, `ScreencastFrame.timestamp`
- `BrowserType.connect_over_cdp(artifacts_dir=)`

Also picks up the upstream `Credentials.create(rp_id, ...)` shape, so Python can keep options-bag properties optional. The assertion port keeps the previous failure messages by reading `FrameExpectErrorDetails` from protocol errors.

The rolling skill is updated to reference `driver/playwright-src` instead of `~/code/playwright`.


scripts/documentation_provider.py
```diff
@@ -82,7 +82,6 @@ def _patch_case(self) -> None:
                                 option = self_or_override(option)
                                 option_name = to_snake_case(name_or_alias(option))
                                 option["name"] = option_name
-                                option["required"] = False
```
Skn0tt, line 85: in `credentials.create(rp_id)`, this was making `rp_id` optional, which is incorrect.

playwright/_impl/_frame.py
```diff
@@ -199,6 +198,8 @@ async def _expect(
             },
             title=title,
         )
+        if result is None:
+            return {}
         if result.get("received"):
```
dgozman: I don't think there is `received` in result anymore.

playwright/_impl/_frame.py
```diff
@@ -186,7 +185,7 @@ async def _expect(
         expression: str,
         options: FrameExpectOptions,
         title: str = None,
-    ) -> FrameExpectResult:
+    ) -> dict:
```
dgozman: Might as well return None, there is nothing in the dict.

scripts/documentation_provider.py
dgozman: Ouch! We never had non-optional options. Let's discuss this on the meeting.

tests/async/test_assertions.py
```diff
@@ -159,8 +159,8 @@ async def test_assertions_locator_to_contain_class(page: Page, server: Server) -
     with pytest.raises(AssertionError) as excinfo:
         await expect(locator).to_contain_class("does-not-exist", timeout=100)
 
-    assert excinfo.match("Locator expected to contain class 'does-not-exist'")
-    assert excinfo.match("Actual value: foo bar baz")
+    assert excinfo.match("Expect failed")
```
dgozman: I don't think any expectations for assertions should change.

playwright/_impl/_assertions.py
```diff
@@ -100,35 +98,16 @@ async def _expect_impl(
...
-            received = result.get("received") or {}
-            aria_snapshot = None
-            if isinstance(received, dict):
-                aria_snapshot = received.get("ariaSnapshot")
-                value = received.get("value")
-                actual = parse_value(value) if value is not None else None
-            else:
-                actual = received
+        try:
+            await self._call_expect(expression, expect_options, title)
```
dgozman: Hmm... I don't see where we are handling `ErrorDetails` received in `Frame._expect` error case and appending received/ariaSnapshot as before. This logic is still needed, but is triggered upon `ErrorDetails` instead of a `Result`.

playwright/_impl/_helper.py
```diff
@@ -358,6 +359,8 @@ def parse_error(error: ErrorPayload, log: Optional[str] = None) -> Error:
     exc = base_error_class(patch_error_message(error["message"]) + log)
     exc._name = error["name"]
     exc._stack = error["stack"]
+    exc._details = error.get("details")
```
dgozman: Where is this `"details"` set? I don't think this comes from the protocol.

playwright/_impl/_assertions.py
```diff
@@ -120,12 +120,13 @@ async def _expect_impl(
                 )
             error_message = result.get("errorMessage")
             error_message = f"\n{error_message}" if error_message else ""
+            log = result.get("log") or ""
             aria_snapshot_message = (
                 f"\nAria snapshot:\n{aria_snapshot}" if aria_snapshot else ""
             )
             _record_soft_or_raise(
                 AssertionError(
-                    f"{out_message}\nActual value: {actual}{error_message} {format_call_log(result.get('log'))}{aria_snapshot_message}"
```
dgozman, line 128: Why don't we call `format_call_log` anymore?

playwright/_impl/_connection.py
```diff
@@ -422,6 +422,7 @@ def dispatch(self, msg: ParsedMessagePayload) -> None:
                 parsed_error = parse_error(
                     error["error"], format_call_log(msg.get("log"))  # type: ignore
                 )
+                parsed_error._details = msg.get("errorDetails")
```
dgozman: I think this one should be:

```py
parsed_error._details = self._replace_guids_with_channels(msg.get("errorDetails"))
```

tests/async/test_screencast.py
```diff
@@ -35,17 +35,19 @@ def on_frame(frame: ScreencastFrame) -> None:
...
+        await page.evaluate("() => document.body.style.backgroundColor = 'red'")
+        # Force a couple of paint cycles so engines that only emit on visual change
+        # still produce a frame. Mirrors upstream `ensureSomeFrames`.
+        for _ in range(3):
+            await page.evaluate(
+                "() => new Promise(f => requestAnimationFrame(() => requestAnimationFrame(f)))"
+            )
+        await page.screenshot()
+        await asyncio.wait_for(event.wait(), timeout=10)
+    finally:
```
dgozman: Why is there a try/finally now?

tests/async/test_screencast.py
```diff
@@ -70,3 +72,44 @@ async def test_show_overlays_and_overlay_apis_should_not_throw(page: Page) -> No
...
+        await page.screencast.start(on_frame=on_frame, size=size)
+        try:
+            await page.goto(server.EMPTY_PAGE)
+            await page.evaluate("() => document.body.style.backgroundColor = 'red'")
+            for _ in range(100):
+                await page.evaluate(
+                    "() => new Promise(f => requestAnimationFrame(() => requestAnimationFrame(f)))"
+                )
+            await page.screenshot()
+        finally:
```
dgozman: Same question.

tests/async/test_tracing.py
```diff
@@ -269,10 +269,8 @@ async def test_should_work_with_playwright_context_managers(
             [
                 re.compile(r'Navigate to "/empty\.html"'),
                 re.compile(r"Set content"),
-                re.compile(r'Wait for event "page\.expect_event\(console\)"'),
```
dgozman, line 272: Does this match upstream changes?

dgozman (approved):

PR #3111, opened 2026-06-14
fix(network): handle non-UTF-8 bytes in post_data without raising

### LINKED ISSUE #3098: [Bug]: Request.post_data raises UnicodeDecodeError for non-UTF8 request bodies

wali-reheman:
post_data.decode() used strict UTF-8 decoding which raises UnicodeDecodeError when the POST body contains invalid UTF-8 bytes. Use errors='replace' to return a best-effort string, matching the JavaScript Playwright API. post_data_buffer remains the binary-safe API.

Fixes https://github.com/microsoft/playwright-python/issues/3098

### DIFF
--- a/playwright/_impl/_network.py
+++ b/playwright/_impl/_network.py
@@ -209,10 +209,10 @@ class Request(ChannelOwner):
     def post_data(self) -> Optional[str]:
         data = self._fallback_overrides.post_data_buffer
         if data:
-            return data.decode()
+            return data.decode("utf-8", errors="replace")
         base64_post_data = self._initializer.get("postData")
         if base64_post_data is not None:
-            return base64.b64decode(base64_post_data).decode()
+            return base64.b64decode(base64_post_data).decode("utf-8", errors="replace")
         return None
 
     @property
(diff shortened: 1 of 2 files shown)


Skn0tt:
Closing per https://github.com/microsoft/playwright-python/pull/3100#issuecomment-4660036966.

PR #3114, opened 2026-06-18
fix: avoid retaining a completed task per protocol message on Python 3.14 (#41357)

### LINKED ISSUE #3121: [Bug]: context.route retains completed Tasks until Connection close (memory leak)
## Environment

- Playwright (Python): **1.59.0 and 1.60.0** (both affected)
- Python: **3.14.4**
- OS: Linux (Ubuntu 26.04 x64)

## Summary

Registering a `context.route(...)` / `page.route(...)` handler causes the Playwright Python client
to retain references to the **completed** asyncio Tasks it spawns per intercepted request — both the
route-handler task and its underlying `Connection.send` task. These done tasks are **not** released
when the page, context, or browser is closed — only when the Playwright **Connection** closes (end of
`async_playwright()`). On a long-lived Connection scraping many pages, this grows memory roughly
linearly with the number of intercepted requests.

It is the **interception**, not the block/allow decision: a handler that `continue_()`s every request
leaks identically to one that `abort()`s every request.

## Reproduction (plain Playwright, no extra deps)

```python
import asyncio, gc

N = 24
PAGE = "data:text/html,<html><body>" + "".join(
    f'<img src="http://x-{i}.example/i.png">' for i in range(80)
) + "</body></html>"  # 80 sub-resource requests per page, no network needed

def done_tasks():
    gc.collect()
    return sum(1 for o in gc.get_objects() if isinstance(o, asyncio.Task) and o.done())

async def loop(browser, with_route):
    for _ in range(N):
        ctx = await browser.new_context()
        if with_route:
            async def h(route):
                try: await route.abort()
                except Exception: pass
            await ctx.route("**/*", h)
        page = await ctx.new_page()
        try: await page.goto(PAGE, wait_until="load", timeout=8000)
        except Exception: pass
        await page.close()
        await ctx.close()

async def main():
    async with async_playwright() as pw:
        b = await pw.chromium.launch(headless=True, args=["--no-sandbox"])
        await loop(b, with_route=True)
        await b.close()                      # browser.close() does NOT free them
        print("WITH route, after browser.close():", done_tasks())
        b2 = await pw.chromium.launch(headless=True, args=["--no-sandbox"])
        await loop(b2, with_route=False)     # control
        await b2.close()
        print("NO route (control):", done_tasks())

asyncio.run(main())
```

## Observed

```
WITH context.route("**/*"): done-task count climbs ~160 per context
    (8 ctx -> 1281, 16 -> 2561, 24 -> 3841); NOT freed by browser.close()
NO route (control):          0 accumulation
```

A referrer trace shows the retained `Channel.send` / route-handler tasks held in a `set` reachable
from `playwright._impl._transport.PipeTransport` / `_connection.Connection`; they free only when the
Connection closes. asyncio's own registry is not the holder (`asyncio.all_tasks()` stays small).

## Impact

A long scraping run on a single shared Connection (the normal pattern) grows Python RSS ~linearly
with intercepted requests and eventually OOMs. Observed in a real pipeline: ~3.5 MB/URL, 721 MB at
200 image-heavy pages.

## Workaround (CDP-based blocking)

Block via CDP instead of `context.route` — no per-request Python handler, no task retention:

```python
cdp = await page.context.new_cdp_session(page)
await cdp.send("Network.enable")
await cdp.send("Network.setBlockedURLs", {"urls": ["*.png*", "*.woff*", ...]})
```

Same reproduction with CDP blocking instead of `context.route`: **1** done task after 20 pages
(vs ~1600 with `context.route`).

yen0304:
Fixes microsoft/playwright-python#3121.

## Summary

Registering `context.route(...)` / `page.route(...)` caused the async client to retain one **completed** `asyncio.Task` per intercepted request until the **Connection** closed (end of `async_playwright()`). Closing the page, context, or browser did not release them, so memory grew roughly linearly with the number of intercepted requests on a long-lived connection.

## Root cause

Every protocol send in `Channel._inner_send` awaited the long-lived transport `on_error_future` alongside its per-message `callback.future`:

```python
done, _ = await asyncio.wait(
    {self._connection._transport.on_error_future, callback.future},
)
```

`on_error_future` is created once per connection (`Transport.__init__`) and only resolves when the transport errors/closes. On **Python 3.14**, asyncio added an await-graph (`Future._asyncio_awaited_by`) that keeps **strong references** between a task and the futures it awaits. Because each send runs in its own task (route handlers are dispatched as tasks) and awaits the never-resolving `on_error_future`, every completed send task stayed registered in `on_error_future._asyncio_awaited_by` and was only released when that future — and the whole `Connection` — went away.

This is interception-agnostic (a handler that `continue_()`s leaks the same as one that `abort()`s) because the leak is in the generic send path, which route handling exercises once per request.

## Reproduction

Pinned down with `gc.get_referrers`; the retention chain is:

```
completed Channel.send Task  ->  set (_asyncio_awaited_by)  ->  on_error_future (Future)  ->  PipeTransport
```

Minimal, browserless confirmation (distinct task per wait against a long-lived future):

| | retained done tasks after 300 sends |
|---|---|
| Python 3.12 | 1 |
| Python 3.14 | 300 |

Full repro from the issue (80 sub-resources/page, intercept-and-abort):

| | done tasks after `browser.close()` |
|---|---|
| 3.14 before fix | 961 |
| 3.14 after fix | 1 |
| 3.12 (before/after) | 1 |

## The fix

Forward transport errors to the pending `callback.future` through a done callback instead of awaiting `on_error_future` directly, so each send's await graph stays local to its own short-lived future and is released as soon as the send completes. Behaviour is preserved:

- a send still raises the transport error if the transport dies mid-flight (verified by force-killing the driver process mid-send: the call raises rather than hanging, on both 3.12 and 3.14);
- the done callback is always removed in a `finally` block;
- no double-resolution of `callback.future` (avoids the `InvalidStateError` that a naive `set_exception` forward would race with `Connection.cleanup`).

## Test plan

- [x] New regression test `test_route_should_not_leak_done_tasks` — intercepts 100 requests across 5 contexts and asserts retained completed tasks stay bounded. Verified it **fails on unpatched 3.14** (leaked 200) and **passes after the fix** (leaked 0); passes on 3.12 too.
- [x] Existing transport-error propagation still raises (no hang) on 3.12 and 3.14.
- [x] `black` / `isort` / `flake8` clean on the changed files.

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -120,16 +120,31 @@ class Channel(AsyncIOEventEmitter):
         callback = self._connection._send_message_to_server(
             self._object, method, _augment_params(params, timeout_calculator)
         )
-        done, _ = await asyncio.wait(
-            {
-                self._connection._transport.on_error_future,
-                callback.future,
-            },
-            return_when=asyncio.FIRST_COMPLETED,
-        )
-        if not callback.future.done():
-            callback.future.cancel()
-        result = next(iter(done)).result()
+        # Wake up the pending send if the transport errors, but do not `await` the long-lived
+        # `on_error_future` directly. On Python 3.14 the asyncio await-graph keeps every waiting
+        # task in `on_error_future._asyncio_awaited_by` until that future resolves (i.e. the whole
+        # connection closes), leaking one task per protocol message. Forwarding the error via a
+        # done callback instead keeps each send's await graph local to its own short-lived future.
+        on_error_future = self._connection._transport.on_error_future
+
+        def _interrupt_on_transport_error(_: "asyncio.Future") -> None:
+            if not callback.future.done():
+                callback.future.cancel()
+
+        if on_error_future.done():
+            on_error_future.result()  # raises the transport error, if any
+        on_error_future.add_done_callback(_interrupt_on_transport_error)
+        try:
+            result = await callback.future
+        except asyncio.CancelledError:
+            # The send was interrupted. If the transport errored, surface that error to match the
+            # previous `asyncio.wait({on_error_future, callback.future})` behaviour; otherwise let
+            # the caller's cancellation propagate.
+            if on_error_future.done() and not on_error_future.cancelled():
+                raise cast(BaseException, on_error_future.exception())
+            raise
+        finally:
+            on_error_future.remove_done_callback(_interrupt_on_transport_error)
         # Protocol now has named return values, assume result is one level deeper unless
         # there is explicit ambiguity.
         if not result:
--- a/playwright/_impl/_network.py
+++ b/playwright/_impl/_network.py
@@ -555,10 +555,24 @@ class Route(ChannelOwner):
             getattr(asyncio.current_task(self._loop), "__pw_stack__", inspect.stack(0)),
         )
         target_closed_future = self.request._target_closed_future()
-        await asyncio.wait(
-            [fut, target_closed_future],
-            return_when=asyncio.FIRST_COMPLETED,
-        )
+        # Wait until either the action finishes or the page closes, but do not `await` the
+        # long-lived `target_closed_future` directly: on Python 3.14 its await graph would keep
+        # this route-handler task referenced until the page is collected. Bridge both into a
+        # short-lived local future via done callbacks instead.
+        if not (fut.done() or target_closed_future.done()):
+            waiter = self._loop.create_future()
+
+            def _wake(_: "asyncio.Future") -> None:
+                if not waiter.done():
+                    waiter.set_result(None)
+
+            fut.add_done_callback(_wake)
+            target_closed_future.add_done_callback(_wake)
+            try:
+                await waiter
+            finally:
+                fut.remove_done_callback(_wake)
+                target_closed_future.remove_done_callback(_wake)
         if fut.done() and fut.exception():
             raise cast(BaseException, fut.exception())
         if target_closed_future.done():
--- a/tests/async/test_browsercontext_route.py
+++ b/tests/async/test_browsercontext_route.py
@@ -317,8 +317,7 @@ async def test_should_override_post_body_with_empty_string(
 
     req = await asyncio.gather(
         server.wait_for_request("/empty.html"),
-        page.set_content(
-            """
+        page.set_content("""
             <script>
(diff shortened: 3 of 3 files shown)


mturac:
Nice fix. I tried this branch locally against the microsoft/playwright-python#3121 shape on Python 3.14.4 / Chromium. It removes the accumulating `Channel.send` tasks, but I still see the last route-handler batch retained after `context.close()` and `browser.close()`:

```
after ctx 1: 20 main.<locals>.h + init
after ctx 2: 20 main.<locals>.h + init
after ctx 3: 20 main.<locals>.h + init
after browser close: 20 main.<locals>.h + init
```

I think there is one more `asyncio.wait()` instance on the route path: `Route._race_with_page_close()` waits on `[fut, target_closed_future]`, so on Python 3.14 the route-handler task can still be recorded in the page close future's await graph. Avoiding `asyncio.wait()` there as well dropped my local repro to just `Connection.run.<locals>.init`.

It may also be worth tightening the regression to count `Channel.send` and route handler tasks separately; the current `< 50` threshold can still pass with one retained handler batch.

yen0304:
Great catch, thanks @mturac — you're exactly right. `Route._race_with_page_close()` is a second `asyncio.wait()` against a long-lived future (`page._closed_or_crashed_future`), so the route-handler task (`BrowserContext._on_route`) stayed in that future's await-graph on 3.14.

I pushed a follow-up that bridges both awaitables through a short-lived local future there too, instead of awaiting the page close future directly. With both spots fixed my breakdown drops to just `Connection.run.<locals>.init`:

```
done task breakdown after browser.close():
      1  Connection.run.<locals>.init
TOTAL: 1
```

I also tightened the regression test per your suggestion: it now counts the per-request internal tasks (`Channel.send` and `_on_route`) separately and asserts `<= 2`, rather than the loose total threshold that a single retained handler batch could slip past. Verified it still fails on an unpatched 3.14 and passes after the fix; functional checks (fulfill / continue / abort, plus a page-close-during-route race and a mid-send transport kill) pass on both 3.12 and 3.14.

Skn0tt:
Hi, I think this should be fixed in Python stdlib. I filed upstream, if we cannot get it fixed there we can reconsider.

PR #3134, opened 2026-07-15
chore: roll to 1.62.0-alpha-1784109367000

Skn0tt:
## Summary

- Roll the driver to `1.62.0-alpha-2026-07-16` (Node `24.18.0`).
- Adapt channel sends to the new protocol-level timeout metadata field.
- Port the public API changes: `routeFromHar(interceptAPIRequests=)`, `storageState(credentials=)`, `Locator.wait_for_function()`, the `scroll` action option, screenshot `webp` output, and `APIResponse.timing()`.
- Mirror upstream's non-BiDi Firefox skip for the client-redirect navigation race.

AbortSignal is intentionally not ported to Python, I might connect `__abort__` to task cancellation. I'll do it in a followup because it's possibly breaking.


playwright/_impl/_browser_context.py
```diff
@@ -260,10 +261,11 @@ async def _on_route(self, route: Route) -> None:
                 handled = await route_handler.handle(route)
             finally:
                 if len(self._routes) == 0:
-                    asyncio.create_task(
+                    create_task_and_ignore_exception(
```
dgozman, line 264: Why this change? Can you link to the upstream counterpart?

playwright/_impl/_network.py
```diff
@@ -714,24 +703,25 @@ def connect_to_server(self) -> "WebSocketRoute":
         if self._connected:
             raise Error("Already connected to the server")
         self._connected = True
-        asyncio.create_task(
+        create_task_and_ignore_exception(
```
dgozman: Same question here - why is this suddenly ignoring exception?

playwright/_impl/_page.py
```diff
@@ -1603,15 +1614,19 @@ async def call(self, func: Callable) -> None:
                 result = func(source, *func_args)
             if inspect.iscoroutine(result):
                 result = await result
-            await self._channel.send(
-                "resolve", None, dict(result=serialize_argument(result))
+            create_task_and_ignore_exception(
```
dgozman: It looks like most of these `create_task_and_ignore_exception` calls are wrapping `channel.send`. Perhaps create a `channel.send_may_fail` helper?

dgozman (approved):

playwright/_impl/_browser_context.py
Skn0tt: it's the `.catch(() => {})` here: https://github.com/microsoft/playwright/blob/48a2efe68af84c16fd270aefaffd81ad197a063b/packages/playwright-core/src/client/browserContext.ts#L248-L249

playwright/_impl/_network.py
Skn0tt: .catch, see https://github.com/microsoft/playwright/blob/48a2efe68af84c16fd270aefaffd81ad197a063b/packages/playwright-core/src/client/network.ts#L553-L565

PR #3158, opened 2026-07-28
fix(sync): clear __pw_stack__ on task completion to prevent memory le…

### LINKED ISSUE #3157: [Bug]: sync API attaches live stack frames to every API-call task (__pw_stack__), turning any task retention into a leak of the caller's stack and locals
### Version
playwright 1.58.0, Python 3.14.6 (also observed structurally on 3.12), macOS, arm64

### Description

`_sync_base.py::SyncBase._sync` attaches the result of `inspect.stack(0)` to every asyncio task it creates:

```python
task: asyncio.tasks.Task[Any] = self._loop.create_task(coro)
setattr(task, "__pw_stack__", inspect.stack(0))
setattr(task, "__pw_stack_trace__", traceback.extract_stack(limit=10))
```

`inspect.stack()` returns `FrameInfo` objects holding **live frame objects**, not strings. Any mechanism that retains the task beyond the call therefore pins the user's entire call stack, and every local variable in every frame of it, for as long as the task lives. `__pw_stack_trace__` (a `StackSummary` of strings) does not have this problem; `__pw_stack__` does.

We hit this in production through CPython bug python/cpython#152569 (fixed upstream 2026-07-02, backports pending in patch releases): `asyncio.wait(..., FIRST_COMPLETED)` on 3.14.0-3.14.6 permanently retains the calling task in the awaited-by set of any still-pending future. Playwright triggers that on **every protocol call**, because `Channel._inner_send` races `callback.future` against the long-lived `transport.on_error_future`:

```python
done, _ = await asyncio.wait(
    {self._connection._transport.on_error_future, callback.future},
)
```

The combination: every completed sync API call leaked its task, its `__pw_stack__`, all the caller's frames, and all their locals.

Measured impact in our workload (a screenshot-per-frame render loop that decodes each 3840x2160 PNG with PIL in the frame that calls `page.screenshot()`): ~40 MB retained per iteration (the ~33 MB decoded source image plus an ~8 MB resized copy, pinned as locals of the retained caller frame), reaching ~57 GB of compressed anonymous memory after ~1,900 iterations, at which point the kernel SIGKILLed the process. `gc.collect()` frees nothing because the awaited-by entry is a strong root. The `gc.get_referrers` chain from a leaked image object reads:

```
PIL image (33 MB)
  <- caller's frame object (locals pinned)
  <- inspect.FrameInfo
  <- finished asyncio Task for Page.screenshot (task.__pw_stack__)
  <- _asyncio_awaited_by set of a pending Future
  <- transport.on_error_future (lives for the whole browser session)
```

Restructuring our caller so no large locals live in any frame that is on-stack during a playwright call reduced the leak from ~40 MB/iteration to ~0.94 MB/iteration on 3.14.6; the residue is the retained task itself plus its result buffer (the screenshot bytes), which user code cannot release.

The CPython side is fixed, but `__pw_stack__` remains a hazard: any future retention path (another stdlib regression, a user-held task reference, an exception path that keeps the task alive) silently escalates from "a task leaked" to "the entire caller stack and its locals leaked". Even in the non-buggy case the live frames survive until cycle collection, keeping large caller locals alive longer than users expect.

### Suggested fix

Either of:
1. Capture file/line/function data (as `__pw_stack_trace__` already does) instead of live `FrameInfo` objects, converting at capture time in `_sync`, `_connection.wrap_api_call*`, and `_network.py` where `__pw_stack__` is read; or
2. Clear `task.__pw_stack__` when the call completes (e.g. in the `add_done_callback` alongside the greenlet switch), so the frames can never outlive the call.

### Steps to reproduce the retention shape

On Python 3.14.0-3.14.6 (before the gh-152569 backport): run any sync-API loop, e.g. `page.screenshot()` per iteration, and observe finished `Task` objects with `__pw_stack__` accumulating in `transport.on_error_future._asyncio_awaited_by` (surviving `gc.collect()`), each holding live frames whose `f_locals` are reachable. On fixed interpreters the same objects exist transiently until cycle collection.

atiqur-rahman-pro:
### Summary of Changes
In `playwright/_impl/_sync_base.py`, `inspect.stack(0)` is attached to `task.__pw_stack__`. Because live `FrameInfo` objects pin all caller local variables in memory, retaining the task causes a memory leak.

This PR updates `task.add_done_callback` to delete `task.__pw_stack__` upon task completion, ensuring frame objects and local variables are garbage collected immediately.

Fixes #3157

### DIFF
--- a/playwright/_impl/_sync_base.py
+++ b/playwright/_impl/_sync_base.py
@@ -108,7 +108,12 @@ class SyncBase(ImplWrapper):
         setattr(task, "__pw_stack__", _capture_stack_trace())
         setattr(task, "__pw_stack_trace__", traceback.extract_stack(limit=10))
 
-        task.add_done_callback(lambda _: g_self.switch())
+        def _on_done(_: Any) -> None:
+            if hasattr(task, "__pw_stack__"):
+                delattr(task, "__pw_stack__")
+            g_self.switch()
+
+        task.add_done_callback(_on_done)
         while not task.done():
             self._dispatcher_fiber.switch()
         asyncio._set_running_loop(self._loop)


Skn0tt:
Closing per https://github.com/microsoft/playwright-python/issues/3157#issuecomment-5128649259

PR #3160, opened 2026-07-29
chore(driver): roll to 1.62.0

### LINKED ISSUE #38919: [Bug]: locale is not propagated to workers in firefox
### Version

95a728a037070f31afb7decebbc4d9e2284e26ba

### Steps to reproduce

https://github.com/microsoft/playwright/blob/95a728a037070f31afb7decebbc4d9e2284e26ba/tests/library/browsercontext-locale.spec.ts#L166

This test passes in CI, but not locally. I assume it's because CI locale happens to give the same output as es-MX. If you change the locale to ru-RU, it should also fail in CI.

It seems that workers are not respecting the context locale.

### Expected behavior

Workers should be respecting context locale.

### Actual behavior

Workers are using system locale.

### Additional context

found during https://github.com/microsoft/playwright-python/pull/3026

### Environment

```shell
System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 133.67 MB / 18.00 GB
  Binaries:
    Node: 20.19.0 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/node
    Yarn: 1.22.22 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/yarn
    npm: 9.9.4 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/npm
    pnpm: 10.28.1 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/pnpm
    Deno: 2.5.2 - /opt/homebrew/bin/deno
  IDEs:
    VSCode: 1.108.1 - /opt/homebrew/bin/code
    Claude Code: 2.1.15 - /opt/homebrew/bin/claude
  Languages:
    Bash: 3.2.57 - /bin/bash
```

dcrousso:
Rolls the driver to `1.62.0` and regenerates the sync and async APIs from the `v1.62.0` documentation.

Upstream removed the `routeFromHar` `interceptAPIRequests` option before `1.62.0` in <https://github.com/microsoft/playwright/pull/41964> and <https://github.com/microsoft/playwright/pull/41965>.  This drops `intercept_api_requests`, the `HarRouter.add_api_request_route` helper, the `routeAPIRequestsFromHar` and `unrouteAPIRequestsFromHar` calls, and their tests.  The tests asserting API requests are not served from HAR by default remain.

Firefox `153` fixes <https://github.com/microsoft/playwright/issues/38919>, so the worker locale test now uses the context locale expectation for every browser.

The `1.62.0` driver requires each `screencastFrame` event to be acknowledged after its callback finishes.  The Python client now sends `screencastFrameAck` after sync callbacks and awaited async callbacks, preserving backpressure.

Validation:

- `pre-commit` passes for the modified screencast files
- Chromium screencast coverage passes with `12` tests
- The async backpressure test passes on Chromium, Firefox, and WebKit
- `tests/async` on Chromium passes with `1581` tests and `7` skipped tests
- HAR coverage passes with `68` tests
- API generation used `PW_SRC_DIR` pinned to `v1.62.0` at `e3950d9c140d007bd52853b45813c6274b24e36f`


Skn0tt (commented):
I believe this is missing the screencast ack, see https://github.com/microsoft/playwright/commit/bc97a1306ceb2c4517b1e6de4a00efb64a94d99c.

playwright/_impl/_screencast.py
```diff
@@ -54,21 +54,39 @@ def __init__(self, page: "Page") -> None:
...
+                    {
+                        "data": data,
+                        "timestamp": params.get("timestamp", 0),
+                        "viewportWidth": params["viewportWidth"],
+                        "viewportHeight": params["viewportHeight"],
+                    }
+                )
+        finally:
+            if result is not None and hasattr(result, "__await__"):
+                self._page._loop.create_task(
```
Skn0tt, line 74: this task should be somehow returned into the `page._channel.on()` caller in L54,  otherwise exceptions in `onFrame` customer code are silently swallowed. This isn't a problem in Node.js land because there's the `unhandled rejection` concept, this doesn't exist in Python though.

Otherwise this is the same as I came up with in https://github.com/microsoft/playwright-python/pull/3161, which is a good sign. Let's figure out why these Chromium tests are hanging.

Skn0tt:
The failing CI seems to be related to https://github.com/microsoft/playwright/pull/41303 and `page.close()` hanging, i'm looking into it.

Edit: See https://github.com/microsoft/playwright-python/pull/3161#issuecomment-5142354567.

dcrousso:
closing in favor of <https://github.com/microsoft/playwright-python/pull/3161>

PR #3161, opened 2026-07-30
chore: roll Playwright to 1.62.0

### LINKED ISSUE #3145: [Feature]: Allow controlling JPEG frame rate in page.screencast.start()to reduce bandwidth and CPU usage
### 🚀 Feature Request

I'm running multiple Playwright browser contexts concurrently, using `page.screencast.start()` to capture real-time frames and stream them as a "live view" (think: headless browser farm + dashboard preview).

In this use case:
- 1–5 fps is **more than enough** for human viewing
- But `page.screencast` currently emits frames at a fixed (high) rate internally
- With N browsers × default FPS, CPU usage becomes the bottleneck long before network/bandwidth

There's no way to tell the browser "don't capture 25fps, I only need 5".

### Example

_No response_

### Motivation

We run multiple Playwright browser contexts concurrently to stream real-time browser views (browser farm / dashboard preview).

page.screencast.start()currently captures frames at a fixed high frame rate. In our live-view scenario, 1–5 fps is sufficient, but we cannot configure this limit. As a result, CPU usage scales linearly with the number of tabs—even though most frames are never visually meaningful to the viewer.

This makes screencast impractical for multi-browser deployments due to unnecessary CPU overhead.

Skn0tt:
Closes https://github.com/microsoft/playwright-python/issues/3145


Skn0tt:
The failing CI seems to be related to https://github.com/microsoft/playwright/pull/41303 and `page.close()` hanging, i'm looking into it.

Skn0tt:
https://github.com/microsoft/playwright-python/pull/3161/commits/5aee3618f6a31f8286004eb4b4c61f2bb4369dbc should fix it by moving our fixtures here closer to what upstream does, and relying on `context.close()` for cleanup solely instead of also having `page.close()` on page fixture teardown. This only affects this repo and not our customers, the `playwright-pytest` integration already has this behaviour.

dcrousso (approved):

PR #836, opened 2021-08-05
chore(protocol): implement client side hello

mxschmitt:

PR #3168, opened 2026-08-04
fix(sync): wait for initialize before leaving __enter__

Skn0tt:
Fixes half of https://github.com/microsoft/playwright-python/issues/3165

`sync_api/_context_manager.py` blocked until the Playwright `__create__`, but
since #836 we really have to block until `initialize` finishes. Empty
`with sync_playwright()` was racing teardown against that still-in-flight RPC
and printing `Future exception was never retrieved` / `Task was destroyed but
it is pending` on exit.

Migrating to wait on `playwright_future` (like async) also lets us drop the
now-unused `call_on_object_with_known_name` path.

Does not touch the separate JSON-serialization / register-after-send half.

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -305,7 +305,6 @@ class Connection(EventEmitter):
         self._dispatcher_fiber = dispatcher_fiber
         self._transport = transport
         self._transport.on_message = lambda msg: self.dispatch(msg)
-        self._waiting_for_object: Dict[str, Callable[[ChannelOwner], None]] = {}
         self._last_id = 0
         self._objects: Dict[str, ChannelOwner] = {}
         self._callbacks: Dict[int, ProtocolCallback] = {}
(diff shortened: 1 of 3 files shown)


playwright/_impl/_connection.py
```diff
@@ -341,7 +340,11 @@ async def run(self) -> None:
         self._root_object = RootChannelOwner(self)
 
         async def init() -> None:
-            self.playwright_future.set_result(await self._root_object.initialize())
+            try:
+                self.playwright_future.set_result(await self._root_object.initialize())
+            except BaseException as exc:
+                self.playwright_future.set_exception(exc)
```
dcrousso: i think the async context manager cancels `playwright_future` when the transport fails first so this may raise `InvalidStateError` and replace the original failure in `_init_task`

playwright/_impl/_connection.py
```diff
@@ -341,7 +340,11 @@ async def run(self) -> None:
...
+                raise
```
dcrousso: `_init_task` does not appear to be awaited or inspected so i think rethrowing here will emit `Task exception was never retrieved` after the caller has already received the startup error

dcrousso (changes requested):

playwright/_impl/_connection.py
Skn0tt: Good catch. `set_result` / `set_exception` are now both gated on `not playwright_future.done()`, so if async `__aenter__` already cancelled after a transport error we just back off.

playwright/_impl/_connection.py
Skn0tt: Yep - callers only wait on the future, nobody awaits `_init_task`. Dropped the re-raise so the task does not log "exception was never retrieved" after enter already got the error via the future.

dcrousso (approved):

PR #3135, opened 2026-07-16
fix(remote): reject waiters before disconnecting

Skn0tt:
This PR aligns remote browser shutdown ordering with Playwright Node.js. Pending page and context waiters now reject before `browser.close()` resolves or the `disconnected` event fires.

Waiter telemetry is best-effort during shutdown, so it can't mask the original rejection.

Closes #3097

PR #3170, opened 2026-08-06
fix(connection): reject pending callbacks on transport death

Skn0tt:
Every send used to race the protocol callback against `transport.on_error_future`, because a dead pipe only completed that future and never tore down in-flight calls.

This matches the Node client more closely: `Connection.run()` calls `cleanup()` when the transport ends, which rejects every pending callback. Sends just await their own future.

For remote connect, `browser._on_close` is still `call_soon`'d before `cleanup()`, so waiters still see reject-before-disconnect ordering.

### DIFF
--- a/playwright/_impl/_browser_type.py
+++ b/playwright/_impl/_browser_type.py
@@ -257,15 +257,8 @@ class BrowserType(ChannelOwner):
                     for page in context.pages:
                         page._on_close()
                     context._on_close()
-            connection.cleanup(reason)
-            # Give a chance to any API call promises to reject upon page/context closure.
-            # This happens naturally when we receive page.onClose and browser.onClose from the server
-            # in separate tasks. However, upon pipe closure we used to dispatch them all synchronously
-            # here and promises did not have a chance to reject.
-            # The order of rejects vs closure is a part of the API contract and our test runner
-            # relies on it to attribute rejections to the right test.
-            if browser:
                 connection._loop.call_soon(browser._on_close)
+            connection.cleanup(reason)
 
         transport.once("close", handle_transport_close)
 
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -144,13 +144,7 @@ class Channel(AsyncIOEventEmitter):
             self._object, method, augmented_params, timeout
         )
         try:
-            done, _ = await asyncio.wait(
-                {
-                    self._connection._transport.on_error_future,
-                    callback.future,
-                },
-                return_when=asyncio.FIRST_COMPLETED,
-            )
+            result = await callback.future
         except asyncio.CancelledError as exc:
             await self._connection._abort(
                 self._object,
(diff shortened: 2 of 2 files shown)


playwright/_impl/_browser_type.py
```diff
@@ -257,15 +257,8 @@ def handle_transport_close(reason: Optional[str]) -> None:
...
-            connection.cleanup(reason)
-            # Give a chance to any API call promises to reject upon page/context closure.
-            # This happens naturally when we receive page.onClose and browser.onClose from the server
-            # in separate tasks. However, upon pipe closure we used to dispatch them all synchronously
-            # here and promises did not have a chance to reject.
-            # The order of rejects vs closure is a part of the API contract and our test runner
-            # relies on it to attribute rejections to the right test.
-            if browser:
                 connection._loop.call_soon(browser._on_close)
+            connection.cleanup(reason)
```
dcrousso, line 261: this seems like it'd reintroduce the problem solved by <https://github.com/microsoft/playwright-python/pull/3135/>

playwright/_impl/_connection.py
```diff
@@ -144,23 +144,14 @@ async def _inner_send(
...
         )
         try:
-            done, _ = await asyncio.wait(
-                {
-                    self._connection._transport.on_error_future,
-                    callback.future,
-                },
-                return_when=asyncio.FIRST_COMPLETED,
-            )
+            result = await callback.future
```
dcrousso, line 147: this seems like it'd reintroduce the problem solved by <https://github.com/microsoft/playwright-python/pull/3144/>

Skn0tt: Can you elaborate on how? Task cancellation still triggers `__abort__`, which ends the request on the server side and eventually makes this callback reject.

playwright/_impl/_browser_type.py
Skn0tt: Hmm, the test added is still green though. Can you explain how?

dcrousso: that test stays green because `page._on_close()` synchronously rejects its local `Waiter` before `browser._on_close` is queued

this change queues `browser._on_close` before `connection.cleanup()` rejects its callback, so `disconnected` can run first

a pending `page.evaluate("new Promise(() => {})")` would expose the ordering regression

playwright/_impl/_connection.py
dcrousso: the distinction is that `__abort__` is still sent, but the caller no longer waits for it to complete

since `callback.future` is directly awaited, canceling the task also cancels that future

`_abort()` then sends `__abort__`, awaits the already canceled future, catches `CancelledError` immediately, and returns

the later server response is discarded by the `callback.future.cancelled()` branch in `dispatch()`, rather than making the callback reject

the existing test stays green because it checks the eventual effect of sending `__abort__`, not whether cancellation waits for the protocol response

Skn0tt:
Gave it a shot, but it looks like this change isn't worth the hassle just for being more in-line with JS! Let's stick with what we have.

PR #3175, opened 2026-08-17
fix(evaluate): 0.0 is serialized as -0

ARDA7787:


Skn0tt:
As per guidelines, please start with filing an issue.

PR #3177, opened 2026-08-19
fix(asyncio): do not hang when async_playwright() is cancelled while connecting

### LINKED ISSUE #42296: [Bug]: async_playwright() leaves blocking background task if cancelled quickly
### Version

1.62.0

### Steps to reproduce

With Pythin 3.14.7 and Playwright 1.62.0:

```python
import asyncio

async def run_browser():
    async with async_playwright():
        pass

async def run():
    task = asyncio.create_task(run_browser())
    await asyncio.sleep(0.01)  # for me: some value between 0.001 and 0.3
    task.cancel()
    try:
        await task
    except asyncio.CancelledError:
        pass

asyncio.run(run())
```

### Expected behavior

The program exits.

### Actual behavior

The program hangs.

### Additional context

_No response_

### Environment

```shell
Linux 7.1.3
Python 3.14.7
playwright 1.62.0 (latest version on PyPI)
```

yury-s:
## Summary
- stop the connection (driver process and background tasks) when `__aenter__` fails or is cancelled, reusing `__aexit__` like the sync context manager does
- allow `PipeTransport.request_stop()` before the driver has spawned
- release pending protocol callbacks when the `Connection.run()` task is cancelled at loop shutdown, so their cancellation is not absorbed by `Channel._abort()` waiting for a reply that can never arrive
- retrieve callback exceptions in `Channel._abort()` even when its wait is interrupted by another cancellation

Fixes https://github.com/microsoft/playwright/issues/42296

### DIFF
--- a/playwright/_impl/_connection.py
+++ b/playwright/_impl/_connection.py
@@ -356,10 +356,8 @@ class Connection(EventEmitter):
         try:
             await self._transport.run()
         except asyncio.CancelledError:
-            # This task is cancelled by asyncio.run() at loop shutdown. Release
-            # any tasks waiting on protocol replies that will never arrive,
-            # otherwise their cancellation is absorbed by Channel._abort()
-            # waiting for the reply and the loop never finishes closing.
+            # Cancelled at loop shutdown - release tasks waiting on replies
+            # that will never arrive.
             if not self._closed_error:
                 self.cleanup()
             raise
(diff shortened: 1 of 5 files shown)


playwright/_impl/_connection.py
```diff
@@ -353,7 +353,14 @@ async def init() -> None:
 
         await self._transport.connect()
         self._init_task = self._loop.create_task(init())
-        await self._transport.run()
+        try:
+            await self._transport.run()
+        except asyncio.CancelledError:
```
dgozman, line 358: Why only on `CancelledError` and not on any error? How is this one special?

playwright/_impl/_connection.py
```diff
@@ -452,30 +459,32 @@ def _send_message_to_server(
     async def _abort(
         self, object: ChannelOwner, callback: ProtocolCallback, reason: str
     ) -> None:
+        if not self._closed_error:
```
dgozman, line 462: I think we would benefit from another test, where we cancel a task like click(), while at the same time closing the connection. I don't think existing tests cover this change.

yury-s: The transport should raise and we don't need this check at all.

PR #3189, opened 2026-08-26
fix(evaluate): serialize and parse RegExp values

### LINKED ISSUE #3188: [Bug]: RegExp values are not serialized by evaluate(), and returned RegExps leak internal protocol JSON
### Version

1.62.0

### Steps to reproduce

```py
import re

with sync_playwright() as p:
    browser = p.chromium.launch()
    page = browser.new_page()

    # 1. Passing a compiled pattern in
    print(page.evaluate("(v) => [typeof v, String(v)]", re.compile(r"a\d+", re.I)))

    # 2. Getting a RegExp back out
    print(page.evaluate("() => /a\\d+/gi"))

    # 3. Nested in a structure
    print(page.evaluate("() => ({x: [/foo/m]})"))

    browser.close()
```

### Expected behavior

1. `['object', '/a\\d+/i']`, and the value is a real `RegExp` inside the page.
2. `re.compile('a\\d+', re.IGNORECASE)`
3. `{'x': [re.compile('foo', re.MULTILINE)]}`

This is what playwright-dotnet does today. `EvaluateArgumentValueConverter.cs` serializes `Regex` (line 138) and parses `"r"` back into a `Regex` (line 351).

### Actual behavior

1. `['undefined', 'undefined']`, the argument is silently dropped
2. `{'r': {'p': 'a\\d+', 'f': 'gi'}}`
3. `{'x': [{'r': {'p': 'foo', 'f': 'm'}}]}`

Two separate problems, both silent:

- `serialize_value()` in `playwright/_impl/_js_handle.py` has no branch for `re.Pattern`, so a compiled pattern falls through to `{"v": "undefined"}`.
- `parse_value()` has no branch for `"r"`, so it hits the bare `return value` at the end and hands back the raw wire format. There is no way to tell from the outside that this dict is not the actual result.

The protocol supports this in both directions and the driver already implements both sides (`{r: {p, f}}` on serialize, `new RegExp(v.r.p, v.r.f)` on parse), so this is only missing on the Python side.

### Additional context

Happy to send a PR, I have one ready. The serialize direction can reuse the existing `escape_regex_flags()` in `_str_utils.py`, which already maps `re.IGNORECASE|DOTALL|MULTILINE` to `i`/`s`/`m` for locators and route matching. The parse direction needs the inverse.

Two things worth an opinion before I do:

1. JavaScript flags with no `re` equivalent (`g`, `y`, `d`, `u`, `v`). I would drop them rather than raise, since `/foo/g` is very common and raising would be worse than today's behavior. Note that playwright-dotnet's `FromInlineFlags` throws on these, which looks like a separate bug over there.
2. A JS pattern that is not valid Python `re` syntax, for example `/(?<name>x)/` (Python spells it `(?P<name>x)`). With this change that would raise from `re.compile` instead of returning the dict. I think raising is acceptable since the dict was never usable anyway, but let me know if you would rather it degrade some other way.

### Environment

```
- Operating System: macOS 26.3.1
- CPU: arm64
- Browser: Chromium
- Python Version: 3.14.0
```

FadeHack:
`serialize_value()` has no branch for `re.Pattern`, so passing a compiled pattern to `evaluate()` silently sends `undefined`. `parse_value()` has no branch for `"r"`, so a RegExp coming back from the page reaches the caller as the raw wire dict instead of a pattern:

```py
>>> page.evaluate("(v) => String(v)", re.compile(r"a\d+", re.I))
'undefined'
>>> page.evaluate("() => /a\\d+/gi")
{'r': {'p': 'a\\d+', 'f': 'gi'}}
```

The protocol already carries regexes as `{r: {p, f}}` and the driver implements both sides, so this just adds the two missing branches on the Python side. playwright-dotnet does the same in `EvaluateArgumentValueConverter.cs`.

Flags reuse the existing `escape_regex_flags()` helper. On the way back, the JavaScript-only flags (`g`, `y`, `d`, `u`, `v`) have no `re` equivalent and are dropped.

Fixes https://github.com/microsoft/playwright-python/issues/3188

### DIFF
--- a/playwright/_impl/_js_handle.py
+++ b/playwright/_impl/_js_handle.py
@@ -16,15 +16,17 @@ import base64
 import collections.abc
 import datetime
 import math
+import re
 import struct
 import traceback
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Pattern, Union
 from urllib.parse import ParseResult, urlparse, urlunparse
 
 from playwright._impl._connection import Channel, ChannelOwner, from_channel
 from playwright._impl._errors import Error, is_target_closed_error
 from playwright._impl._map import Map
+from playwright._impl._str_utils import escape_regex_flags, parse_regex_flags
 
 if TYPE_CHECKING:  # pragma: no cover
     from playwright._impl._element_handle import ElementHandle
@@ -182,6 +184,8 @@ def serialize_value(
         return {"s": value}
     if isinstance(value, ParseResult):
         return {"u": urlunparse(value)}
+    if isinstance(value, Pattern):
+        return {"r": {"p": value.pattern, "f": escape_regex_flags(value)}}
 
     if value in visitor_info.visited:
         return dict(ref=visitor_info.visited[value])
@@ -246,6 +250,9 @@ def parse_value(value: Any, refs: Optional[Dict[int, Any]] = None) -> Any:
             error._name = value["e"]["n"]
             error._stack = value["e"]["s"]
             return error
+
+        if "r" in value:
+            return re.compile(value["r"]["p"], parse_regex_flags(value["r"]["f"]))
 
         if "a" in value:
             a: List = []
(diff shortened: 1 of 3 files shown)


FadeHack:

PR #3201, opened 2026-09-15
Type credentials and OPFS in storage state

shubhambhar007:
Storage snapshots can include credentials and OPFS entries, but the Python `StorageState` type rejects both. This adds those optional fields while keeping the existing required origin fields intact, plus a small regression test for the type definitions.

Refs microsoft/playwright#42732

Evidence: the same storage-state example produced two mypy “Extra key” errors before this change and passes after it.

```text
shubhambhardwaj@shubhambhardwaj-mac playwright-python-42732 % ../playwright-typecheck-venv/bin/mypy --no-incremental --follow-imports=normal ../storage_state_typecheck.py
Success: no issues found in 1 source file
```

I could not run the pytest test locally because the checkout is missing the test harness dependency `pixelmatch`.


yury-s:
They are intentionally untyped.

